PR #240, opened 2020-08-11
MCMC, first pass. Some behavior is still abnormal, marked by TODOs.

rimangan:


popgen/MCMC_Selection.go
```diff
@@ -0,0 +1,103 @@
...
+//muZero and sigmaZero represent the starting hyperparameter values.
+func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations int, logSpace bool) ([]float64, []float64, []bool) {
+	muList := make([]float64, iterations)
+	sigmaList := make([]float64, iterations)
+	acceptList := make([]bool, iterations)
+	//initialization to uninformative standard normal
+	t := InitializeTheta(muZero, sigmaZero, len(data.sites))
+	for i := 0; i < iterations; i++ {
+		tCandidate := GenerateCandidateThetaPrime(t)
+		if Metropolis_Accept(tCandidate, t, data, logSpace) {
```
craiglowe, line 93: I think that tCandidate and t are swapped.  t should be first?

popgen/MCMC_Selection.go
```diff
@@ -0,0 +1,103 @@
...
+			oldGivenNew = oldGivenNew * numbers.NormalDist(tOld.alpha[i], tOld.mu, tOld.sigma)
+		}*/
+	return oldGivenNew / newGivenOld
+}
+
+func Bayes_Ratio(old Theta, thetaPrime Theta, data AFS, logSpace bool) float64 {
+	if logSpace {
+		return AFSLogLikelihood(data, old.alpha) * old.probability / (AFSLogLikelihood(data, thetaPrime.alpha) * thetaPrime.probability)
+	}
+	return AFSLikelihood(data, old.alpha) * old.probability / (AFSLikelihood(data, thetaPrime.alpha) * thetaPrime.probability)
```
craiglowe, line 43: I think that this should be "prime / old" instead of "old / prime".  The accept probability should go higher when the proposal has a higher likelihood.

cmd/vcfAFS/vcfAFS.go
```diff
@@ -0,0 +1,41 @@
...
+	defer out.Close()
+	for i := 0; i < len(f); i++ {
+		fmt.Fprintf(out, "%f\n", f[i])
+	}
+}
+
+func usage() {
+	fmt.Print(
+		"vcfAFS - Returns allele frequency spectrum information in a text file for graphing.\n" +
+			"vcfAFS - vcfFile.vcf outFile.txt")
```
das106, line 24: Could you put a newline at the end of this line. When i run the cmd it gets pushed up against the expected args fatal and is a bit hard to read.

cmd/selection_MCMC/selection_MCMC.go
```diff
@@ -0,0 +1,50 @@
+package main
```
craiglowe, line 1: would be good to remove underscores.  To follow conventions I think you should consider "selectionMcmc/selectionMcmc.go"

PR #241, opened 2020-08-11
More GoDocs and unused functions removed

rimangan:


rimangan:
FilterToGroups was a redundant function with an existing function in the same file.

PR #242, opened 2020-08-12
Christi simulation code

cfauci:
Changed the name of Sophie's old newick tree code to expandedTree (package and filename). Updated Simulate code to use dna package and created test files. All tests pass.


das106:
Looks great! I am seeing a few duplicate functions between expandedTree.go and simulate.go. Are those intended or should they be removed?

PR #243, opened 2020-08-12
Proposal to add bit flag to Giraf

das106:
It would be useful to have a bit flag in the giraf struct to store boolean fields such as fwd/rev read similar to how this information is stored in the sam format. In this pull request I make the necessary changes to the Giraf struct, read/write functions, GSW and any other packages whose tests failed after making the change. Currently I am only storing 5 of the flags that are covered by the sam flag, however the bit position of the added flags are the same as their sam counterpart so additional info can be added as desired.


giraf/giraf.go
```diff
@@ -14,6 +14,7 @@ type Giraf struct {
 	QName     string
 	QStart    int
 	QEnd      int
+	Flag      uint16
```
craiglowe, line 17: Why not make these separate bool variables?  More space?  Even if we smash them together in the encoding for the file, maybe it makes sense to keep them separate in the data struct?  I did see that you have nice functions for accessing the individual bits, so maybe that is enough to keep it friendly?

das106: We could just make them bools in the struct and just encode them into a flag when we write to file i suppose. The only downside I see is a larger memory footprint (2 bytes for the flag vs 16 bytes for 16 bools). On the scale of the whole giraf struct an extra 14 bytes probably isn't much. I could go either way.

PR #244, opened 2020-08-13
channel/goroutine bug fixes and minior optimizations

edotau:


cmd/gsw/axtHelper.go
```diff
@@ -6,18 +6,37 @@ import (
...
 			vcf.WriteVcf(file, i)
 		}
+		file.Close()
+
+		unSortedVcf := vcf.Read(tmp)
+		final := fileio.EasyCreate(output)
+		vcf.NewWriteHeader(final, header)
+		vcf.Sort(unSortedVcf)
+
+		for _, v := range unSortedVcf {
```
craiglowe, line 34: Maybe a different name?  I think on this line, unSortedVcf is actually sorted?

cmd/gsw/axtHelper.go
```diff
@@ -6,18 +6,37 @@ import (
...
 	"github.com/vertgenlab/gonomics/vcf"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
 )
 
 func convertAxt(axtFile, format, targetFa, output string) {
 	switch format {
 	case "vcf":
```
craiglowe, line 17: It looks like this now converts and sorts.  Maybe convertAxt should just convert and then sort it afterwards?

cmd/gsw/axtHelper.go
```diff
@@ -6,18 +6,37 @@ import (
...
 		vcfChannel := goChannelAxtVcf(axtFile)
+		tmp := output + ".tmp"
+		file := fileio.EasyCreate(tmp)
```
craiglowe, line 20: I am not sure what we gain by writing to disk.  Why not read it into a slice from the channel and then sort that slice.  It looks like right now we go from a channel to file, a file to a slice, and then a slice to a sorted slice.  Seems like we can just cut out the file.

PR #247, opened 2020-08-14
Pairwise multiFa alignment to Vcf

rimangan:
Cmd takes a pairwise multiFa alignment and calls variants. Sensitive to indels, passes tests.


convert/convert.go
```diff
@@ -210,3 +223,56 @@ func getWigChromIndex(s string, wigSlice []*wig.Wig) int {
...
+	var deletion bool = false
+	var insertionAlnPos int
+	var deletionAlnPos int
+	answer := make([]*vcf.Vcf, 0)
+
+	for i := 0; i < len(f[0].Seq); i++ { //loop through reference alignment positions
+		if f[0].Seq[i] == dna.Gap { //reference is gap (insertion)
+			if pastStart {
+				if !insertion {
+					insertionAlnPos = i - 1 //TODO: breaks for alignments starting with an insertion(gap in Aln postion 0 of the reference)
```
craiglowe, line 241: What happens now when this is in the input?  Can we print an error message and die?

convert/convert.go
```diff
@@ -210,3 +223,56 @@ func getWigChromIndex(s string, wigSlice []*wig.Wig) int {
...
+				}
+				insertion = true
+			}
+		} else if f[0].Seq[i] != f[1].Seq[i] {
+			pastStart = true
+			if insertion { //catches the case where an insertion, now complete, is followed directly by a snp.
+				answer = append(answer, &vcf.Vcf{Chr: chr, Pos: int64(fasta.AlnPosToRefPos(f[0], insertionAlnPos) + 1), Id: ".", Ref: dna.BaseToString(f[0].Seq[insertionAlnPos]), Alt: dna.BasesToString(f[1].Seq[insertionAlnPos:i]), Qual: 100.0, Filter: "PASS", Info: ".", Format: ".", Notes: "."})
+			}
+			if f[1].Seq[i] == dna.Gap { //alt is gap (deletion)
```
craiglowe, line 250: I am getting a little thrown off by the last of symmetry between insertions and deletions.  Can you add a few comments or an overview of your thought process?  I am having a little trouble wrapping my head around the logical flow inside of this for-loop that looks at each base in the alignment.

convert/convert_test.go
```diff
@@ -0,0 +1,25 @@
+package convert
+
+import (
+	"github.com/vertgenlab/gonomics/dna"
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/vcf"
+	"testing"
+)
+
+var seqA []dna.Base = dna.StringToBases("--TTTC--ATGAATAA")
```
craiglowe, line 10: I had thought that the TODO comment in the code above mentioned that the reference sequence starting with gaps would not be handled well.

cmd/multiFaToVcf/multiFaToVcf.go
```diff
@@ -0,0 +1,43 @@
...
+	"log"
+	"github.com/vertgenlab/gonomics/convert"
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/vcf"
+)
+
+func multiFaToVcf(inFile string, chr string, outFile string) {
+	f := fasta.Read(inFile)
+	v := convert.PairwiseFaToVcf(f, chr)
+	vcf.Write(outFile, v)
```
edotau, line 15: We should bring this up again at code review because it keep coming up. The short term solution might be something that looks like this:
```
vcfWriter := fileio.EasyCreate(outFile)
header := vcf.NewHeader("nameFastaSample")
vcf.NewWriteHeader(vcfWriter, header)
vcf.WriteVcfToFileHandle(vcfWriter, v)
vcfWriter.Close() //defer vcfWriter.Close() at the top also works
```

I think most people like one wrap function to easily call, so long term vcf.Write() should be something like this:
`vcf.Write(filename string, header *vcf.VcfHeader, variants []*vcf.Vcf)`
```
func Write(filename string, header *VcfHeader, variants []*Vcf) {
	vcfWriter := fileio.EasyCreate(filename)
        WriteHeader(vcfWriter, header)
        WriteVcfToFileHandle(vcfWriter, variants)
        vcfWriter.Close()
}
```

convert/convert.go
rimangan: Oh I should remove this comment, the pastStart was an addition in the latest commit that solves this problem

convert/convert_test.go
rimangan: Yes this is outdated, sequences starting with insertions now pass

PR #249, opened 2020-08-14
Updated reconstruct code

cfauci:
I've removed repeated functions from simulate.go and reconstruct.go, and the test in reconstruct_test.go that was timing out.


reconstruct/reconstruct.go
```diff
@@ -3,49 +3,49 @@ package reconstruct
...
-		fastas = append(fastas, branches[i].Fasta)
+	for j := 0; j < len(branches); j++ {
+		fastas = append(fastas, branches[j].Fasta)
 	}
 
 	var fas []*fasta.Fasta
 	//loop sites
-	for i := 0; i < len(fastas); i++ {
-		fas = append(fas, fastas[i])
+	for k := 0; k < len(fastas); k++ {
```
craiglowe, line 31: I am having trouble seeing what this loop does.  In the previous loop we went over all the nodes and added the sequence at that node to a slice.  It looks like this now goes over that slice and makes a new one that is equal to the one we already had?  Seems like we could remove "fas" and this loop and just write fastas instead of fas

reconstruct/reconstruct.go
```diff
@@ -3,49 +3,49 @@ package reconstruct
...
+	"github.com/vertgenlab/gonomics/expandedTree"
 	"github.com/vertgenlab/gonomics/fasta"
-	"github.com/vertgenlab/gonomics/tree_newick"
 )
 
 //final function to run
-func Reconstruct(root *tree_newick.NTree, filename_output string) {
-	leaf := Get_leaf(root)
-	branches := Get_branch(root)
+func Reconstruct(root *expandedTree.ETree, outFilename string) {
```
craiglowe, line 11: Minor, but it might be easier in the long-run if you separate out the reconstruction of the sequence from writing that sequence to a file.  I would have reconstruct return a []*fasta.Fasta instead of returning nothing and writing that []*fasta.Fasta to a filename.

reconstruct/reconstruct.go
```diff
@@ -93,78 +93,41 @@ func Yhat(r []float64) int {
...
-	return leaf
-}
-
-//set the state of the tree given the Fasta and the position
-func Set_state(node *tree_newick.NTree, pos int) {
-	leaf := Get_leaf(node)
-	var leaf_names []string
+//set the state of the tree given the Fasta and the position (zero-based)
+func SetState(node *expandedTree.ETree, pos int) {
+	leaf := expandedTree.GetLeaf(node)
```
craiglowe, line 98: From looking later in the function, is the only reason to get the leaves to know if this node we are currently at is a leaf?  Can we just not calculate this and instead use "if node.Right == nil && node.Left == nil" to test if we are currently at a leaf later in the function?

reconstruct/reconstruct.go
```diff
@@ -93,78 +93,41 @@ func Yhat(r []float64) int {
...
 	}
+
+	//didn't have a situation where position could be the last base of a sequence? Was that because of stop codons?
 	node.State = 4
-	node.Stored = []float64{0, 0, 0, 0}
-	for i := 0; i < len(leaf_names); i++ {
-		if node.Name == leaf_names[i] {
-			if len(node.Fasta.Seq) > pos {
+	node.Stored = []float64{0, 0, 0, 0} //everything starts as N
+	for i := 0; i < len(leafNames); i++ {
```
craiglowe, line 107: Can we replace this loop and the "if" on the next line with "if node.Left == nil && node.Right == nil"?  It seems like we are looping over the list of leaf names to see if we are at a leaf.  Is that right?

reconstruct/reconstruct.go
```diff
@@ -93,78 +93,41 @@ func Yhat(r []float64) int {
...
+		if node.Name == leafNames[i] {
+			if len(node.Fasta.Seq) > pos { //keeps it from returning an index out of range error
+				node.State = int(node.Fasta.Seq[pos]) //State = base at given position
+				// node.Stored = []float64{0, 0, 0, 0} //do i need to respecify thins every time or within the larger loop?
+				node.Stored[node.State] = 1 //the position in the list which corresponds to the base of State becomes 1 (state =0 (A) then 0 position of stored becomes 1)
+			} else if len(node.Fasta.Seq) == pos { //I'm gonna add this here because it feels necessary?
```
craiglowe, line 113: I think we should print an error (log.Fatal("Error message\n")) and die if pos >= len(node.Fasta.Seq)

reconstruct/reconstruct.go
```diff
@@ -93,78 +93,41 @@ func Yhat(r []float64) int {
...
+	//didn't have a situation where position could be the last base of a sequence? Was that because of stop codons?
 	node.State = 4
-	node.Stored = []float64{0, 0, 0, 0}
-	for i := 0; i < len(leaf_names); i++ {
-		if node.Name == leaf_names[i] {
-			if len(node.Fasta.Seq) > pos {
+	node.Stored = []float64{0, 0, 0, 0} //everything starts as N
+	for i := 0; i < len(leafNames); i++ {
+		if node.Name == leafNames[i] {
+			if len(node.Fasta.Seq) > pos { //keeps it from returning an index out of range error
```
craiglowe, line 109: This should not be solved her, but in whatever is calling this function.  Pos should not be coming in as a number greater than positions available.  Good to check here, but it should be a check and then aborting if it has been called in a way that does not make sense

reconstruct/reconstruct.go
```diff
@@ -93,78 +93,41 @@ func Yhat(r []float64) int {
...
+				node.State = int(node.Fasta.Seq[pos]) //State = base at given position
+				// node.Stored = []float64{0, 0, 0, 0} //do i need to respecify thins every time or within the larger loop?
```
craiglowe, line 111: I would not allocate memory, which calling []float64{} will do, but you will need to set the positions that are not node.State to zero (along with setting the value at node.State to 1.  Maybe loop from 0 to 3 and have an if that checks if the variable looping is equal to node.State and use that if to set the value in node.Stored to either 0 or 1?

reconstruct/reconstruct_test.go
```diff
@@ -16,20 +15,19 @@ var input = []struct {
...
 	for _, test := range input {
 
-		//example of how to run simulation: 1) read in tree (no fastas) 2) simulate random gene 3) simulate evolution 4) remove ancestors for reconstruction 4) set up fastas
-		tre, er := tree_newick.ReadNewick(test.newick_filename)
+		//example of how to run simulation: 1) read in tree (no fastas) 2) simulate random gene 3) simulate evolution 4) remove ancestors for reconstruction 4) assign fastas
+		tre, er := expandedTree.ReadNewick(test.newick_filename)
 		if er != nil {
 		}
-		simulate.Rand_gene("test", test.length)
+		simulate.RandGene("test", test.length, 0.42) //galGal6
```
craiglowe, line 24: I would create a variable that is called gcContent and set it to 0.42, then pass in gcContent to the function.  This makes it easier to read for someone seeing the code for the first time.  Otherwise they might not be sure what the 0.42 is for.

reconstruct/reconstruct.go
```diff
@@ -236,18 +177,18 @@ func Fix_fc(root *tree_newick.NTree, node *tree_newick.NTree) []float64 {
...
-			leafs[i].State = int(leafs[i].Fasta.Seq[j])
+func LoopNodes(root *expandedTree.ETree) { //what is fixing the nodes doing?
+	leaves := expandedTree.GetLeaves(root)
+	branches := expandedTree.GetBranch(root)
+	for j := 0; j < len(leaves[0].Fasta.Seq); j++ {
+		for i := 0; i < len(leaves); i++ {
+			leaves[i].State = int(leaves[i].Fasta.Seq[j])
 		}
 	}
 	for j := 0; j < len(branches); j++ {
```
cfauci, line 188: Change term from j to something else. Should be inside the J loop above so each column is handled independently in leaves and branches

reconstruct/reconstruct.go
```diff
@@ -236,18 +177,18 @@ func Fix_fc(root *tree_newick.NTree, node *tree_newick.NTree) []float64 {
...
-func Loop_nodes(root *tree_newick.NTree) {
-	leafs := Get_leaf(root)
-	branches := Get_branch(root)
-	for j := 0; j < len(leafs[0].Fasta.Seq); j++ {
-		for i := 0; i < len(leafs); i++ {
-			leafs[i].State = int(leafs[i].Fasta.Seq[j])
+func LoopNodes(root *expandedTree.ETree) { //what is fixing the nodes doing?
+	leaves := expandedTree.GetLeaves(root)
+	branches := expandedTree.GetBranch(root)
+	for j := 0; j < len(leaves[0].Fasta.Seq); j++ {
```
cfauci, line 183: Reconstruct calls loopNodes for every column. work out which pieces of logic here can be removed

reconstruct/reconstruct.go
```diff
@@ -93,139 +87,86 @@ func Yhat(r []float64) int {
...
 			sum := 0.0
 			for j := 0; j < 4; j++ {
 				for k := 0; k < 4; k++ {
 					sum = sum + Prob(i, j, node.Left.BranchLength)*Prob(i, k, node.Right.BranchLength)*node.Left.Stored[j]*node.Right.Stored[k]
-				}
+				}//probability that base i becomes either base j or k time the probability stored in the []float64 of that base appearing
 			}
 			node.Stored[i] = sum
 		}
-	} else {
```
cfauci, line 180: go back to previous version

edotau (dismissed):

simulate/simulate.go
```diff
@@ -18,15 +18,14 @@ func RandGene(name string, length int, GCcontent float64) []*fasta.Fasta {
 	seq := []dna.Base{dna.A, dna.T, dna.G}
 	randLength := length - 6
 
-	r := rand.Float64()
-
 	if length%3 != 0 {
 
 		fmt.Print("length must be divisible by three")
```
craiglowe, line 23: Should be log.Fatal

reconstruct/reconstruct_test.go
```diff
@@ -1,48 +1,38 @@
...
 
 func Test_reconstruct(t *testing.T) {
-	start := time.Now()
 	for _, test := range input {
 
-		//example of how to run simulation: 1) read in tree (no fastas) 2) simulate random gene 3) simulate evolution 4) remove ancestors for reconstruction 4) set up fastas
-		tre, er := tree_newick.ReadNewick(test.newick_filename)
+		//example of how to run simulation: 1) read in tree (no fastas) 2) simulate random gene 3) simulate evolution 4) remove ancestors for reconstruction 4) assign fastas
+		tre, er := expandedTree.ReadNewick(test.newickFilename)
 		if er != nil {
```
craiglowe, line 25: Should be a log.Fatal inside here

rimangan (dismissed):
Looks great!

craiglowe (commented):
Two quick comments from Riley during review

rimangan (approved):

craiglowe (approved):
Great work!

LCBartelt (approved):

PR #250, opened 2020-08-18
GVCF Tajima

rimangan:
In progress. Functions to calculate Tajima's D from gVCFs. More useful than the fasta version for dealing with 1000HGP data or other large scale datasets.


popgen/GvcfZipper.go
```diff
@@ -0,0 +1,80 @@
...
+	"sort"
+	"strings"
+	"log"
+	"github.com/vertgenlab/gonomics/vcf"
+	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/dna"
+)
+
+//TajimaZipperVcf returns the Tajima's D value for each entry in a sorted bed slice using a pre-SORTED gVCF file for variant information.
+func TajimaZipperVcf(b []*bed.Bed, gVCFFile string) []float64 {
```
das106, line 13: I didn't think of this yesterday, but I think this will break if there are overlapping beds. Is that a dealbreaker? You may want to add a note in the documentation too.

popgen/GvcfZipper.go
```diff
@@ -0,0 +1,80 @@
...
+	var answer []float64
+	var all []*IndividualAllele
+	answer = make([]float64, 0)
+	alpha := vcf.GoReadGVcf(gVCFFile)//must be sorted beforehand
+	sort.Sort(bed.ByGenomicCoordinates{b})
+	var currBedIndex int = 0
+	var currVariants []*vcf.Vcf
+	currVariants = make([]*vcf.Vcf, 0)
+	for i := range alpha.Vcfs {
+		for 1 > 0 {//infinite condition for break loop
```
das106, line 23: This is fine, but just in case you didn't know you can also make an infinite loop by not giving any arguments to for. e.g. 
`for {
...
if x {break}
}`

popgen/GvcfTajima.go
```diff
@@ -0,0 +1,132 @@
...
+	var firstTime bool = true
+	var intervals []interval.Interval
+	intervals = make([]interval.Interval, len(b))
+	for i := 0; i < len(b); i++ {
+		intervals[i] = b[i]
+	}
+	tree := interval.BuildTree(intervals)
+	for i := range alpha.Vcfs {
+		if len(interval.Query(tree, i, "any")) > 0 {//in other words, if the variant overlaps any of the beds
+			if !strings.ContainsAny(i.Alt, "<>") {//gVCF converts the alt and ref to []DNA.base, so structural variants with <CN0> notation will fail to convert. This check allows us to ignore these cases.
```
craiglowe, line 100: Should this be a quick note up top, that these variants are skipped/ignored

popgen/GvcfTajima.go
```diff
@@ -0,0 +1,132 @@
...
+func GVCFCalculateA2(n int) float64 {
+	var answer float64
+
+	for i := 1; i < n - 1; i++ {
+		answer += (1.0 / math.Pow(float64(i), 2))
+	}
+	return answer
+}
+
+func GVCFCalculateA1(n int) float64 {
```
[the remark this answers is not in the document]
rimangan: A1 is dependent only on S, the number of segregating sites, so techincally this should work for gVCF or for Fasta. I just gave it this name for consistency.

popgen/GvcfTajima.go
```diff
@@ -0,0 +1,140 @@
...
+} 
+
+//IndividualDist is a helper function for Tajima's D calculations that calculates the distance (the number of segregating sites that are different by state) between two IndividualAllele structs.
+func IndividualDist(a IndividualAllele, b IndividualAllele) int {
+	var answer int = 0
+	if len(a.sites) != len(b.sites) {
+		log.Fatalf("IndividualAllele sites lists must be the same length to calculate distance.")
+	}
+	for i := 0; i < len(a.sites); i++ {
+		if len(a.sites[i]) != len(b.sites[i]) {
```
craiglowe, line 30: How do we feel about gaps?  Do we want a check where it fails if there are gaps?  Right now it counts them as a difference

PR #252, opened 2020-08-20
Logspace binomial coefficients for large numbers

rimangan:


numbers/numbers.go
```diff
@@ -157,6 +157,30 @@ func BinomCoefficient(n int, k int) int {
...
+//BinomCoefficientLog returns log(n choose k), where log is the natural logarithm. 
+//Ideal for large numbers as this raises the overflow ceiling considerably.
+func BinomCoefficientLog(n int, k int) float64 {
+	if n < 0 || k < 0 || k > n {
+		log.Fatalf("The binomial coefficient call could not be handled: n=%d and k=%d\n", n, k)
+	}
+	if n-k > k {
+		k = n - k
+	}
+	// this special case is handled here so that we don't ask for negative memory for denom
```
craiglowe, line 169: I think this may not be needed anymore since you are no longer allocating memory

numbers/numbers.go
```diff
@@ -157,6 +157,30 @@ func BinomCoefficient(n int, k int) int {
...
+	if k == n {
+		return 0.0
+	}
+	var x, y int
+	var numer, denom float64 = 0.0, 0.0
+	for x = k + 1; x < n+1; x++ {
+		numer = numer + math.Log(float64(x))
```
craiglowe, line 176: Can we use the numbers.MultiplyLog(a,b) or whatever it is called

PR #253, opened 2020-08-21
Simpleio

edotau:


giraf/giraf.go
```diff
@@ -17,7 +17,7 @@ type Giraf struct {
 	Flag      uint8
 	PosStrand bool
 	Path      *Path
-	Aln       []*cigar.Cigar // current cigar will need to be expanded
+	Aln       []simpleio.ByteCigar// current cigar will need to be expanded
```
craiglowe, line 20: I am not sure if the simpleio package is in this request so I don't think I can see the data structure.  Would it make sense to put this data structure in the cigar package?

cmd/gsw/ggTools.go
```diff
@@ -64,24 +64,27 @@ func RunGgTools() {
 		log.Printf("Reading: %s\n", inFile)
 		switch true {
 		case chain.IsChainFile(inFile):
-			log.Printf("chain input detected!\n")
+			log.Printf("Input chain detected...\n")
```
craiglowe, line 67: The changes to this file seem like they could be separated from many of the other changes in the pull request.  Maybe this pull request could be broken into a handful of smaller ones?

simpleGraph/bytesAlign.go
```diff
@@ -0,0 +1,422 @@
...
+			answer = append(answer, simpleio.ByteCigar{RunLen: uint32(lengthOfRead - front - runLen), Op: 'S'})
+		}
+		return answer
+	} else {
+		return cig
+	}
+
+}
+
+func SimpleSequenceGraph(ref <-chan *fasta.Fasta, vcfMap map[string][]*vcf.Vcf) *SimpleGraph {
```
craiglowe, line 139: Maybe this should be in a separate file and a separate pull request from the changes to the alignment code?

simpleGraph/bytesAlign.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+	if vcf.Snp(vcfsChr[len(vcfsChr)-2]) || isHaplotypeBlock(vcfsChr[len(vcfsChr)-2]) {
+		AddEdge(altAllele, lastNode, 1)
+	}
+	AddEdge(lastMatch, lastNode, 1)
+	SetEvenWeights(lastMatch)
+	return genome
+}
+
+func checkAlignment(aln *giraf.Giraf, genome *SimpleGraph) bool {
```
craiglowe, line 304: It might be good to split apart this file into those functions that deal with making the graph, and those that deal with aligning to the graph

simpleGraph/bytesAlign.go
```diff
@@ -0,0 +1,422 @@
...
+		}
+	}
+
+	log.Printf("Mapped %d out of %d\n", numReads-unmapped, numReads)
+	log.Printf("%f of the reads are mapping correctly\n", percentOfFloat(numReads-unmapped, numReads))
+
+	wg.Done()
+}
+
+func ReadGenomeGraph(filename string) *SimpleGraph {
```
craiglowe, line 365: I think this could even be in a separate file and pull request: 1) making the graph, 2) reading the graph, 3) aligning to the graph

simpleGraph/dynamicAln.go
```diff
@@ -0,0 +1,211 @@
...
+			if i == 0 && j == 0 {
+				m[i][j] = 0
+			} else if i == 0 {
+				m[i][j] = m[i][j-1] + gapPen
+				trace[i][j] = 'I'
+			} else if j == 0 {
+				m[i][j] = m[i-1][j] + gapPen
+				trace[i][j] = 'D'
+			} else {
+				m[i][j], trace[i][j] = simpleio.ByteMatrixTrace(m[i-1][j-1]+scores[alpha[i-1]][beta[j-1]], m[i][j-1]+gapPen, m[i-1][j]+gapPen)
```
craiglowe, line 29: hard to say for sure, because I know I have not seen the simpleio package, but it seems like maybe functions deep in the alignment code should be with the other alignment code?  From the name I am guessing that simpleio deals mostly with input and output and making it simpler?

PR #254, opened 2020-08-24
Fancy Cigar in Aligner

DanielSnellings:
Added conversion to fancy cigar in aligner. Not sure if this will get merged or not, but I figured it would be good to put it up for discussion. Performance impact seems to be <5% depending on read with the majority of reads <2%. Uncomment the lines in toGiraf.go to test


simpleGraph/toGiraf.go
```diff
@@ -81,9 +83,91 @@ func GraphSmithWatermanToGiraf(gg *SimpleGraph, read *fastq.FastqBig, seedHash m
...
 }
 
+func GirafToExplicitCigar(giraf *giraf.Giraf, graph *SimpleGraph) []*cigar.Cigar {
+	var answer []*cigar.Cigar
+	var seqIdx, refIdx, pathIdx int
+	refIdx = giraf.Path.TStart
+	var k, runLenCount int64
+
+	if giraf.Aln[0].Op == '*' {
+		return nil
```
craiglowe, line 103: Do we want this to be nil?  '*' means unaligned read, right?  Do we want to keep the cigar as a '*' for unaligned or should we do nil?  This should probably be consistent across all cigars, right?

simpleGraph/toGiraf.go
```diff
@@ -81,9 +83,91 @@ func GraphSmithWatermanToGiraf(gg *SimpleGraph, read *fastq.FastqBig, seedHash m
...
+	}
+
+	for i := 0; i < len(giraf.Aln); i++ {
```
craiglowe, line 106: This looks really similar to cigar.MakeExplicit.  Can we call that?

simpleGraph/toGiraf.go
edotau: yes, this is not consistent throughout the code base but i think the slice of cigars should be set to nil if it is unaligned. When we print (to string and/or to file) we check if == nil, print *

cigar/fancy.go
```diff
@@ -14,6 +14,11 @@ func MakeExplicit(cigar []*Cigar, sequence []dna.Base, reference []dna.Base) []*
 	var answer []*Cigar
 	var seqIdx, refIdx int
 	var k, runLenCount int64
+
+	if cigar[0].Op == '*' {
+		return []*Cigar{&Cigar{Op: '*'}}
```
craiglowe, line 19: I think Eric weighed in here saying that the original version (return nil) is correct and I think my earlier comment threw us off.  Seems like there are two ways to do it currently in the code, but we are going to try to converge on the "nil = unaligned"

DanielSnellings: Switched it back to nil

PR #255, opened 2020-08-25
Craig three bit update

craiglowe:
I have not added all the tests that are needed, but I am putting up a new version of dnaThreeBit so that our compressed dna format can now handle Ns

### DIFF
--- a/dnaThreeBit/convert.go
+++ b/dnaThreeBit/convert.go
@@ -3,7 +3,7 @@ package dnaThreeBit
 import (
 	"bytes"
 	"log"
-	"unicode/utf8"
+	"github.com/vertgenlab/gonomics/dna"
 )
 
 func RuneToThreeBitBase(r rune) ThreeBitBase {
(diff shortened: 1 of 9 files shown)


dnaThreeBit/convert.go
```diff
@@ -0,0 +1,89 @@
...
+func FromString(s string) *ThreeBit {
+	answer := &ThreeBit{Seq: []uint64{}, Len: 0}
+	for _, runeValue := range s {
+		answer = Append(answer, RuneToThreeBitBase(runeValue))
+	}
+	return answer
+}
+
+func ToString(fragment *ThreeBit) string {
+	var buffer bytes.Buffer
```
edotau, line 68: something I learned recently is after go 1.13, the strings library has something similar to bytes.Buffer but optimized for concatenating strings changing this is a decent speed up. strings.Builder comes out ahead in situations like this one.  I can send you some of the benchmarks I did over slack.

```
func ToString(fragment *ThreeBit) string {
	var buffer strings.Builder

	for i := 0; i < fragment.Len; i++ {
		buffer.WriteRune(dna.BaseToRune(GetBase(fragment, i)))
	}
	return buffer.String()
}
```

PR #258, opened 2020-08-26
New log integrate

rimangan:


numbers/integrate.go
```diff
@@ -6,6 +6,27 @@ import (
...
+	var currLeft float64 = a //this variable stores the left bound of the current rectangle.
+	var currRight float64 = a + deltaX
+	var answer float64
+	//first time, sets answer as the area of the first rectangle
+	answer = MultiplyLog(MidpointLog(f(currLeft), f(currRight)), logDeltaX)
+
+	for i := 1; i < n; i++ {
+		currLeft += deltaX
+		currRight += deltaX
+		answer += MultiplyLog(MidpointLog(f(currLeft), f(currRight)), logDeltaX)
```
craiglowe, line 25: Not an error, but a chance to get a speed up.  I think you are evaluating f(x) twice as much as you need to because currRight of this time through the loop is currLeft of the next time through the loop.  If you do an initial evaluation of f(currLeft) you can then go through the loop only evaluating f(currRight) each time, then setting your variable for f(currLeft) to be f(currRight), so it is ready for the next time through the loop

numbers/logSpace.go
```diff
@@ -7,7 +7,23 @@ import (
 
 //these two values mark the overflow/underflow boundaries for math.Exp(x)
 func LogCanConvert(x float64) bool {
-	return x > 7.09782712893383973096e+02 || x < -7.45133219101941108420e+02
+	return x > 709.782712893383973096e+02 || x < -745.133219101941108420e+02
```
craiglowe, line 10: Is this change correct?  Those original values look correct to me.  Maybe you didn't notice the "e+02" at the end?
If I am thinking about this correctly, I think you actually want the negation of this: x < 700 && x > -700
I might be mixed up, but something to double-check

numbers/logSpace.go
```diff
@@ -7,7 +7,23 @@ import (
...
+}
+
+//AverageLog returns the average of a list of logSpace numbers
+func AverageLog(x []float64) float64 {
```
craiglowe, line 14: Can we have tests for AverageLog and MidpointLog?  I am not sure I saw them when looking quick

numbers/distribution_test.go
```diff
@@ -25,6 +26,14 @@ func TestBinomialDist(t *testing.T) {
 	}
 }
 
+func TestBinomialDistLog(t *testing.T) {
+	input1 := math.Exp(BinomialDistLog(20, 4, 0.6))
+	expected1 := 2.696862e-04
+	if fmt.Sprintf("%e", input1) != fmt.Sprintf("%e", expected1) {
```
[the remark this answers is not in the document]
rimangan: Yes, I was having some trouble comparing numbers due to rounding. Not sure if there is a more permanent solution for this but this seems to fix the problem

numbers/logSpace.go
rimangan: Oh yes, these were correct beforehand. I'll adjust

PR #260, opened 2020-08-26
simpleReader struct

edotau:


edotau:
the simple reader struct are only minor tweaks to fileio.EasyReader. they could very well be programmed into EasyReader, but I will leave that up for discussion.

PR #266, opened 2020-08-28
Simple reader

edotau:
fixing compile errors


dna/bytesDna.go
```diff
@@ -0,0 +1,61 @@
...
+		return C
+	case 'g':
+		return G
+	case 't':
+		return T
+	case 'n':
+		return N
+	case '-':
+		return Gap
+	//VCF uses star to denote a deleted allele
```
craiglowe, line 33: Should we add this elsewhere too?

dna/bytesDna.go
```diff
@@ -0,0 +1,61 @@
...
+
+func ByteSliceToDnaBases(b []byte) []Base {
+	var answer []Base = make([]Base, len(b))
+	for i, byteValue := range b {
+		answer[i] = ByteToBase(byteValue)
+	}
+	return answer
+}
+
+func ByteDnaBasesToString(dnaBases []Base) string {
```
craiglowe, line 52: Is there an existing DnaBasesToString that this should replace?

fileio/simpleio.go
```diff
@@ -27,28 +50,48 @@ func (reader *SimpleReader) Read(b []byte) (n int, err error) {
...
 	if err == nil {
-		if curr[len(curr)-1] == '\n' {
-			return curr[:len(curr)-1], false
+		if curr.Slice[len(curr.Slice)-1] == '\n' {
+			reader.buffer.Reset()
+			_, err = reader.buffer.Write(curr.Slice[:len(curr.Slice)-1])
+			common.ExitIfError(err)
+			return reader.buffer, false
 		} else {
 			log.Fatalf("Error: end of line did not end with a white space character...\n")
```
craiglowe, line 96: "white space" -> "new line"?

fileio/simpleio.go
```diff
@@ -27,28 +50,48 @@ func (reader *SimpleReader) Read(b []byte) (n int, err error) {
...
-// the reader encounters EOF.
-func ReadLine(reader *SimpleReader) ([]byte, bool) {
-	curr, err := reader.ReadBytes('\n')
+// ReadLine will return a bytes.Buffer pointing to the internal slice of bytes. Provided this function is called within a loop,
+// the function will read one line at a time, and return bool to continue reading. Important to note the buffer return points to
+// the internal slice belonging to the reader, meaning the slice will be overridden if the data is not copied. Please be aware the
+// reader will call close on the file once the reader encounters EOF.
+func ReadLine(reader *SimpleReader) (*bytes.Buffer, bool) {
+	var err error
+	curr := reader.pool.Get().(*Line)
```
craiglowe, line 85: It seems like this is used to allocate and reuse a single Line (aka byte slice) each time we call the function.  Is it simpler/faster to remove the complexity of the memory pool (where we never know how many objects will be allocated or freed) and just have the reader hold a slice of byte that is re-used?  Essentially, instead of having a pool that you give and take "curr" from.  You just have "curr" in the reader.  I think this works since you only need one.

giraf/giraf.go
```diff
@@ -17,6 +17,7 @@ type Giraf struct {
 	Flag      uint8
 	PosStrand bool
 	Path      *Path
+	Cigar     []cigar.ByteCigar
```
craiglowe, line 20: Does this hold the same info as Aln, minus the base sequence of inserted bases?  I think this is here to be more efficient.  Should we change cigar.Cigar to be more efficient instead of have duplicate copies?  Maybe we could have Cigar as []cigar.ByteCigar and then a slice of dna that is the list of inserted sequences, such that if the ByteCigar had two insertions (a 10bp insertion and then later on a 5bp insertion), there would be a single slice (of length 15) holding those bases, which could be left as nil if those were not included.

fileio/simpleio.go
edotau: That is a very good point, let me test this real quick

edotau: pool definitely help, what is unclear to me is if reusing a single Line (aka byte slice) actually helps.

dna/bytesDna.go
edotau: Yes i believe this can replace the current DnaBasesToString()

fileio/simpleio.go
edotau: be sure to checkout simpleio_test.go. I added what I thought was an interesting benchmark. If you are benchmarking purely runtime, using the pool is faster, but only by a few nanoseconds (not noticeably faster).
`BenchmarkSimpleReader-8    	       1	1062399970 ns/op`
`BenchmarkEasyReader-8      	       1	 1133461460 ns/op`
But if you look like the memory allocations on each line, pool wins by a landslide:
`BenchmarkSimpleReader-8    	       1	1062399970 ns/op	  291272 B/op	    4561 allocs/op`
`BenchmarkEasyReader-8      	       1	1133461460 ns/op	217882672 B/op	 3404566 allocs/op`

giraf/giraf.go
edotau: If it were up to me, I'd prefer to live in a world without fancy cigar all together (no []dna.Base whatsoever).

fileio/simpleio.go
craiglowe: I am not sure why the sync.pool would be faster on your machine.  Maybe something about the scheduling on processors is different on a mac?  This might be beyond my understanding of modern hardware.  It seems like it would be more time to keep putting the byte slice into the sync.pool and then asking for it back again.  I did an almost exact copy-and-paste of your simple reader, but without the sync.pool and just have a byte slice instead.  Because gz is so slow, everything benchmarks pretty similar on gzipped files, but I see a difference on my linux machine for uncompressed files.  Can you take a quick look at the benchmarks at the bottom of simpleio_test.go on branch: craigTestReader?

edotau: oh there is actually a huge difference. With your benchmark I actually see a big difference for gzip files as well. I'm trying to figure out how I had something different yesterday, the only difference I can think of is that i was using a pointer to a slice byte

edotau: the small modification you suggested actually cuts gzip reading in half as well. What I was trying to accomplish with the pool was I wanted to read the data, analyze it and/or write the results to file; put the buffer back before making a copy. I was trying to figure out a way to pass back a pointer to the underlying byte slice without actually copying what I was readying.

giraf/giraf.go
craiglowe: Yes, I am aware that you are not a fan of the fancy cigar.  I think we should still continue to explore compression techniques/encodings that will allow us to reduce the amount of data that needs to be stored.  One relationship that has held true in the past is that the cost of sequencing has decreased faster than I/O speeds have increased.  This has made compression increasingly important (e.g. bam -> cram).  While we can gain a lot from binary encoding of the data and general compression with gzip, minimizing the amount of data we need to store is another path forward (that can be combined with the other two paths).  Most reads will be perfect matches to the graph (increasingly so as the graph grows), so we don't need to store the sequence for them since the cigar implicitly encodes it.  The only exception to the cigar implicitly encoding the sequence is insertions, which is why if we can find a compromise to encode these inserted sequences, we can not store the actual sequence of the read.

DanielSnellings: I'll throw in my 2 cents here. I do quite like the fancy cigar, primarily for the enhanced compression that we can achieve as Craig mentions. That said, I agree with Eric that we should drop the Fancy Cigar in the human readable version of Giraf. During code review it seems that people were quite adamant that we keep the seq field in the human readable version which in my opinion undercuts the major utility of the fancy cigar. As-is we only stand to increase file size by incorporating the Fancy Cigar. I have had some discussions with Eric about this and my understanding is that we would like the human readable Giraf to have a sequence field and a semi-fancy byte cigar (encodes mismatches but not sequence). However, for the binary Giraf would would use a fancy cigar and a few other tricks to improve compression (read name prefixes + run-length encoding qual scores).

edotau:
I made one final push to update the benchmarks file. hopefully they correct now

giraf/giraf.go
edotau: i guess these discussions are unsettling to me because it feels like people think the alignment problem is solved. Even with my recent improvements, aligning reads to a full graph (stickleback) still proves to be difficult. I think it'll be hard to convince anyone to use the giraf format if we can't output an informative alignment.

craiglowe (approved):
Looks great!  Excited to have this checked in!

rimangan (approved):

DanielSnellings (approved):

PR #267, opened 2020-08-29
Deference pointer

edotau:


fastq/fastqBig.go
```diff
@@ -36,7 +38,45 @@ func ToFastqBig(a *Fastq) *FastqBig {
...
+	copy(answer.SeqRc, answer.Seq)
+	dna.ReverseComplement(answer.SeqRc)
+
+	//performs two bit conversion
+	answer.Rainbow = dnaTwoBit.TwoBitRainbowDeReference(answer.Seq)
+	answer.RainbowRc = dnaTwoBit.TwoBitRainbowDeReference(answer.SeqRc)
+
+	line, done = fileio.ReadLine(reader)
+	if done {
+		return nil, true
```
craiglowe, line 71: Should checks like this be an error instead of returning nil and true?  Seems like this would mean we were reading a fq file that had not been fully written.

fastq/phredQual.go
```diff
@@ -16,18 +18,31 @@ func ToQualUint8(qual []rune) []uint8 {
 	return answer
 }
 
+// ToQual will convert a slice of bytes to uint8 and subtracting a 33 offset to the values
+func ToQual(qual []byte) []uint8 {
+	var answer []uint8 = make([]uint8, len(qual))
+	for i := 0; i < len(qual); i++ {
+		answer[i] = uint8(qual[i]) - asciiOffset
```
craiglowe, line 25: I don't think you need this conversion.  I believe that "byte" is not an actual type by itself.  I believe it is just an alternative name that can be used for uint8.  I believe "bytes" is just for humans and the compiler will see it as "uint8".

fastq/phredQual.go
```diff
@@ -16,18 +18,31 @@ func ToQualUint8(qual []rune) []uint8 {
 	return answer
 }
 
+// ToQual will convert a slice of bytes to uint8 and subtracting a 33 offset to the values
+func ToQual(qual []byte) []uint8 {
+	var answer []uint8 = make([]uint8, len(qual))
```
craiglowe, line 23: I don't think you need this copy here if you don't want it.  Since byte and uint8 should be the same type, you can just loop through "qual" subtracting 33 from each value.  However, depending on how you use this, you may be expecting []byte to be untouched or maybe overwritten, in which case you will need the copy.

giraf/convert.go
```diff
@@ -131,3 +130,54 @@ func fromStringToQual(s string) []uint8 {
...
+
+func ToString(g *Giraf) string {
+	buf := &strings.Builder{}
+	buf.Grow(400)
+	var err error
+	_, err = buf.WriteString(g.QName)
+	common.ExitIfError(err)
+	err = buf.WriteByte('\t')
+	common.ExitIfError(err)
+	_, err = buf.WriteString(strconv.Itoa(g.QStart))
```
craiglowe, line 142: Is the style of processing these values one-at-a-time chosen because fmt.Sprintf is slower?

craiglowe (approved):
Looks great!  Few small comments

DanielSnellings (approved):

rimangan (approved):

craiglowe (approved):
Great work!

PR #268, opened 2020-08-29
Detect mismatches during gsw traceback and store in byte cigar (no seq field)

DanielSnellings:
This was forked from erics simpleReader branch to include his compile error fixes. This should wait to be merged until SimpleReader is merged. 

Removed the fancy cigar conversion in gsw in favor of a hybrid between ByteCigar and FancyCigar - i.e. detects mismatches but does not store the seq

Only files in which changes were made are:
simpleGraph/localAlignment.go
simpleGraph/globalAlignment.go
simpleGraph/toGiraf.go


simpleGraph/localAlignment.go
```diff
@@ -120,9 +122,13 @@ func SmithWaterman(alpha []dna.Base, beta []dna.Base, scores [][]int64, gapPen i
 	return m[maxI][maxJ], route, minI, maxI, minJ, maxJ
 }
 
-func tripleMaxTrace(a int64, b int64, c int64) (int64, rune) {
+func tripleMaxTrace(prev int64, a int64, b int64, c int64) (int64, rune) {
```
edotau, line 125: why do we need to pass in prev? i am having a hard time wrapping why head around that

DanielSnellings: So the way tripleMaxTrace determines if it is a match or a mismatch is by determining if the score goes up (match) or down (mismatch). In order to determine how the score changed, I need to know what the score was from the previous iteration.

rimangan (approved):

craiglowe (approved):
Great work!

edotau (approved):

PR #269, opened 2020-08-29
Gsw aligner update

edotau:


edotau:
This pull request is is the sum of the earlier two. If I try to separate anything else out it gets difficult to tell if stuff compiles.

fastq/pairedEnd.go
```diff
@@ -121,3 +121,22 @@ func GoWriteFqPair(readOne string, readTwo string, data <-chan *PairedEnd) {
 	go WritingChan(readOne, readTwo, data, &wg)
 	wg.Wait()
 }
+
+// ReadFqBigPair will take 2 readers which will convert paired end read fastq files into a paired end FastqBig struct.
+// Note: while this function is return as a pointer, it's purpose is to be deferenced at the next function call.
+// In additon the pointers to read one and read two will also be dereference. When sending a pointer to a struct through
+// a channel, or a struct with pointers inside, memory allocated will be placed on the heap hindering performance.
+func ReadFqBigPair(readerOne *fileio.SimpleReader, readerTwo *fileio.SimpleReader) (*PairedEndBig, bool) {
```
DanielSnellings, line 129: Does sending the value as a pointer the dereferencing it in the subsequent function actually help performance? If the value is returned as a pointer then it will go to the heap and dereferencing it won't move it to the stack right?

edotau: In short yes, this has a dramatic effect on performance only if you are passing a struct through a channel. This about optimizing what goes through the channel. However, the downside to reading non-pointers inside a loop, you can no longer check for nil to exit the loop, so a small trick you can do is to initialize a pointer check for nil, and dereference it before you send it through the channel.

edotau: Also, i worry that the message is starting to get cloudy. I don't want it to seem like ALL heap allocations are bad, that is a very narrow point of view. This Read function addresses hold **_long_** the memory stays on the heap, not necessarily decreasing the the heap allocation.

DanielSnellings: I see. Would this be a fair statement: "When pointers are returned from functions they can be efficiently garbage collected, but when sent through channels they cannot be garbage collected until the main function has returned"?

giraf/giraf.go
```diff
@@ -30,23 +32,28 @@ type Giraf struct {
...
 	TStart int      // The path starts on the TStart base (0-based, closed) of Nodes[0]
 	Nodes  []uint32 // The node Id/Index of all the nodes in the path
 	TEnd   int      // The path ends on the TEnd base (0-based, open) of Nodes[len(Nodes)-1]
 }
 
+// Note is a scruct containing additional information from the graph alignment
 type Note struct {
-	Tag   string
-	Type  rune
+	Tag   []byte
```
DanielSnellings, line 44: as long as you are modifying the tag, why not make it type `[2]byte` instead of a `[]byte`

edotau: Yes i am thinking about it. I see a few potential problems coming up once we move towards that. I had thought []byte was a good middle ground, but let me turn it over in my head.

fastq/pairedEnd.go
edotau: i am not sure you can put it that way. i am not saying you cannot send a pointer to a channel, but if you want to reduce your heap allocation is the easiest start.

Stack allocation requires that the lifetime and memory footprint of a variable can be determined at compile time. So when you send a pointer (or structs that contain pointers) to a channel, the GC has no idea which channel you sent the data to or when you are done with that pointer and will allocate it on the heap for the duration of the function.

simpleGraph/search.go
```diff
@@ -1,107 +1,637 @@
...
+	}
+
+	if len(answer) == 0 {
+		return []SeedDev{currPart}
+	} else {
 		return answer
+	}
+}
+
+func extendToTheLeftHelperDev(node *Node, read *fastq.FastqBig, nextPart SeedDev) []SeedDev {
```
craiglowe, line 489: Changes to this function, or a move?

simpleGraph/search.go
```diff
@@ -1,107 +1,637 @@
...
+func restartSeedHelper(helper *seedHelper) {
+	helper.currHits = helper.currHits[:0]
+	helper.keyIdx, helper.keyOffset, helper.readOffset, helper.nodeOffset = 0, 0, 0, 0
+	helper.nodeIdx, helper.nodePos = 0, 0
+	helper.seqKey, helper.codedNodeCoord = 0, 0
+	helper.leftMatches = 0
+}
+
+//seedBuildHelper.nodeIdx, seedBuildHelper.nodePos int64 = 0, 0
+func seedMapMemPool(seedHash map[uint64][]uint64, nodes []*Node, read *fastq.FastqBig, seedLen int, perfectScore int64, scoreMatrix [][]int64, finalSeeds []SeedDev, tempSeeds []SeedDev, seedBuildHelper *seedHelper) []SeedDev {
```
craiglowe, line 549: Are there changes beyond reusing memory.  It takes me a little time to reason through all the offsets.  I am not sure if there are changes there or not

craiglowe (commented):
Looks great!  Excited to see the speed up!

DanielSnellings (approved):

cfauci (approved):

craiglowe (approved):
Great work!

PR #271, opened 2020-09-04
gVCF Ancestor annotation tools

rimangan:
Several functions for identifying the ancestral state of a gVCF variant and the ability to annotate gVCF lines with this data and adjust the header.
Small miscellaneous change to the number of significant digits in the log integrate tests.


dna/compare.go
```diff
@@ -83,3 +83,16 @@ func CompareSeqsIgnoreCaseAndGaps(alpha []Base, beta []Base) int {
 func CompareSeqsCaseSensitiveIgnoreGaps(alpha []Base, beta []Base) int {
 	return compareSeqsIgnoreGaps(alpha, beta, false)
 }
+
+//SeqEqual returns true if two DNA sequences are identical, false otherwise.
+func SeqEqual(alpha []Base, beta []Base) bool {
```
edotau, line 88: I think this already exits as compareSeq()

edotau: There's also a few different options. See 
```
func CompareSeqsIgnoreCase(alpha []Base, beta []Base) int {
	return compareSeqs(alpha, beta, true)
}

func CompareSeqsCaseSensitive(alpha []Base, beta []Base) int {
	return compareSeqs(alpha, beta, false)
}

func CompareSeqsIgnoreCaseAndGaps(alpha []Base, beta []Base) int {
	return compareSeqsIgnoreGaps(alpha, beta, true)
}

func CompareSeqsCaseSensitiveIgnoreGaps(alpha []Base, beta []Base) int {
	return compareSeqsIgnoreGaps(alpha, beta, false)
}
```

rimangan: Ah that's helpful, I was confused by the int return but it makes sense now. I'll delete this redundancy.

DanielSnellings (approved):

edotau (approved):

cfauci (approved):

LCBartelt (approved):

PR #272, opened 2020-09-11
Binary Giraf v1

DanielSnellings:
First pass at binary giraf encoding. Several TODOs remain, but the core functionality works.


giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,269 @@
...
+		common.ExitIfError(err)
+		giraf.WriteGirafToFileHandle(outfile, &curr)
+	}
+
+	// Close reader
+	err = reader.bg.Close()
+	common.ExitIfError(err)
+}
+
+// The Read method for the BinWriter struct decompresses a single giraf record and writes to file
```
rimangan, line 60: Could you tell me a bit about this syntax with the type before the func name?

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,269 @@
...
+// DecompressGiraf will decode a binary giraf file (.giraf.fe) and output a giraf file (.giraf)
+func DecompressGiraf(filename string, graph *simpleGraph.SimpleGraph) {
+	// Initialize infile
+	infile := fileio.EasyOpen(filename)
+	defer infile.Close()
+	reader := NewBinReader(infile.BuffReader)
+	var err error
+
+	// Initialize outfile
+	outfile := fileio.EasyCreate(strings.TrimSuffix(filename, ".fe"))
```
rimangan, line 45: What is the significance of .fe for binary? Is this a type of compression, like gz or something else?

DanielSnellings: None whatsoever, this may change later. In general I was thinking it may be best to append a suffix to the filename like gzip instead of sam/bam style. '.fe' was only chosen for grins :)

giraf/binaryGiraf/binReader.go
DanielSnellings: This is the syntax for declaring type methods. i.e. the Read function is a method 'owned' by the BinWriter type. This style was based on the bgzf writers from biogo that I was looking at. Most readers and writers in the standard go library seem to declare read and write as methods, but I don't see why we couldn't move this to our standard procedural syntax if that is what people want.

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,269 @@
...
+
+// The BinReader struct wraps the bgzf reader from the biogo repository with a bytes buffer to store encoded giraf records
+type BinReader struct {
+	bg       *bgzf.Reader
+	currData *bytes.Buffer
+}
+
+// NewBinReader creates a new BinReader
+func NewBinReader(file io.Reader) *BinReader {
+	reader, err := bgzf.NewReader(file, 1) //TODO: Play with different levels of concurrency
```
edotau, line 28: My intuition tells me that this would not get that much faster when you increase this number since reading in a file will we bounded by IO not CPU. If i understand this correctly what I think ends up happening is you spend time setting up the goroutines and they sit around and don't have work to do because there are more CPU operations than IO processes. What you can do is have one thread help out with the decompression and the other thread read the file which yields to at most 2x speed up, but not much more.

My take on bgzf reader is that it implements really cool multi-threading functionality, but concurrency/ parallelism is only beneficial if your function takes longer than setting up your channels and goroutines and the calls to sync to make your processes thread safe. Base on my intuition, I think concurrency/ parallelism here would either slow down reading process (or remain the same in terms of time) and for sure increase the memory footprint because of the additional goroutines that are initialized only to sit around and do almost nothing.

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,269 @@
...
+	// blockSize (uint32)
+	bytesRead, err = br.bg.Read(buffer[:4])
+	if err == io.EOF || bytesRead != 4 {
+		return answer, io.EOF
+	}
+	common.ExitIfError(err)
+	blockSize := int(binary.LittleEndian.Uint32(buffer[:4]))
+
+	// fill currData buffer with blockSize of bytes
+	data := make([]byte, blockSize) // read giraf to intermediate buffer //TODO: this is not great, find a way to not make an intermediate buffer
```
edotau, line 80: One suggestion is to have this intermediate buffer as a struct field inside BinReader{}, so you can reuse the same intermediate buffer without having to declare `make([]byte, blockSize)` every single time. You just have to remember to clear it and set the blocksize at the beginning, which I think is a small price to pay compared to declaring `data := make([]byte, blockSize)` every time you call this function.

giraf/binaryGiraf/helper.go
```diff
@@ -0,0 +1,86 @@
...
+	return answer
+}
+
+// ReadToChan will input a fileio.EasyReader and decompress girafs in the file into a stream (channel) of giraf.Giraf.
+// ReadToChan is designed to be run as a goroutine. See GoReadToChan for internally handled goroutines.
+func ReadToChan(file *fileio.EasyReader, graph *simpleGraph.SimpleGraph, data chan<- giraf.Giraf, wg *sync.WaitGroup) {
+	var curr giraf.Giraf
+	var err error
+	reader := NewBinReader(file.BuffReader)
+	for curr, err = reader.Read(graph); err != io.EOF; curr, err = reader.Read(graph) {
```
edotau, line 40: I think is is slightly inefficient because you have to check err twice (once for io.EOF) then again for nil or not nil. In most cases this will be nil so depending on how you set up your checks, you can actually get away with checking the err only once. The speed up is not life changing, but something to keep in mind is that even though logic statements are equivalent, the order of operations does make a small difference on performance.

edotau (commented):
Please see QStart comment and the Read() method comment, I think those are the only major change request I have. Everything else I just gave my thoughts and opinions on the design choices which I think are debatable.

giraf/binaryGiraf/binReader.go
DanielSnellings: Good point. Since this function is not reading bytes (and therefore won't satisfy the io.Reader interface) it would be best to stick to procedural syntax. I will make this change.

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// qName (string)
+	answer.QName = string(br.currData.Next(qNameLen))
+
+	// flag (uint8)
+	answer.Flag = br.currData.Next(1)[0]
+
+	// tStart (uint32)
+	answer.QStart = int(binary.LittleEndian.Uint32(br.currData.Next(4)))
+	answer.Path.TStart = answer.QStart //TODO: Is this true?
```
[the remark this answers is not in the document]
DanielSnellings: Ah, in that case we should have a conversation about whether QStart and QEnd should be removed from the Giraf struct. Isn't this information covered in the cigar? Off the top of my head I don't really see many use cases where you would need to know the start and end of the aligned seq, but not the cigar. It may be worth dropping QStart and QEnd.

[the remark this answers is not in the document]
edotau: No

giraf/binaryGiraf/binReader.go
DanielSnellings: A good point. I have an implementation of this that should reduce slice declaration a lot. Ideally it would read a blockSize worth of bytes directly into the bytes.Buffer but I have not found a way to do that. Let me know if you have any ideas.

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,301 @@
...
+					currNodeId++
+					currNode = graph.Nodes[answer.Path.Nodes[currNodeId]]
+				}
+
+				answer.Seq = append(answer.Seq, currNode.Seq[refIdx])
+				refIdx++
+			}
+		case 'X':
+			answer.Seq = append(answer.Seq, fancyBases[:cigar.RunLen]...)
+			fancyBases = fancyBases[:cigar.RunLen]
```
craiglowe, line 201: I have somehow confused myself.  I was picturing this line having "cigar.RunLen:" instead of ":cigar.RunLen".  I was thinking that you were processing the first cigar.RunLen bases from fancyBases (on the previous line) and then removing those bases on this line.  Maybe a few comments would help me?

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,301 @@
...
+			log.Fatalf("ERROR: Unrecognized tag type: %s", string(currNote.Type))
+		}
+		answer.Notes = append(answer.Notes, currNote)
+	}
+}
+
+// determineQStartQEnd parses the cigar to determine where the alignment starts and ends
+func determineQStartQEnd(answer *giraf.Giraf) (int, int) {
+	var start, end int
+	if answer.Cigar == nil {
```
craiglowe, line 284: Should this be a log.Fatal?

giraf/binaryGiraf/binWriter.go
```diff
@@ -0,0 +1,271 @@
...
+
+// NewBinWriter creates a new BinWriter
+func NewBinWriter(file io.Writer) *BinWriter {
+	return &BinWriter{
+		bg: bgzf.NewWriter(file, 1), //TODO: Play with different levels of concurrency
+	}
+}
+
+// CompressGiraf will encode a giraf file (.giraf) and output a binary giraf file (.giraf.fe)
+func CompressGiraf(filename string) {
```
craiglowe, line 32: Maybe output filename as an option?

giraf/binaryGiraf/binWriter.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	// Close writer
+	err = writer.bg.Close()
+	common.ExitIfError(err)
+}
+
+// Byte size of BinGiraf fixed size fields excluding blockSize.
+// Exluded Fields: qName, path, byteCigar, fancySeq.Seq, qual, notes
+var binGirafFixedSize int = 29
```
craiglowe, line 52: Global variables make me nervous.  Can this be a "const"?

giraf/binaryGiraf/binWriter.go
```diff
@@ -0,0 +1,271 @@
...
+	binary.LittleEndian.PutUint32(currBuf[:4], uint32(fancySeq.Len))
+	bw.buf.Write(currBuf[:4]) // fancySeq.Len (uint32)
+
+	for _, val := range fancySeq.Seq {
+		binary.LittleEndian.PutUint64(currBuf[:8], val) // TODO: does not need to be uint64?
+		bw.buf.Write(currBuf[:8])                       // fancySeq.Seq (uint64)
+	}
+
+	// alnScore (int64)
+	binary.LittleEndian.PutUint64(currBuf[:8], uint64(g.AlnScore))
```
craiglowe, line 121: Do you need the reslicing?  Just curious.

giraf/binaryGiraf/binary.go
```diff
@@ -0,0 +1,50 @@
...
+	textLen        uint32   // length of below text field
+	text           string   // plain header text in giraf file
+	qNamePrefixLen uint32   // length of below qNamePrefix field
+	qNamePrefix    string   // prefix of read name, may be updated to algorithmic compression of read names
+}
+
+type binGiraf struct {
+	blockSize uint32 // total length of the alignment record, excluding this field
+	// THIS HAS BEEN MOVED TO FLAG //hasNamePrefix bool   // bool to determine whether qNamePrefix is valid for this read
+	qNameLen uint8  // length of below qName field
```
craiglowe, line 26: I think limiting this to 256 characters makes sense, but do we have a check for this to give an error?

giraf/binaryGiraf/binary.go
```diff
@@ -0,0 +1,50 @@
...
+	qName    string // name of read, qNamePrefix is prepended if hasNamePrefix
+	// QStart and QEnd are determined by clipped bases in Cigar
+	flag uint8 // bitwise flags
+	// PosStrand is dropped, stored in flag
+	tStart      uint32               // start position on first node (nodes[0]) in path, zero base
+	tEnd        uint32               // end position on last node (nodes[len(nodes)-1]) in path, zero base
+	pathLen     uint16               // length of below path
+	path        []uint32             // list of nodes in alignment, len of path is = pathLen
+	numCigarOps uint16               // number of cigar operations
```
craiglowe, line 35: Too small for chromosome to chromosome alignments?

craiglowe (commented):
Very cool!  Lots of great work here!  It is exciting to see this.

giraf/binaryGiraf/binary.go
DanielSnellings: added a check to writer

giraf/binaryGiraf/binary.go
DanielSnellings: moved to uint32

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,301 @@
...
+	var answer giraf.Giraf
+	var bytesRead int
+	var err error
+	var i, j, runLen uint16
+	var buffer [4]byte
+
+	// blockSize (uint32)
+	bytesRead, err = br.bg.Read(buffer[:4])
+	if err == io.EOF || bytesRead != 4 {
+		return answer, io.EOF
```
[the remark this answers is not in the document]
craiglowe: ok, I had not realized that all that header info had already been handled.  However, I am glad that my incorrect comment seems to have lead to an unrelated improvement.

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,301 @@
...
+	// read block to readBffr and fill currData buffer with read bytes
+	bytesRead, err = io.ReadFull(br.bg, br.readBffr)
+	if err == io.ErrUnexpectedEOF || bytesRead != blockSize {
+		log.Fatalln("ERROR: truncated record")
+	}
+	common.ExitIfError(err)
+	br.currData.Write(br.readBffr) // copy data from intermediate buffer to bytes buffer
+
+	// qNameLen (uint8)
+	qNameLen := int(br.currData.Next(1)[0])
```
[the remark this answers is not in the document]
craiglowe: ok, I understand.

giraf/binaryGiraf/binReader.go
```diff
@@ -0,0 +1,301 @@
...
+	for i = 0; i < pathLen; i++ { // path ([]uint32)
+		node = binary.LittleEndian.Uint32(br.currData.Next(4))
+		answer.Path.Nodes = append(answer.Path.Nodes, node)
+	}
+
+	// cigar ([]cigar.ByteCigar)
+	numCigarOps := binary.LittleEndian.Uint16(br.currData.Next(2)) // numCigarOps (uint16)
+	for i = 0; i < numCigarOps; i++ {
+		answer.Cigar = append(answer.Cigar, cigar.ByteCigar{
+			RunLen: binary.LittleEndian.Uint16(br.currData.Next(2)), // byteCigar.RunLen (uint16)
```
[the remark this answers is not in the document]
craiglowe: ok, for another time then.  I want to keep this in my mind though because we have some chromosome to chromosome alignments planned.

giraf/binaryGiraf/binReader.go
craiglowe: ah yes, I always forget about unaligned reads

giraf/binaryGiraf/binWriter.go
craiglowe: Is it faster to make it an array that you re-slice instead of having it be a slice of length 8?

craiglowe (approved):
Looks great!  I included some responses to comments

rimangan (approved):

cfauci (approved):

PR #274, opened 2020-09-12
Binomial logSpace optimizations

rimangan:
In MCMC, we integrate over functions with a binomial component, which include a binomial coefficient (n choose k) that remains constant over the range of the function. In the old version, the coefficient gets calculated each time. This is a complicated calculation, so the program was rather inefficient, dedicating a lot of time to this process.
In this new version, all of the binomial coefficients for a particular n value are calculated up front and stored in memory, dramatically speeding up the program. 
Passes tests.


rimangan:
I also did a go fmt so a bunch of files came along with like one or two lines changed from small formatting issues

numbers/distribution_test.go
```diff
@@ -26,14 +25,6 @@ func TestBinomialDist(t *testing.T) {
 	}
 }
 
-func TestBinomialDistLog(t *testing.T) {
```
rimangan, line 29: This test has been moved to logDistribution_test.go and also modified.

cmd/selection_MCMC/selection_MCMC.go
```diff
@@ -8,9 +8,9 @@ import (
 	"log"
 )
 
-func selectionMCMC(filename string, outfile string, muZero float64, sigmaZero float64, iterations int, logSpace bool) {
+func selectionMCMC(filename string, outfile string, muZero float64, sigmaZero float64, iterations int) {
```
craiglowe, line 11: Very minor, but we can change this to selectionMcmc?

cmd/simulateBed/simulateBed.go
```diff
@@ -6,9 +6,20 @@ import (
 	"github.com/vertgenlab/gonomics/bed"
 	"github.com/vertgenlab/gonomics/common"
 	"log"
+	"math/rand"
+	"time"
 )
 
-func simulateBed(regionCount int, simLength int64, noGapFile string, outFile string) {
+func simulateBed(regionCount int, simLength int64, noGapFile string, outFile string, randSeed bool, setSeed int64) {
```
craiglowe, line 13: Not a problem for now, but in the long-term this should probably be a separate pull request.

numbers/logDistribution.go
```diff
@@ -9,3 +14,28 @@ func BinomialDistLog(n int, k int, p float64) float64 {
 	expression := MultiplyLog(s, f)
 	return MultiplyLog(coefficient, expression)
 }
+
+//BinomialDistLogMap returns log(BinomialDist), where log is the natural logarithm.
+//This function is similar to BinomialDistLog but passes in a map[int][]float64, where the int key
+//refers to n and the []float64 map values are the corresponding binomial coefficients for index k.
+//Useful to not recalculate the binomial coefficient each time when binomial densities must be constantly evaluated in logSpace, like in MCMC.
+func BinomialDistLogMap(n int, k int, p float64, binomMap *map[int][]float64) float64 {
```
craiglowe, line 22: Minor, but would it be better to have this be a slice instead of a map?  The indices will be consecutive integers from zero to some number, which is a good fit for a slice.  It should be both less memory and faster.

numbers/logDistribution.go
```diff
@@ -9,3 +14,28 @@ func BinomialDistLog(n int, k int, p float64) float64 {
...
+	if _, ok := (*binomMap)[n]; !ok {
+		(*binomMap)[n] = AddBinomMapEntry(n)
```
craiglowe, line 24: I was wondering if we should have checks to make sure k, n, and p are allowable values, but I am thinking that is all taken care of because all calls will need to go through BinomCoefficientLog eventually.

popgen/MCMC_Selection.go
```diff
@@ -82,15 +84,25 @@ func InitializeTheta(m float64, s float64, k int) Theta {
 
 //MetropolisHastings implements the MH algorithm for Markov Chain Monte Carlo approximation of the posterior distribution for selection based on an input allele frequency spectrum.
 //muZero and sigmaZero represent the starting hyperparameter values.
-func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations int, logSpace bool) ([]float64, []float64, []bool) {
+func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations int) ([]float64, []float64, []bool) {
+	//profiling test code
+	f, err := os.Create("testProfile.prof")
```
craiglowe, line 89: This should probably be commented out in the "main" branch with your "DEBUG" syntax.

popgen/MCMC_Selection.go
```diff
@@ -82,15 +84,25 @@ func InitializeTheta(m float64, s float64, k int) Theta {
...
+
 	muList := make([]float64, iterations)
 	sigmaList := make([]float64, iterations)
 	acceptList := make([]bool, iterations)
 	//initialization to uninformative standard normal
 	t := InitializeTheta(muZero, sigmaZero, len(data.sites))
 	for i := 0; i < iterations; i++ {
 		tCandidate := GenerateCandidateThetaPrime(t)
-		if Metropolis_Accept(t, tCandidate, data, logSpace) {
+		if Metropolis_Accept(t, tCandidate, data, &binomMap) {
```
craiglowe, line 105: I am interested in the choice of using a pointer to a map.  I think that maps (like slices) have their underlying data passed by reference, with only a header-like collection of data being copied.  I am not asking for a change, just interested to hear your thoughts.

craiglowe (commented):
Looks wonderful!  A few comments here and there, but I am excited to get this checked in!

cfauci (approved):

craiglowe (approved):
Great Work!

LCBartelt (approved):

PR #276, opened 2020-09-14
Simulate cmd

cfauci:
Made changes to simulate.go to handle fasta assignment at nodes differently and created simulateEvol cmd to create a whole tree fasta file and a leaves-only fasta file using simulate package.


cmd/simulateEvol/simulateEvol.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/vertgenlab/gonomics/simulate"
+	"log"
+)
+
+func SimulateEvol(rootFastaFile string, treeFile string, simOutFile string, leafOutFile string) {
+	tree := expandedTree.ReadTree(treeFile, rootFastaFile)
+	var fastas []*fasta.Fasta
+	simulate.Simulate(rootFastaFile, tree)
+	nodes := expandedTree.GetTree(tree)
+	fmt.Printf("%s, %s, %s, \n", rootFastaFile, treeFile, simOutFile)
```
edotau, line 17: This is ultimately up to you, but I would consider using log.Printf() here only because I think if someone would try to pipe this to stdout it will be included

edotau (approved):
Just had a quick suggestion for the fmt.Printf(), and other than that nice work! I get excited every time you get stuff checked in because I want that flying chicken.

cmd/simulateEvol/simulateEvol.go
```diff
@@ -0,0 +1,56 @@
...
+		"simulateEvol takes in a root fasta and a newick formatted tree with branch lengths and simulates evolution along the tree. It returns a list of fastas for the whole tree for reference and a list of fastas from leaves for reconstruction.\n" +
+			"Usage:\n" +
+			"simulateEvol <rootFasta.fasta> <newickTree.txt> <outFile.fasta>\n" +
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var expectedNumArgs = 4
+
```
rimangan, line 39: TODO: RandSeed and SetSeed options would be appropriate here

cmd/simulateEvol/simulateEvol.go
```diff
@@ -0,0 +1,56 @@
...
+func SimulateEvol(rootFastaFile string, treeFile string, simOutFile string, leafOutFile string) {
+	tree := expandedTree.ReadTree(treeFile, rootFastaFile)
+	var fastas []*fasta.Fasta
+	simulate.Simulate(rootFastaFile, tree)
+	nodes := expandedTree.GetTree(tree)
+
+	for i := 0; i < len(nodes); i++ {
+		fastas = append(fastas, nodes[i].Fasta)
+		if nodes[i].Fasta != nil {
```
rimangan, line 20: looks like there is something missing here

cmd/simulateEvol/simulateEvol.go
```diff
@@ -0,0 +1,56 @@
...
+	fasta.Write(simOutFile, fastas)
+
+	simulate.RemoveAncestors(simOutFile, tree, leafOutFile)
+}
+
+func usage() {
+	fmt.Print(
+		"simulateEvol takes in a root fasta and a newick formatted tree with branch lengths and simulates evolution along the tree. It returns a list of fastas for the whole tree for reference and a list of fastas from leaves for reconstruction.\n" +
+			"Usage:\n" +
+			"simulateEvol <rootFasta.fasta> <newickTree.txt> <outFile.fasta>\n" +
```
rimangan, line 32: This usage shows only three arguments, cmd takes 4

cmd/simulateEvol/simulateEvol.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/vertgenlab/gonomics/expandedTree"
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/simulate"
+	"log"
+)
+
+func SimulateEvol(rootFastaFile string, treeFile string, simOutFile string, leafOutFile string) {
+	tree := expandedTree.ReadTree(treeFile, rootFastaFile)
+	var fastas []*fasta.Fasta
+	simulate.Simulate(rootFastaFile, tree)
```
craiglowe, line 15: Not to do now, but I think it might make sense in the long-term to bring simulate and extendedTree into the reconstruct package since they really cater to that code.  Like I said, not now, longer term idea.

cmd/simulateEvol/simulateEvol.go
craiglowe: ... or maybe this if statement is no longer needed?

craiglowe (commented):
Looks great.  Riley had a few comments that may be quick to address and would be good to have before merging in.

craiglowe (approved):
I would make a node of Riley's comment that his set-seed setup could be used in this program as well.  I think that is a good idea.

edotau (approved):

rimangan (approved):

DanielSnellings (approved):

PR #277, opened 2020-09-15
Updated binomial calculation to slice instead of map

rimangan:
Optimization following Craig's comments yestereday.


rimangan:
Also Craig I put a smalltest.vcf in the selection_MCMC cmd folder. This is the file I've been doing all of my testing on, so it would be useful for both of us to have access to this.

edotau (approved):

DanielSnellings (approved):

edotau (approved):

craiglowe (approved):
Looks great!

PR #278, opened 2020-09-16
quickPrimateRecon fixes, reorganization, and testing

rimangan:
quickPrimateRecon, which performs estimation of the HCA from a five way alignment of primates, was reorganized, with its internal functions moved from the cmd itself to the reconstruct package. Some edits were made to how estimations of messy bases were made, with the messyToN argument added. Also, I have written a testing function for it in case any future updates break the existing functionality.

### DIFF
--- a/cmd/quickPrimateRecon/quickPrimateRecon.go
+++ b/cmd/quickPrimateRecon/quickPrimateRecon.go
@@ -3,108 +3,18 @@ package main
 import (
 	"flag"
 	"fmt"
-	"github.com/vertgenlab/gonomics/dna"
 	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/reconstruct"
 	"log"
 )
 
-func quickPrimateRecon(infile string, outfile string) {
+func quickPrimateRecon(infile string, outfile string, messyToN bool) {
 	records := fasta.Read(infile)
-	outputRecord := &fasta.Fasta{Name: "Human_Chimp_Ancestor"}
-
-	var outputBase dna.Base
-
-	if len(records) != 5 {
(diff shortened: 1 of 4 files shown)


edotau (approved):

cmd/quickPrimateRecon/quickPrimateRecon.go
```diff
@@ -130,5 +41,5 @@ func main() {
 	inFile := flag.Arg(0)
 	outFile := flag.Arg(1)
 
-	quickPrimateRecon(inFile, outFile)
+	quickPrimateRecon(inFile, outFile, *messyToN)
```
DanielSnellings, line 44: It seems like this function is expecting the infile to have primate sequences in a specific order based on how they are being assigned to variables (looks like human -> bonobo -> chimp -> orangutan -> gorilla). If this is true, it might be worth putting this in the usage message.

craiglowe: This is a good point.  Not needed for this checkin, but maybe a more robust way would be to pass in a map of fasta sequences keyed on the name of the fasta record that is expected to be certain things "human", "chimp", etc?

reconstruct/primateRecon_test.go
```diff
@@ -0,0 +1,30 @@
+package reconstruct
```
craiglowe, line 1: Great idea to put this in reconstruct and have some tests.  Wonderful!

craiglowe (approved):
Looks great!  Might be good to think about how the code expects a certain order that could lead to some bugs that are hard to detect.  Maybe a map key'ed on the seq name would be more robust?

DanielSnellings (approved):

cfauci (approved):

PR #281, opened 2020-09-17
panic instead of fatal

DanielSnellings:
When I am debugging, the 'StringToX' family of functions in common are particularly annoying as they are used in many core functions throughout the codebase, yet give uninformative error messages (e.g. 'trouble converting  to a int'). Since the functions are so commonly used it requires some effort to figure out the parent function where these errors are occuring. I propose that instead of using log.Fatal for these functions we instead panic on an error so that we get a stack trace.


cfauci (dismissed):

craiglowe (commented):
If it helps you, I am happy to make the change if there are no objections at code review.

cfauci (dismissed):

rimangan (approved):

craiglowe (approved):
Nice!

cfauci (approved):

rimangan (approved):

PR #282, opened 2020-09-21
A Nefarious Plot

rimangan:
A small cmd for plotting functions (returns a tab separated value file which can be fed into R or excel for data visualization). Current test example is for the AFS stationarity function for MCMC, which I wanted to plot the behavior of for further validation.

### DIFF
--- a/cmd/plotFunctions/plotFunctions.go
+++ b/cmd/plotFunctions/plotFunctions.go
@@ -23,7 +23,7 @@ func usage() {
 		"plotFunctions-returns a tab separated list of function evaluations for plotting functions.\n" +
 		"To specify the function, use a function keyword. Currently only 'AfsStationarity' is supported.\n" + 
 			"Usage:\n" +
-			" plotFunctions functionKeyWord leftBound rightBound stepSize outFile\n" +
+			" plotFunctions functionKeyWord leftBound rightBound steps outFile\n" +
 			"options:\n")
 	flag.PrintDefaults()
 }
--- a/cmd/plotFunctions/plotFunctions.go
+++ b/cmd/plotFunctions/plotFunctions.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"github.com/vertgenlab/gonomics/numbers"
+	"github.com/vertgenlab/gonomics/popgen"
+	"github.com/vertgenlab/gonomics/common"
+	"log"
+	"fmt"
+	"flag"
+	"math"
+)
+
+func plotFunctions(function string, left float64, right float64, bins int, outFile string, alpha float64) {
+	if function == "AfsStationarity" {
+		numbers.Plot(popgen.AFSStationarityClosure(alpha), left, right, bins, outFile)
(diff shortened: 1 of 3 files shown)


cmd/plotFunctions/plotFunctions.go
```diff
@@ -0,0 +1,51 @@
...
+	if function == "AfsStationarity" {
+		numbers.Plot(popgen.AFSStationarityClosure(alpha), left, right, bins, outFile)
+	} else {//here you can add more else ifs to add additional functions for plotting
+		fmt.Printf("Unrecognized function: %s.\n", function)
+	}
+}
+
+func usage() {
+	fmt.Print(
+		"plotFunctions-returns a tab separated list of function evaluations for plotting functions.\n" +
```
edotau, line 23: did you mean to include a couple spaces around the '-'?
i was thinking something like fmt.Print("plotFunctions - returns ...") looks a little nicer aesthetically, but if that is on purpose, then scratch this comment.

edotau (approved):
great!

cmd/plotFunctions/plotFunctions.go
```diff
@@ -0,0 +1,51 @@
...
+		"To specify the function, use a function keyword. Currently only 'AfsStationarity' is supported.\n" + 
+			"Usage:\n" +
+			" plotFunctions functionKeyWord leftBound rightBound numberOfBins outFile\n" +
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var expectedNumArgs int = 5
+	var alpha *float64 = flag.Float64("alpha", math.Inf(-1), "Specifies the strength of selection for AfsStationarity plots.")
```
craiglowe, line 33: Should the default value be zero (neutral)?  I was thinking that -Infinity is probably outside of what is realistic for alpha?  Is alpha = 2Nes?

cmd/plotFunctions/plotFunctions.go
```diff
@@ -0,0 +1,51 @@
...
+		fmt.Printf("Unrecognized function: %s.\n", function)
+	}
+}
+
+func usage() {
+	fmt.Print(
+		"plotFunctions-returns a tab separated list of function evaluations for plotting functions.\n" +
+		"To specify the function, use a function keyword. Currently only 'AfsStationarity' is supported.\n" + 
+			"Usage:\n" +
+			" plotFunctions functionKeyWord leftBound rightBound numberOfBins outFile\n" +
```
craiglowe, line 26: Is "bins" what we want?  From a quick scan of the code, maybe "stepSize" or "pointsToPlot" would be better?  I think of "bin" as summarizing data over an interval (the width of the bin"), but I think this program is giving me the function value at certain points e.g. f(leftBound), f(leftBound+step), etc

craiglowe (dismissed):
Couple of comments on phrasing the help message and default value.  Looks great!

cfauci (approved):

craiglowe (approved):
Great Work!

Krista-Pipho (approved):

PR #283, opened 2020-09-22
Stationarity distribution sampler

rimangan:
Functionality to sample from stationarity distributions has been added into statCalc. This uses an adapted rejection sampling method that I will discuss in more detail at code review. No formal tests in gonomics, but I have slides to show the outputs.

### DIFF
--- a/numbers/Monte_Carlo.go
+++ b/numbers/Monte_Carlo.go
@@ -61,7 +61,7 @@ func chooseBin(sumHeights float64, binHeights []float64) int {
 			return i
 		}
 	}
-	log.Fatalf("Error in chooseBin: failed to choose a bin, stiffled by indecision.")
+	log.Fatalf("Error in chooseBin: failed to choose a bin, stifled by indecision.")
 	return -1
 }
 
--- a/numbers/Monte_Carlo.go
+++ b/numbers/Monte_Carlo.go
@@ -51,16 +51,29 @@ func FastRejectionSampler(xLeft float64, xRight float64, f func(float64) float64
 		}
 		sumHeights += binHeights[i]
 	}
-	var currBin int
 	for j := 0; j < samples; j++ {
-		currBin = chooseBin(sumHeights, binHeights)
-		currLeft = xLeft + float64(currBin)*stepSize
-		currRight = currLeft + stepSize
-		//DEBUG: fmt.Printf("lenBinHeights: %v. currBin: %v. lenAnswer: %v. j: %v.\n", len(binHeights), currBin, len(answer), j)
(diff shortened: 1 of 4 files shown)


edotau (dismissed):

numbers/Monte_Carlo.go
```diff
@@ -1,23 +1,80 @@
...
+	
+	var support float64 = xRight -xLeft
+	var stepSize float64 = support / float64(bins)
+	var binHeights []float64 = make([]float64, bins)
+	var currLeft float64 = xLeft
+	var currRight float64 = xLeft + stepSize
+	var sumHeights float64
+
+	for i := 0; i < bins; i++ {
+		binHeights[i] = common.MaxFloat64(f(currLeft), f(currRight))
```
craiglowe, line 38: This will do about twice as many function evaluations as you need since for most iterations currLeft is last iterations currRight.  Don't need to change for now, but maybe a TODO?

numbers/Monte_Carlo.go
```diff
@@ -1,23 +1,80 @@
...
+		currLeft += stepSize
+		currRight += stepSize
+		sumHeights += binHeights[i]
+	}
+	var currBin int
+	for j := 0; j < samples; j++ {
+		currBin = chooseBin(sumHeights, binHeights)
```
craiglowe, line 45: I think chooseBin needs to be inside of the rejection sampling so that it gets chosen again if the x,y pair is rejected.  It should be that b,x,y is chosen and if it fails, new values for all three are chosen.

craiglowe (commented):
Looks great!  One comment about re-choosing bin when sampling fails.

edotau (approved):

numbers/Monte_Carlo.go
```diff
@@ -1,23 +1,104 @@
...
+			return i
+		}
+	}
+	log.Fatalf("Error in chooseBin: failed to choose a bin.")
+	return -1
+}
+
 //RejectionSample returns simulated values from an arbitrary function between a specified left and right bound using a simple rejection sampling method.
-func RejectionSample(xleft float64, xright float64, f func(float64) float64, maxIteration int) float64 {
+func RejectionSample(xLeft float64, xRight float64, yMax float64, f func(float64) float64, maxIteration int, returnX bool) float64 {
```
craiglowe, line 93: Another option would be for the function to return an (x, y) pair.  If the calling function only wanted one they could use an underscore to get rid of the other.  This would also let calling functions get both if they ever wanted that.

craiglowe (commented):
Looks wonderful!  I think I found one bug/typo.

cfauci (approved):

craiglowe (approved):
Great work!

DanielSnellings (approved):

PR #284, opened 2020-09-24
Recon accuracy cmd

cfauci:
reconAccuracy command determines the number of incorrectly called bases returned by reconstruct as compared to the tree created in simulate. Updates made to reconstruct and reconstruct_test to handle new functions and function changes.


cmd/reconAccuracy/reconAccuracy.go
```diff
@@ -0,0 +1,81 @@
...
+	allNodes = "all Nodes"
+	var found bool = false
+	tot := 0.0
+	sim := fasta.Read(simFilename)
+	recon := fasta.Read(reconFilename)
+
+	answer := make(map[string]float64)
+
+	for i := 0; i < len(sim); i++ {
+		num := 0.0
```
craiglowe, line 24: I would declare "num" once at the top of the function "var num float64" and then just initialize it to zero here "num = 0".  This way it keeps being declared over and over (in theory making this faster although probably not noticable) and it can also lead to some tricky bugs if there is another "num" in a different scope (these can be a nightmare to find), or if you try to access this after the loop.

cmd/reconAccuracy/reconAccuracy.go
```diff
@@ -0,0 +1,81 @@
...
+		found = false
+		for j := 0; j < len(recon); j++ {
+			if sim[i].Name == recon[j].Name {
+				found = true
+				//DEBUG: log.Printf("\n%s \n%s \n", dna.BasesToString(sim[i].Seq), dna.BasesToString(recon[j].Seq))
+				for k := 0; k < len(sim[0].Seq); k++ {
+					if sim[i].Seq[k] != recon[j].Seq[k] {
+						num = num + 1
```
craiglowe, line 32: maybe call this variable something like: "mistakes", "mismatches"?

cmd/reconAccuracy/reconAccuracy.go
```diff
@@ -0,0 +1,81 @@
...
+					}
+				}
+			}
+		}
+		if found == false {
+			log.Fatal("Did not find all simulated sequences in reconstructed fasta.")
+		} else {
```
craiglowe, line 39: Not a logical change, but might be easier to read if statements in "else" just come after the "if" closes.  The statements in the "else" block would join the "tot = tot + num" line in coming after the "if" statement

cmd/reconAccuracy/reconAccuracy.go
```diff
@@ -0,0 +1,81 @@
...
+			answer[sim[i].Name] = acc
+		}
+		tot = tot + num
+	}
+	accuracy := tot / (float64(len(sim)) * float64(len(sim[0].Seq))) * 100.0
+	//DEBUG: fmt.Printf("tot: %f, len(sim): %f, len(sim[0].Seq): %f \n", tot, float64(len(sim)), float64(len(sim[0].Seq)))
+	acc := 100 - accuracy
+	answer[allNodes] = acc
+	for name, accuracy := range answer {
+		log.Printf("%s %f \n", name, accuracy)
```
craiglowe, line 52: Not needed for now, but think about printing this info to a file.  You can use fileio.EasyCreate to open the file.  Then use fmt.Fprintf to write to the file instead of log.Printf

reconstruct/reconstruct.go
```diff
@@ -3,39 +3,73 @@ package reconstruct
...
-//		}
-//		tot = tot + num
-//	}
-//	accuracy := tot / (float64(len(sim)) * float64(len(sim[0].Seq))) * 100.0
-//	acc := 100 - accuracy
-//	fmt.Print("accuracy over all nodes= ", acc, "%", "\n")
-//	return acc
-//}
+//returns the percentage accuracy by base returned by reconstruct of each node and of all nodes combined (usage in reconstruct_test.go)
+func ReconAccuracy(simFilename string, reconFilename string) map[string]float64 {
```
craiglowe, line 11: This looks like the same/similar function to the one in the command.  I think you will want to remove the one in the command (see if my comments from that one apply to this one) and then in the command you would just call this function and then loop through the map you get back to print the accuracies (you already have that loop at the end of that function).

craiglowe (commented):
Great work!  I left some comments for you to think about.  Biggest one would be to consider removing some duplicate code from the command function and calling the similar code in the reconstruct.go file from the command.

cmd/reconAccuracy/reconAccuracy.go
```diff
@@ -0,0 +1,50 @@
...
+//TODO: this code will need to change drastically for sequences of varying lengths.
+//The loop through the sequence is restricted by a single fasta and the tot calculation will need to calculate the total number of bps
+//ReconAccuracy calculates the total number of incorrectly reconstructed base pairs in a tree and returns a percentage of correct base calls
+func ReconAccuracy(simFilename string, reconFilename string, outFilename string) {
+	answer := reconstruct.ReconAccuracy(simFilename, reconFilename)
+	out := fileio.EasyCreate(outFilename)
+	defer out.Close()
+
+	for name, accuracy := range answer {
+		fmt.Fprintf(out, "%s %f \n", name, accuracy)
```
craiglowe, line 20: Can you remove the space after the number and before the end of line?  Obviously not a huge deal, but some parsing functions will get confused and think there are three columns instead of two since they are two delimiters present (spaces) instead of only one.  The format string would be: "%s %f\n".  Another thing to consider is using a tab instead of a space.  In my experience this can make things a little easier since many of the Unix/Linux tools default to tab-separated columns.  \t is the magic <tab> character.  Kind of similar to \n.  The format string would be "%s\t%f\n"

craiglowe (approved):
Looks great!  I had one comment about the output string format.

craiglowe (approved):
Great work!

DanielSnellings (approved):

rimangan (approved):

PR #291, opened 2020-09-28
only go fmt

craiglowe:
Should be no changes to calculations here.  Only "go fmt" formatting changes that slipped through.  Hopefully this makes future checkins quicker since these changes can be browsed quickly now, but sometimes need to be stared at longer in a normal check-in to wonder why the change was made.

### DIFF
--- a/axt/axt_test.go
+++ b/axt/axt_test.go
@@ -44,6 +44,7 @@ func TestWriteAndRead(t *testing.T) {
 		}
 	}
 }
+
 //TODO: Finish rev. comp sequences for testing swap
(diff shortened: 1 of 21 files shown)


cshoben:
Hey, sorry I think I'm doing things on accident. I was trying to approve and somehow was given the option to close this pull request. I'll reopen now.

cshoben (approved):
I think I figured out how to approve!

rimangan (approved):

DanielSnellings (approved):

PR #294, opened 2020-10-02
moved math stuff from common to numbers

craiglowe:
Moving math from common to numbers


numbers/minMax.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func MinFloat64(a float64, b float64) float64 {
+	if a <= b {
+		return a
+	} else {
+		return b
+	}
+}
+
```
rimangan, line 82: Are these triple functions commonly used? Would it be better to just have a function for a slice of ints? Not sure what is best stylistically moving forward

rimangan (approved):
Looks great! This is an important fix!

numbers/minMax.go
edotau: I am against using a slice for this function. You would have to start looping through the slice and you have to allocate additional memory to keep track of the the Min/Max. Using individual numbers also ensures you allocate memory on the stack (at least that is my numberstanding).

edotau (approved):

numbers/minMax.go
edotau: No, i am against using slices for this. You also have to program a loop and also have to allocate additional memory to keep track of the current max/min, whereas here no additonal memory is needed to return an answer. When you work with simple data types (like ints and strings) memory allocation is more likely on the stack. When you start to use more slices, maps or complex structs, if you are sloppy or not careful, the memory escapes to the heap (at least that is my understanding).

craiglowe: I think it makes sense to keep these functions for the reasons Eric gave.  However, in the future we could add versions of min/max that operate on a slice if that would be useful for people.

edotau: The slice version of min/max essentially would look the same as a sort, either ascending or descending order, returning either the head or tail, is that right?

DanielSnellings: Not necessarily. A min/max would require far fewer comparisons than a true sort.

DanielSnellings (approved):

PR #297, opened 2020-10-06
Craig aln trim

craiglowe:
I am planning to take some time to write benchmarks for parts of the aligner and cases where it does not perform well so we can track improvements.  Before doing this, I am removing some old code that I don't think is used anymore.  It is not covered by tests and doesn't seem like it is up-to-date on first glance.  This will always be in older versions of main if we want to look at it or bring anything back.


edotau (approved):
Awesome, much needed. Thanks!!

cshoben (approved):

rimangan (approved):

PR #301, opened 2020-10-08
Chelsea branch

cshoben:
Contained within this push are two test fasta files, code for creating a graph based on that cigar output from a global alignment conducted using the function constantGap(), and an accompanying test file that currently has roughly 43% coverage. 
Note that cigar output in this instance only informs on 0: match, 1: insertion, or 2:deletion. These indels are relative to the first fasta file given by the user, referenced as the 'target' sequence.

### DIFF
--- a/cmd/globalAlignment/globalAlignment.go
+++ b/cmd/globalAlignment/globalAlignment.go
@@ -1,0 +1,92 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/align"
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/simpleGraph"
+	"log"
+)
+
+func usage() {
+	fmt.Print("./globalAlignment - chelsea's global alignment \n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var expectedNum int = 2
+	flag.Usage = usage
+	flag.Parse()
+	if len(flag.Args()) != expectedNum {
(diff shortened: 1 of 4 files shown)


cmd/globalAlignment/globalAlignment.go
```diff
@@ -0,0 +1,102 @@
...
+			queryEnd += aln[i].RunLength
+			curr.Name = fmt.Sprintf("%s_%d_%d", query.Name, targetEnd, targetEnd)
+			simpleGraph.AddNode(answer, curr)
+			simpleGraph.AddEdge(answer.Nodes[i-1], curr, 0.5)
+		case aln[i].Op == 2:
+			curr.Seq = target.Seq[targetEnd : targetEnd+aln[i].RunLength]
+			targetEnd += aln[i].RunLength
+			simpleGraph.AddNode(answer, curr)
+			simpleGraph.AddEdge(answer.Nodes[i-1], curr, 0.5)
+		}
```
craiglowe, line 98: what are your thoughts on this code handling corner cases like a deletion coming right after an insertion?  Have you thought about generalizing this to handling matching being broken into identity and mismatches?

craiglowe (commented):
Looks wonderful!  Few quick thoughts

edotau (dismissed):
Great work, I've seen most of this already so I don't have much more to add. I am most impressed with how much you put into writing the test function.

We might want to eventually move the cigarToGraph function into the standard library, but in order to avoid some of our import cycles problems, I am not yet sure which package the function should live in.

cshoben:
Made the needed changes that Christi pointed out. Thank you everyone for your feedback! Read for approval.

edotau (approved):

DanielSnellings (approved):

craiglowe (approved):
Wonderful!

PR #303, opened 2020-10-13
MCMC verbose, runtime optimization, and documentation

rimangan:
Some changes:
-Documentation for functions associated with the selectionMCMC.
-Likelihoods are now calculated upon parameter set generation, which should cut runtime in half.
-Verbose global variable can be set to produce an executable with additional log.Prints for debug. Let's discuss this convention at code review.

### DIFF
--- a/popgen/mcmcSelection.go
+++ b/popgen/mcmcSelection.go
@@ -6,48 +6,55 @@ import (
 	"math"
 	"math/rand"
 	"fmt"
+	"log"
 	//DEBUG packages:
-	/*"log"
-	"os"
+	/*"os"
 	"runtime/pprof"*/)
+
+//To access debug prints, set verbose to 1 and then compile.
+var verbose int = 0
 
 type Theta struct {
 	alpha       []float64
 	mu          float64
 	sigma       float64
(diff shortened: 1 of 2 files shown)


popgen/mcmcSelection.go
```diff
@@ -6,48 +6,55 @@ import (
...
 	"fmt"
+	"log"
 	//DEBUG packages:
-	/*"log"
-	"os"
+	/*"os"
 	"runtime/pprof"*/)
 
+//To access debug prints, set verbose to 1 and then compile.
+var verbose int = 0
```
edotau, line 15: this should be defined as a const

edotau: When you use const at compile time, the computer will essentially go through your code and copy and paste a 0 (or any number you set it)and replace verbose. This does not affect the run time, but compile time is affected. This is different when you use var.

edotau (dismissed):

DanielSnellings (approved):

cfauci (approved):

craiglowe (approved):
Great Progress!

PR #305, opened 2020-10-19
nkpCache

rimangan:
MCMC evaluates integrals over the product of a stationarity function and a binomial function, but the answers to the binomial function are recalculated each time, despite always representing constant values. Therefore, I'm implementing a caching system for these values in a variable called the nkpCache, a [][][]float64 in which the first array layer represents n, the number of alleles, the second layer represents [k], the number of successes, and the third layer represents [p], the allele frequency. Thus, nkpCache[n][k][p] = binompdf(n, k, p).


common/math.go
```diff
@@ -8,6 +8,15 @@ import (
 	"math/rand"
 )
 
+func IntSliceContains(a []int, b) bool {
```
craiglowe, line 11: No need to move it, but for future things I would keep the "numbers" package in mind.

popgen/mcmcSelection.go
```diff
@@ -154,3 +180,16 @@ func findMaxN(data AFS) int {
 	}
 	return answer
 }
+
+func cacheIndexToAlleleFrequencyCache(bins int) []float64 {
+	var answer []float64 = make([]float64, bins + 1)
+	//this is done with linear interpolation, as index zero corresponds to an allele frequency of the LeftBound and the index at len[n][k], the number of values p, is the rightBound allele frequency. y = mx + b. First we find m and b.
+	//b is simply the LeftBound.
+	m := (LeftBound - RightBound) / (bins + 1)
```
craiglowe, line 188: Where do LeftBound and RightBound come from?  Should it be RightBound - LeftBound?  I don't see them defined in this function or this file.  Maybe global to this package and defined in another file?  Should (bins + 1) be "bins"?

popgen/mcmcSelection.go
```diff
@@ -154,3 +180,16 @@ func findMaxN(data AFS) int {
...
+	for i := 0; i < len(bins+1); i++ {
+		answer[i] = float64(i)*m + LeftBound
```
craiglowe, line 190: I think answer[bins] should equal RightBound, which is why I think "m" should be divided by only "bins"

popgen/mcmcSelection.go
```diff
@@ -122,13 +123,28 @@ func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations
...
 	maxN := findMaxN(data)
-	binomMap := make([][]float64, maxN+1)
+	allN := findAllN(data)
+	alleleFrequencyCache = cacheIndexToAlleleFrequencyCache(bins)
+	nkpCache := make([][][]float64, maxN+1)
+	var coefficient float64
+
+	for n := 0; n < len(allN); n++ {
+		nkpCache[allN[n]] = make([][]float64, allN[n])
+		for k := 0; k < len(allN[n]); k++ {
```
craiglowe, line 133: Should "k" start at one?

popgen/mcmcSelection.go
```diff
@@ -122,13 +123,28 @@ func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations
...
+			coefficient = numbers.BinomCoefficient(n, k)
+			nkpCache[allN[n]][k] = make([]float64, bins+1)
+			for p := 0; p < len(bins); p++ {
```
craiglowe, line 136: Should it be len(bins) + 1?

popgen/mcmcSelection.go
```diff
@@ -146,6 +162,16 @@ func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations
 	}
 }
 
+func findAllN(data AFS) []int {
+	var answer []int = make([]int, 0)
+	for i := 0; i < len(data.sites); i++ {
+		if !common.IntSliceContains(data.sites[i].n) {
```
craiglowe, line 168: I think this needs "answer" too

popgen/mcmcSelection.go
rimangan: the argument here is we don't need to initialize these values because all variants have a frequency of at least k = 1? Am I understanding correctly?

popgen/mcmcSelection.go
rimangan: bins + 1 in fact, bins is already an int, so the len() was erroneous.

popgen/mcmcSelection.go
rimangan: Great catch!

numbers/integrate.go
```diff
@@ -45,6 +45,7 @@ func LogIntegrateIterative(f func(float64) float64, a float64, b float64, maxIte
 		n := n * 10
 		curr := LogIntegrate(f, a, b, n)
 		if (prev-curr)/curr < relativeError {
+			//DEBUG: log.Printf("In LogIntegrateIterative: i=%v.", i)
```
edotau, line 48: If this function gets called often, it might be worth pulling out and declare the variables `n and curr` out of this i loop so you don't end up allocating memory from i to maxIter.

popgen/mcmcSelection.go
```diff
@@ -122,20 +127,35 @@ func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations
...
 	maxN := findMaxN(data)
-	binomMap := make([][]float64, maxN+1)
+	allN := findAllN(data)
+	var alleleFrequencyCache []float64 = cacheIndexToAlleleFrequencyCache(bins)
+	nkpCache := make([][][]float64, maxN+1)
+	var coefficient float64
+
+	for n := 0; n < len(allN); n++ {
+		nkpCache[allN[n]] = make([][]float64, allN[n])
+		for k := 1; k < allN[n]; k++ {
```
edotau, line 137: This triple nested loop gets a little crazy allocating memory for k and p. Although it is possible the complier performs these optimizations during run time, it's very possible it doesn't and if that is the case, you are basically do an extra k*n operations by doing` k := 1` and `p := 0` under the n loop.

edotau (commented):
a couple small and simple optimizations that might speed things up

popgen/mcmcSelection.go
rimangan: I'm not sure I understand where there is an extra calculation, could you expand a bit on this?

edotau: because you call k := 1 under the `for n := 0; n < len(allN); n++ {` you allocate new memory for every new iteration of n. whereas simply doing k = 0 (no colon) you reuse the same memory. it's very subtle but k := 1 and k = 1 behaves a little different

edotau: on top of this there is the p := 0, under the k for loop. For for every iteration of K, calling p := 0 is different from p = 0. `p := 0` is called k times which results in another n memory allocations.
in summary doing something like:
```
var n, k, p int
for n = 0; n < len; n++ {
	for k = 0; k < len; k ++ {
		for p = 0; p < len; p++ {
				
		}
	}
} 
```
is more desirable than:
```
for n  := 0; n < len; n++ {
	for k  := 0; k < len; k ++ {
		for p := 0; p < len; p++ {
				
		}
	}
}

Comment (edotau) [reply popgen/mcmcSelection.go:137]: if you don't believe me, here is a quick benchmark function you can
`go test -bench=.` gives us:
BenchmarkTripleLoopWithColon-8   	`1997708,	       586 ns/op`
BenchmarkTripleLoopWithout-8       `3108403,	       383 ns/op`
gives us roughly a 30% boost in performance
```
package main

import (
	"testing"
)

func BenchmarkTripleLoopWithColon(b *testing.B) {
	b.ResetTimer()
	for n := 0; n < b.N; n++ {
		WithColon(10,10,10)
	}
}

func BenchmarkTripleLoopWithout(b *testing.B) {
	b.ResetTimer()
	for n := 0; n < b.N; n++ {
		WithOut(10, 10, 10)
	}
}

func WithColon(one, two, three int) int{
	var ans int = 0
	for i := 0; i < one; i++ {
		for j := 0; j < two; j++ {
			for k := 0; k < three; k++ {
				ans+=1
			}
		}
	}
	return ans
}

func WithOut(one, two, three int) int{
	var ans int = 0
	var i, j, k int
	for i = 0; i < one; i++ {
		for j = 0; j < two; j++ {
			for k = 0; k < three; k++ {
				ans+=1
			}
		}
	}
	return ans
}
```

File: numbers/integrate.go [hunk 9]
```diff
@@ -45,6 +45,7 @@ func LogIntegrateIterative(f func(float64) float64, a float64, b float64, maxIte
 		n := n * 10
 		curr := LogIntegrate(f, a, b, n)
 		if (prev-curr)/curr < relativeError {
```
craiglowe, line 47: I am not sure that prev will be greater than curr, so I think this numerator needs to be wrapped in math.Abs

popgen/mcmcSelection.go
craiglowe: I think there is a good general point here that the ":=" operator is quick and easy when writing code, but can at times have drawbacks when performance is important.  One thing that might be good is to go through and look for ":=" within loops and replace it by declaring the variable once at the top of the function.

popgen/mcmcSelection.go
```diff
@@ -146,6 +166,16 @@ func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations
 	}
 }
 
+func findAllN(data AFS) []int {
```
craiglowe, line 169: Very minor comment since this is not the part of the code holding us back, but findAllN and findMaxN could be returned as two variables from the same function so that you only need to go through "data AFS" once.

popgen/nkpCacheIntegral.go
```diff
@@ -0,0 +1,34 @@
+package popgen
+
+import (
+	"github.com/vertgenlab/gonomics/numbers"
+	"math"
+)
+
+func LogIntegrateStationarityCache(alpha float64, n int, k int, stepSize int, alleleFrequencyCache []float64, nkpCache [][][]float64) float64 {
+	var bins int = len(nkpCache[n][k])-1
+	var deltaX float64 = (RightBound - LeftBound) / float64(bins)
```
craiglowe, line 10: This should have "stepSize" as part of the calculation to determine deltaX.  This bug might actually give you a significant speedup because right now I bet that every time through the loop of how many bins to use, it gets a really different answer.  This may cause it to always use 100,000 bins.

popgen/nkpCacheIntegral.go
```diff
@@ -0,0 +1,34 @@
...
+	answer = numbers.MultiplyLog(numbers.MidpointLog(firstLeft, nextLeftEval), logDeltaX)
+
+	for i := stepSize; i < bins; i += stepSize {
+		currLeftIndex += stepSize
+		currRightIndex += stepSize
+		rightEval = f(alpha, currRightIndex, alleleFrequencyCache, nkpCache[n][k][currRightIndex])
+		answer = numbers.AddLog(answer, numbers.MultiplyLog(numbers.MidpointLog(nextLeftEval, rightEval), logDeltaX))
+		nextLeftEval = rightEval
+	}
+	return answer
```
craiglowe, line 29: The way you calculate answer currently is the most readable, but I think we can speed it up with some math tricks.  One thing to note is that adding in log-space is expensive because you need to do both a math.Exp and a math.Log to get the answer.
If we have 5 bins, then we will have step ~= 0.2 going between near-0 and near-1.  Right now we do:
(f(near-0)+f(0.2))/2 * 0.2 + (f(0.2)+f(0.4))/2 * 0.2 + (f(0.4)+f(0.6))/2 * 0.2 + (f(0.6)+f(0.8))/2 * 0.2 + (f(0.8)+f(near-1))/2 * 0.2
additions is: 2 * bins -1
we could do:
0.2 * (f(near-0)/2 + f(0.2) + f(0.4) + f(0.6) + f(0.8) + f(near-1)/2)
this comes out the same, but almost halves our additions

craiglowe (commented):
Looks great.  Made some comments about bugs and possible speed-ups.  One other option is to replace the math.Log(2.0) in your midpoint function with math.Ln2, so that you don't call the math.Log function as much

numbers/integrate.go
rimangan: Great catch!

popgen/mcmcSelection.go
rimangan: Great, thanks!

popgen/nkpCacheIntegral.go
rimangan: Great call! Should have stepSize as a multiplier, as a stepsize of 10 has 10 times the deltaX, right?

popgen/nkpCacheIntegral.go
rimangan: So instead of keeping an sum of areas I can keep a sum of heights and multiply by the width at the end? I'll change variable names accordingly so we can get the speed up and keep some readabilitty

craiglowe (approved):
Great to see this!  Wonderful work!

cfauci (dismissed):

PR #309, opened 2020-10-26
Craig simpson log main

rimangan:
Simpson's rule in logSpace. Passes tests, MCMC runs about 13,000 times faster. Down to 30ms iterations.


numbers/integrate.go
```diff
@@ -157,3 +157,59 @@ func AdaptiveSimpsons(f func(float64) float64, a float64, b float64, errorThresh
...
+	logHOverSix = math.Log(h / 6)
+	leftEstimate = MultiplyLog(logHOverSix, AddLog(AddLog(fa, MultiplyLog(logFour, fLeftMidpoint)), fMidpoint))
+	rightEstimate = MultiplyLog(logHOverSix, AddLog(AddLog(fMidpoint, MultiplyLog(logFour, fRightMidpoint)), fb))
+	estimateFromHalves = AddLog(leftEstimate, rightEstimate)
+
+	//log.Printf("maxDepth:%d, left:%f, right:%f, fromHalves:%f, whole:%f\n", maxDepth, math.Exp(leftEstimate), math.Exp(rightEstimate), math.Exp(estimateFromHalves), math.Exp(estimateFromHalves))
+
+	if estimateFromHalves > wholeEstimate {
+		delta = SubtractLog(estimateFromHalves, wholeEstimate)
+		if delta <= MultiplyLog(logFifteen, errorThresh) {
```
craiglowe, line 190: This "if statement" is the same in both cases and could be pulled out and put below the if that just calculates what delta should be

popgen/mcmcSelection.go
```diff
@@ -40,8 +41,8 @@ func MetropolisAccept(old Theta, thetaPrime Theta) bool {
 //HastingsRatio is a helper function of MetropolisAccept that returns the Hastings Ratio (logspace) between two parameter sets.
 func HastingsRatio(tOld Theta, tNew Theta) float64 {
 	var newGivenOld, oldGivenNew float64
-	newGivenOld = numbers.NormalDist(tNew.mu, tOld.mu, tOld.sigma) //* numbers.GammaDist(tNew.sigma, tOld.sigma*tOld.sigma, tOld.sigma)
-	oldGivenNew = numbers.NormalDist(tOld.mu, tNew.mu, tNew.sigma) //* numbers.GammaDist(tOld.sigma, tNew.sigma*tNew.sigma, tNew.sigma)
+	newGivenOld = numbers.NormalDist(tNew.mu, tOld.mu, tOld.sigma) * numbers.GammaDist(tNew.sigma, tOld.sigma*tOld.sigma, tOld.sigma)
```
craiglowe, line 44: The line and the next should mirror the proposal function which I think uses alpha=50 and beta=50/sigma

popgen/stationarity.go
```diff
@@ -95,13 +96,30 @@ func AFSStationarity(p float64, alpha float64) float64 {
...
 }
 
+func DetectionProbability(p float64, n int) float64 {
+	var pNotDetected float64 = numbers.AddLog(numbers.BinomialExpressionLog(n, 0, p), numbers.BinomialExpressionLog(n, n, p))
+	log.Printf("pNotDetected: %f.", pNotDetected)
+	return numbers.SubtractLog(0, pNotDetected)
+}
+
+func AfsStationarityCorrected(p float64, alpha float64, n int) float64 {
+	log.Printf("p: %f. n: %d. alpha: %f. Detection: %f. Stationarity: %f.", p, n, alpha, math.Exp(DetectionProbability(p, n)), math.Exp(AFSStationarity(p, alpha)))
```
craiglowe, line 106: comment out for main?

craiglowe (commented):
Great to see this!  One suggested update in hastings ratio

numbers/logDistribution.go
```diff
@@ -8,25 +8,29 @@ import (
...
+func BinomialDistLogSlice(n int, k int, p float64, binomCache [][]float64) float64 {
 	//DEBUG: fmt.Printf("n: %v. k: %v. len(binomMap): %v.", n, k, len(binomMap))
-	if binomMap[n] == nil {
-		binomMap[n] = AddBinomMapEntry(n)
+	if binomCache[n] == nil {
+		binomCache[n] = AddBinomMapEntry(n)
 	}
-	s := LogPowInt(p, k)
-	f := LogPowInt(1.0-p, n-k)
+	s := LogPow(p, float64(k))
```
craiglowe, line 30: I think some of these changes are already in main.  I think the thing to do is to change over to main, get the newest version by running git pull, and then going back to this branch and merging in main.

popgen/mcmcSelection.go
```diff
@@ -40,8 +42,8 @@ func MetropolisAccept(old Theta, thetaPrime Theta) bool {
 //HastingsRatio is a helper function of MetropolisAccept that returns the Hastings Ratio (logspace) between two parameter sets.
 func HastingsRatio(tOld Theta, tNew Theta) float64 {
 	var newGivenOld, oldGivenNew float64
-	newGivenOld = numbers.NormalDist(tNew.mu, tOld.mu, tOld.sigma) //* numbers.GammaDist(tNew.sigma, tOld.sigma*tOld.sigma, tOld.sigma)
-	oldGivenNew = numbers.NormalDist(tOld.mu, tNew.mu, tNew.sigma) //* numbers.GammaDist(tOld.sigma, tNew.sigma*tNew.sigma, tNew.sigma)
+	newGivenOld = numbers.NormalDist(tNew.mu, tOld.mu, tOld.sigma) * numbers.GammaDist(tNew.sigma, step, tOld.sigma/step)
```
craiglowe, line 45: I think this should be "step/sigma" instead of "sigma/step"

craiglowe (commented):
Looks great!  I think I found one typo in the hastings ratio and I think we need a merge from the current version of main.

numbers/logSpace.go
```diff
@@ -8,7 +8,7 @@ import (
 
 //LogCanconvert returns true if the input logSpace number can be converted to a number in normal space without overflow or underflow.
 func LogCanConvert(x float64) bool {
-	return x < 709.782712893383973096 || x > -745.133219101941108420
+	return x < 709.4 && x > -745.1
```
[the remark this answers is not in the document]
craiglowe: This is an adjustment of the values and not only a rounding.  For example, I get +Inf when calling math.Exp on 709.5, but not 709.4 (in the go playground).

numbers/logSpace.go
```diff
@@ -90,6 +90,9 @@ func LogPow(x float64, y float64) float64 {
 	if x < 0 {
 		log.Fatalf("LowPowInt does not handle negative x values. x=%e\n", x)
 	}
+	if y == 0.0 {
```
edotau, line 93: One small improvement I can come up with is rearranging the order of the if statements to return `y*math.Log(x)` first; followed by the log.Fatal and y==0 will help exit out of the function sooner. something to start thinking about is the order of your logical equivalent if statements influence the behavior of function slightly

something like:
```
if x > 0 && y != 0.0
      return y * math.Log(x)
else
      if x < 0
             log.Fatal
      if y == 0.0
             return 0.0
```

edotau:
Excellent, looks good to me! Once the merging conflicts are resolved I will approve the pull request.

numbers/logSpace.go
craiglowe: I am interested to learn more about this at code review.  I would have naively thought that either way the code's "path" would have been to 1) test x, 2) test y, 3) return the result from the equation.  Does the speed-up come from branch prediction at the level of the processor where it assumes it will not take the branch?  Or are you changing the tests to make it faster (e.g. == to != for y)?

edotau: As I stare at this a bit more I think in this particular case you are correct that it more or less runs the same way. I was trying to change the tests to make it faster. I was just trying to get the function to return y * math.Log(x) as quickly as possible before it has to check the other if statements.  I think the logic is more obvious if we only have to check for one variable, but since we have to check x and y anyway i think you are correct, unless we do some testing, it ends up being very similar.

edotau (approved):

cfauci (approved):

craiglowe (approved):
Great work!  Feels like you are getting close!

PR #311, opened 2020-10-26
Gene pred struct

cfauci:
Created GenePred Data Struct based on UCSC genePred and Extended genePred. Simulate updates to use gtf will be changed to using genePred next push


genePred/genePred.go
```diff
@@ -0,0 +1,173 @@
...
+	return &current
+}
+
+func StringToIntSlice(text string) []int {
+	values := strings.Split(text, ",")
+	var answer = make([]int, len(values)-1)
+
+	for i := 0; i < len(values)-1; i++ {
+		answer[i] = common.StringToInt(values[i])
+	}
```
edotau, line 118: One thing we could consider doing is to check the final element in the slice (`values[len(values)-1]`) if it equals 0 and log an error if it doesn't.

genePred/genePred.go
```diff
@@ -0,0 +1,173 @@
...
+				log.Fatal("frame is offset by more than 2 positions")
+			}
+			//DEBUG: fmt.Print(exonFrames)
+		}
+	}
+	return exonFrames
+}
+
+func SliceIntToString(slice []int) string {
+	var buffer strings.Builder
```
edotau, line 168: Excellent, string.Builder is the way to go!

genePred/genePred.go
```diff
@@ -0,0 +1,173 @@
...
+	exonEnds := gene.ExonEnds
+	cdsStart := gene.CdsStart
+	var length int
+	var nextExonLength int
+	var nextExonFrame int
+	var exonFrames []int
+	exonFrames = append(exonFrames, 0)
+	var answer int
+
+	for i := 0; i < len(exonEnds)-1; i++ { // - 1 compensates for UCSC format with ending comma
```
edotau, line 133: Wait, just to be sure, since we dealt with the len()-1 when we read in the genePred file, do we have to do it again?

edotau (commented):
Excellent work! Looks great!

Just a couple small comments on condensing the ToString function and a double check on if you accounted for the trailing comma twice.

genePred/genePred.go
```diff
@@ -0,0 +1,166 @@
...
+	} else if words[2] == "-" {
+		current.Strand = '-'
+	} else {
+		log.Fatal("no strand specified")
+	}
+	current.TxStart = common.StringToInt(words[3])
+	current.TxEnd = common.StringToInt(words[4])
+	current.CdsStart = common.StringToInt(words[5])
+	current.CdsEnd = common.StringToInt(words[6])
+	current.ExonStarts = StringToIntSlice(words[7])
```
craiglowe, line 87: I think that the genePred file format will have the exon number as a column in the file after CdsEnd and before ExonStart.  This means that exonNumber from a few lines down would be set equal to "common.StringToInt(words[7])" and then words[8] would be processed to get ExonStarts.
To see an example genePred file, you can go to the ucsc genome browser, click on tools, click on table browser.  Now select a gene track and get the output as "all fields from selected table".  When you get the data it will be a genePred file.

genePred/genePred.go
craiglowe: Would this be the same as checking if the last item of values is equal to the empty string?

genePred/genePred.go
```diff
@@ -0,0 +1,166 @@
...
+	var nextExonLength int
+	var nextExonFrame int
+	var exonFrames []int
+	exonFrames = append(exonFrames, 0)
+	var answer int
+
+	for i := 0; i < len(exonEnds)-1; i++ { // - 1 compensates for not needing to calculate the frame of the exon after the last
+		if i == 0 {
+			//for first exon
+			length = (exonEnds[0] - cdsStart) + 1
```
craiglowe, line 129: This "+ 1" surprises me.  I would have thought that the exonEnds was "open" (aka the base after the end), meaning that the length would be just the start - end.

genePred/genePred.go
```diff
@@ -0,0 +1,166 @@
...
+			if exonTwoFrame == 0 {
+				answer = exonTwoFrame
+			} else {
+				answer = 3 - exonTwoFrame
+			}
+			exonFrames = append(exonFrames, answer)
+			//DEBUG: log.Print(exonFrames)
+		} else {
+			//for all other exons, which depend on the frame being calculated ahead of this step
+			nextExonLength = exonEnds[i] - exonStarts[i] + exonFrames[i] + 1
```
craiglowe, line 143: This plus one surprises me for the same reasons as above.  Not saying it is wrong, but we should make sure to double check with a test file from UCSC

genePred/testGenePred.gpd
```diff
@@ -0,0 +1,2 @@
+test	0	+	0	1000	0	900	0,18,500,800,	2,20,502,802,
```
craiglowe, line 1: I think this should have a "4" in a column after the "900" column to match the UCSC format.  See the file the comes out of the table browser, or this table, https://genome.ucsc.edu/FAQ/FAQformat.html#format9 , where each element in the table is the next column.

craiglowe (commented):
Great work!  Couple of small comments

genePred/genePred.go
cfauci: I was going to ask the same thing, we can talk about it in code review I think so we are all on the same page

edotau: I think if you call `common.StringToInt(values[i])` on the empty string, you get 0 back, but I have to check

craiglowe: We should probably change common.StringToInt to exit with an error if it gets the empty string.  We could add a TODO about that (or a pull request to fix it) if it is silently returning zeros.  Maybe we could use strings.HasSuffix here to make sure that "text" ends with a comma?

edotau: I think checking the suffix for a comma is a great idea! Don't quote me on the common.StringToInt(), I'll have to double check, I am just going off what happened last time when I programmed this and a TODO to double check common.StringToInt would be great, but as for this check in, i think adding something on line 109 to have a check like: `if values[len(values)-1] != "" { log.Fatalf("Error: Last element in values should be equal to the empty string")}` is something I would do.

cfauci: I added 2 checks on lines 89-98 for the suffix check

rimangan (approved):

edotau (approved):

craiglowe (approved):
Great work so far with the GenePred

PR #312, opened 2020-11-02
Lift chain position draft

rimangan:


DanielSnellings:
I know this is still a draft, but in case you were unaware it looks like some streams are being crossed here between lift and MCMC. Eventually these should be split into different pull requests.

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
+package numbers
+
+import (
+	"math"
+	"math/rand"
+	"log"
+)
+
+const RelError float64 = 1.0e-8
```
craiglowe, line 9: Are these constants specific to one function?  I would put them inside that function.  Right now they are kind of general names and available to all functions inside "numbers".

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+const Small float64 = 1.0e-30
+const MaxIterations int = 200
+
+//RandBeta returns a random number drawn from a beta distribution with parameters alpha and beta as an xy point where y is the function density.
+//Implemented from principles in R. C. H. Cheng (1978). Generating beta variates with nonintegral shape parameters. Communications of the ACM 21, 317-322.
+func RandBeta(alpha float64, beta float64) float64 {
+	r := rand.Float64()
```
craiglowe, line 16: Can we remove setting r to a random float?  I don't see this in the Cheng paper

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+	for r <= 0.0 || r >= 1.0 {//catches the case where r is exactly  0.0 or 1.0
```
craiglowe, line 17: I am not sure we need r as a random float, but rand.Float64() is left closed right open, so I don't think it will ever be 1

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+		r =rand.Float64()
+	}
+	
+	//special returns
+	if alpha < 0 || beta < 0 {
+		log.Fatalf("Error in RandBeta: alpha and beta must be greater than 0.  alpha: %f. beta: %f.", alpha, beta)
```
craiglowe, line 23: greater than or equal to zero?

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if math.IsInf(alpha, 1) && math.IsInf(beta, 1) {
+		return 0.5
+	}
+	if alpha == 0 && beta == 0 {
+		return 1.0
```
craiglowe, line 29: Should this be a coin flip that is half the time zero and half the time 1?

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+	if math.IsInf(alpha, 1) || beta == 0 {
+		return 1.0
+	}
+	if math.IsInf(beta, 1) || alpha == 0 {
+		return 0.0
+	}
+
+	var sum, a, b, bet, delta, k1, k2, s, t, u1, u2, v, w, y, z float64
+	var i int
+	a = MinFloat64(alpha, beta)
```
craiglowe, line 40: These two lines look they are are to make a the smaller of the two and b the bigger of the two.  With a single comparison we could do if a <= b { a = alpha, b = beta } else { a = beta, b = alpha }

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+		k1 = delta * (0.0138889 + 0.0416667 * a) / (b * bet - 0.777778)
+		k2 = 0.25 + (0.5 + 0.25 / delta) * a
+
+		for i = 0; i < MaxIterations; i++ {
+			u1 = rand.Float64()
+			u2 = rand.Float64()
+			if u1 < 0.5 {
+				y = u1 * u2
+				z = u1 * y
+				if (0.25 * u2 + z - y >= k1) {
```
craiglowe, line 57: don't need parens

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+					continue
+				}
+			} else {
+				z = u1 * u1 * u2
+				if (z <= 0.25) {
```
craiglowe, line 62: don't need parens

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+					//this block of code was broken into a helper function in the R implementation
+					v = beta * math.Log(u1 / (1.0 - u1))
+					if (v <= math.Log(math.MaxFloat64)) {
```
craiglowe, line 65: don't need parens

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+					}
+					break
+				}
+				if z >= k2 {
+					continue
+				}
+			}
+
+			v = beta * math.Log(u1 / (1.0 - u1))
+			if (v <= math.Log(math.MaxFloat64)) {
```
craiglowe, line 81: don't need parens

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+		return 0.0
+	}
+
+	var sum, a, b, bet, delta, k1, k2, s, t, u1, u2, v, w, y, z float64
+	var i int
+	a = MinFloat64(alpha, beta)
+	b = MaxFloat64(alpha, beta)
+	sum = a + b
+
+	if a <= 1.0 {
```
craiglowe, line 44: more comments might help people if/when they come back to this.  This "if" is to test to see if algorithm BC from page 321 can be used

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+		if i == MaxIterations {
+			log.Fatalf("Failed to converge.")
+			return -1.0
+		}
+
+		if alpha == a {
+			return a / (a + w)
+		}
+		return w / (a + w)
+	} else {
```
craiglowe, line 104: comments to add: else we use algorithm BB from page 320

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+		var u1, u2 float64
+		var bet float64 = math.Sqrt((sum - 2.0) / (2.0 * a * b - sum))
+		var gamma float64 = a * 1.0 / bet
```
craiglowe, line 107: + instead of times?

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+		for r + sum * math.Log(sum / (b + w)) < t {
```
craiglowe, line 108: We have to make sure we execute this loop at least once because 'r' has not been set yet.
The style in golang would be:
for keepGoing := true; keepGoing; keepGoing = r+sum*math.Log(sum / (b + w)) < t {

numbers/randBeta.go
```diff
@@ -0,0 +1,182 @@
...
+				w = math.MaxFloat64
+			}
+
+			z = u1 * u1 * u2
+			r = gamma * v - 1.3862944
+			s = a + r - w
+			if (s + 2.609438 >= 5.0 * z) {
+				break
+			}
+			if s > t {
```
craiglowe, line 128: First part of "step 3" in the paper says we need to set t to math.Log(z) before the comparison on this line

craiglowe (commented):
Couple comments on beta code.  When they say things like the constant is log(4), should we add more digits?

interval/query.go
```diff
@@ -0,0 +1,110 @@
...
+	// How the file is read is dependent on the file extension
+	filetype := path.Ext(inputFile)
+
+	if filetype == ".gz" {
+		// If terminal extension is ".gz" then trim off the gz and get the next extension
+		filetype = path.Ext(inputFile[0 : len(inputFile)-len(filetype)])
+	}
+
+	switch filetype {
+	case ".bed":
```
craiglowe, line 32: Bed seems like the "easy" one to implement and I know you are doing vcf for your own work.  These others seem pretty hard, so I don't think we should fully tackle them until someone needs them.  Should we initially leave them commented out, or is it worth it to do a crude lift and with a warning that only coordinates are lifting and not all the underlying data in the struct?

interval/query.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+	case ".axt":
+		receive := axt.GoReadToChan(inputFile)
+		for val := range receive {
+			send <- val
+		}
+
+	case ".vcf":
+		receive, _ := vcf.GoReadToChan(inputFile)
```
craiglowe, line 93: Is the second argument the header?  Do we need to lift that as well?  I am thinking probably not for most cases, but maybe with something like sam we would need to lift the header so that someone could then make a bam out of it or view it on a browser?

numbers/monteCarlo.go
```diff
@@ -89,25 +89,38 @@ func chooseBin(sumHeights float64, binHeights []float64) int {
...
+func BoundedRejectionSample(boundingSampler func() (float64, float64), f func(float64) float64, xLeft float64, xRight float64, maxIteration int) float64 {
+	var xSampler, ySampler, y float64
+	for i := 0; i < maxIteration; i++ {
+		xSampler, ySampler = boundingSampler()
+		y = f(xSampler)
+		if y > ySampler {
+			log.Fatalf("BoundedRejectionSample: function was not a valid bounding function, ySampler is greater than y.")
+		}
+		if RandFloat64InRange(0.0, ySampler) < y {
+			return y
```
craiglowe, line 114: Should this be returning xSampler?

numbers/monteCarlo.go
```diff
@@ -161,19 +181,21 @@ func RandGamma(a float64, b float64) float64 {
...
 			if p >= 1.0 {
 				x = -1 * math.Log((e-p)/a)
-				if RandExp() >= (1.0-a)*math.Log(x) {
+				if rExp >= (1.0-a)*math.Log(x) {
 					break
 				}
 			} else {
 				x = math.Exp(math.Log(p) / a)
-				if RandExp() >= x {
+				rExp, _ = RandExp()
```
craiglowe, line 192: I don't think this line is needed because you generate a number of line 184 that you can use?

craiglowe (commented):
Great work!  I had a few small comments.

numbers/incompleteBeta.go
```diff
@@ -0,0 +1,51 @@
+package numbers
+
+import (
+	"log"
+	"math"
+	"math/rand"
```
edotau, line 6: Can you run a go test in the numbers package for me? I am getting an error message that says `./incompleteBeta.go:6:2: imported and not used: "math/rand"` when i checked out your branch. I think we should just remove `"math/rand"` but I just want to make sure.

cmd/sequelOverlap/intervalQuery.go
```diff
@@ -71,49 +71,3 @@ func writeToFile(answerChan <-chan *queryAnswer, outfile *fileio.EasyWriter, mer
 		}
 	}
 }
-
```
edotau, line 74: I think you forgot to get rid of the imported packages at the top when you removed these function. Can you make sure to run `go build` to make sure the CMD complies with no errors?

interval/query.go
edotau: Yes, sam header needs to be updated. Maybe we can pull that info from the chains?

chain/methods.go
```diff
@@ -28,6 +27,23 @@ func (ch *Chain) GetChromEnd() int {
 	}
 }
 
+func (ch *Chain) UpdateLift(c string, start int, end int) {
+	ch.TName = c
+	if ch.TStrand {
+		ch.TStart = start
+	} else {
+
```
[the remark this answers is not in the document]
edotau: I'm a little confused. Would anyone ever need to perform a liftover from chain to chain? Isn't that a simple Target <->Query Swap?

chain/lift.go
```diff
@@ -0,0 +1,73 @@
+package chain
+
+import (
+	"log"
+)
+
+//TPosToQPos converts a target position in a chain to the corresponding query position.
+func TPosToQPos(c *Chain, TPos int) int {
+	if TPos < c.TStart || TPos > c.TEnd {
```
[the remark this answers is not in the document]
edotau: Yes, what happens when the target and/or query is negative?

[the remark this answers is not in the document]
edotau: I think it's sometime like:
```
newStart = len(size)-end
newEnd = len(size)-start
```

interval/lift.go
```diff
@@ -0,0 +1,13 @@
+package interval
+
+import (
+	"github.com/vertgenlab/gonomics/fileio"
+)
+
+type Lift interface {
```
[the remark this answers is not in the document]
edotau: Project?

interval/query.go
edotau: Yes, bed should be the simplest case and more widely used

edotau (commented):
I was able to get through about half of this so far. Will try to finish tomorrow.

Can we please try our best to keep some of these check in separate? I understand it's hard to separate things out sometimes, but Interfaces are especially tricky to review and often are methods implemented in several packages. I don't really have a problem with the length, but there are things to address with the chains and the lift interface and that slows dow getting the small bug fixes or the numbers package checked in.

sam/methods.go
```diff
@@ -23,6 +23,11 @@ func (s *SamAln) GetChromEnd() int {
 	return s.GetChromStart() + runlength
 }
 
+func (s *SamAln) UpdateLift(c string, start int, end int) {
+	s.RName = c
+	s.Pos = int64(start + 1)
```
edotau, line 28: What is the SamToBed convention? It's not start plus 1 is it? I imagine it's something like the end coordinate of the reference the read is aligned to, but that gets complicated because I think now you have to take into account the cigar run length information.

chain/methods.go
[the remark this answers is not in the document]
rimangan: yes I don't think there is much of a use case for lifting chains, but it theoretically should be possible so I was just considering writing out the methods. I'll comment it out for now

cmd/sequelOverlap/sequelOverlap.go
```diff
@@ -28,7 +28,7 @@ func sequelOverlap(options *Settings) chan *queryAnswer {
 
 	tree := buildTree(intervals, options.Aggregate)
 
-	queryChan := goReadToIntervalChan(options.Input)
+	queryChan := Interval.GoReadToIntervalChan(options.Input)
```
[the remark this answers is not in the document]
rimangan: This is a great suggestion

sam/methods.go
edotau: Never mind I found it. Dan implemented the logic in GetChromEnd() in methods.go

interval/query.go
rimangan: This will be interesting to handle. We may need a distinct readToLift return for the data types with a header.

numbers/monteCarlo.go
rimangan: yes thanks, my mistake. In fact, it may be useful to return the x y point

numbers/incompleteBeta.go
rimangan: Yes, this is removed in the newest commit and numbers now passes tests.

cmd/sequelOverlap/intervalQuery.go
rimangan: This now compiles as well in the new commit.

cfauci (approved):

edotau (approved):

LCBartelt (approved):

craiglowe (approved):
Great work!

PR #314, opened 2020-11-11
Simulate rewrite

cfauci:
Uses new genepred struct and has logic for exonFrames and multiple exons with non-coding sequence in between


simulate/simulate.go
```diff
@@ -121,168 +130,489 @@ func changeBase(originalBase dna.Base) dna.Base {
...
+				p += increment - 1
 			}
 		}
+
 	}
+	finalSequence = BaseExtToBases(newSequence)
+	return finalSequence
+}
 
+func CreateCodons(seq []BaseExt, gene *genePred.GenePred, exon int) []CodonExt {
```
rimangan, line 257: function documentation would be a good investment

rimangan (approved):
Looks great! Some documentation would help, and can the large blocks of commented code be removed?

simulate/simulate.go
```diff
@@ -121,166 +132,360 @@ func changeBase(originalBase dna.Base) dna.Base {
...
+	Sequence := make([]BaseExt, len(seq))
+	for i := 0; i < len(seq); i++ {
+		Sequence[i].Base = seq[i]
+		Sequence[i].SeqPos = i
+	}
+	return Sequence
+}
 
+//BaseExtToBases converts a slice of BaseExt to a slice of dna.Base
+func BaseExtToBases(seq []BaseExt) []dna.Base {
```
craiglowe, line 382: I think this is taking bases that have been added in any order, looks at the position in their struct, and then puts them back in order.  I am thinking that you might be using append(), when you could be sticking things directly where they need to be by making a slice of a given length ( whatever = make([]dna.Base, 100) ) and then placing bases (or anything else) directly in it using: whatever[i] = myBase

simulate/simulate.go
```diff
@@ -121,166 +132,360 @@ func changeBase(originalBase dna.Base) dna.Base {
...
+}
+
+//CodonExtsToCodons converts a slice of CodonExt to a slice of dna.Codon
+func CodonExtsToCodons(cE []CodonExt) []dna.Codon {
+	var codons []dna.Codon
+	var codon dna.Codon
+
+	for ext := 0; ext < len(cE); ext++ {
+		for c := 0; c < 3; c++ {
+			codon.Seq[c] = cE[ext].Seq[c].Base
```
craiglowe, line 430: I would have thought you would need to allocate codon.Seq before running this line.  Does it work?  Maybe I am missing something.

craiglowe (commented):
Great work!  I made a few comments

simulate/debug.gpd
```diff
@@ -0,0 +1 @@
+Test	0	+	0	14	2	14	3	2,6,11,	6,11,14,	0,2,0,
```
edotau, line 1: Not the biggest deal but I was wondering if you got the file extension right? I usually see it as `file.gp`, but I am not sure.

simulate/simulate_test.go
```diff
@@ -36,15 +36,30 @@ var MutateSeqTests = []struct {
 	{"debug.fasta", 0.5, "debug.gpd"}, //branch length of 1 gives higher chance of returning a new base so you can see a difference even with a short sequence
```
edotau, line 36: Doesn't have to be apart from this checkin, but I think there is a convention in golang to have a folder called testdata and have these miscellaneous files in that folder. Line 36 would then turn into {"testdata/debug.fasta", 0.5, "testdata/debug.gpd"}

simulate/simulate.go
```diff
@@ -121,166 +132,360 @@ func changeBase(originalBase dna.Base) dna.Base {
...
+		Sequence[i].SeqPos = i
+	}
+	return Sequence
+}
 
+//BaseExtToBases converts a slice of BaseExt to a slice of dna.Base
+func BaseExtToBases(seq []BaseExt) []dna.Base {
+	var newSequence []dna.Base
+
+	for i := 0; i < len(seq); i++ { //counter to put things back in order
```
edotau, line 385: I think this j loop could get pretty crazy, we should consider defining `var i, j int` at the top and avoid the `:=` (colon equals) in nested loops.

simulate/simulate.go
cfauci: i actually wrote that, but didn't end up using it, so it's untested. You're right though, it probably doesn't work. I'll add a test for it because I think it's worth keeping here in case

simulate/simulate.go
cfauci: I was hoping you would have some suggestions like this to help optimize a little bit along the way I know this isnt written super efficiently. I made the update

simulate/simulate.go
```diff
@@ -9,6 +9,17 @@ import (
...
 
+//CodonExt holds a slice of 3 bases and their original positions
+type CodonExt struct {
+	Seq []BaseExt
+}
+
+//BaseExt holds the original position of a base in the starting Seq and the base at that position
+type BaseExt struct {
+	Base   dna.Base
+	SeqPos int
```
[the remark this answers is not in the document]
cfauci: It's used a couple of times, once as an argument as position and once in a loop to restore the original order of the bases

edotau (approved):

rimangan (approved):

craiglowe (approved):
Great work!

PR #316, opened 2020-11-12
clean fastq tests and fix bug in umi

craiglowe:
I was cleaning up some tests in the fastq package and I think I stumbled on a bug in extracting UMIs from TenX reads.  One thing to note is that right now we are expecting a 6 bp UMI, but it seems like most people are currently talking about 10bp UMI's from TenX.  We would need to add an option to handle these cases, which may be most cases going forward.

### DIFF
--- a/fastq/tenXGenomics.go
+++ b/fastq/tenXGenomics.go
@@ -37,9 +37,10 @@ func NextSeq(fwdReader *fileio.EasyReader, revReader *fileio.EasyReader) (*Linke
 //2) Next 6 bases is a 6 base Umi
 //3) finally adaptors and genomic sequence
 //TODO: consider trimming off adapter sequences too
+//TODO: it looks like tenX is now using a 10bp UMI, maybe an option to handle that is needed?
 func FastqPairLinked(fqPair *PairedEnd) *LinkedRead {
-	tenXG := LinkedRead{Fwd: nil, Rev: fqPair.Rev, Bx: GetBarcode(fqPair.Fwd, 0, 16), Umi: GetBarcode(fqPair.Fwd, 17, 23)}
-	tenXG.Fwd = TrimFastq(fqPair.Fwd, 24, len(fqPair.Fwd.Seq))
+	tenXG := LinkedRead{Fwd: nil, Rev: fqPair.Rev, Bx: GetBarcode(fqPair.Fwd, 0, 16), Umi: GetBarcode(fqPair.Fwd, 16, 22)}
+	tenXG.Fwd = TrimFastq(fqPair.Fwd, 22, len(fqPair.Fwd.Seq))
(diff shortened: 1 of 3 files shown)


edotau:
Hmmm I have to take a closer look. This is specific to the 10x Linked-Reads-Whole genome and based on that library construct. I don't think this was intended for the single cell construct.

fastq/tenXGenomics.go
```diff
@@ -37,9 +37,10 @@ func NextSeq(fwdReader *fileio.EasyReader, revReader *fileio.EasyReader) (*Linke
 //2) Next 6 bases is a 6 base Umi
 //3) finally adaptors and genomic sequence
 //TODO: consider trimming off adapter sequences too
+//TODO: it looks like tenX is now using a 10bp UMI, maybe an option to handle that is needed?
```
edotau, line 40: I agree with changing 17 to 16, but not 22 to 24. But to my understand the linked construct is (16+6+128=150)

craiglowe: I don't think there is a 22 -> 24 change being proposed, but maybe I am not seeing it right now.  Did you mean 24 -> 22?  The 24 -> 22 change (old line 42, new line 43) is because I am thinking that the barcode + umi is 22 bases (16+6=22), so the 23rd base is index 22 (zero based).
There are three changes I am proposing:
1) 17 -> 16 for the start of the UMI (inclusive)
2) 23 -> 22 for the end of the UMI (not inclusive)
3) 24 -> 22 for the start of the non-barcode, non-umi read (inclusive)
I am not an expert in the TenX tech, but I think we are on the same page with the first 16 being barcode and the next 6 being the UMI.  I think we just need to make sure we align on the same starts and stops to have the computer parse out the bases we want.

craiglowe (commented):
Quick response to comment.

fastq/tenXGenomics.go
edotau: Yes, I think I wrote that flipped. I think in my head I thought that was changing the 6bp UMI to 10 bp, but now that I look at the math again I think the 16+6 = 22 is right.

Ok I am on board for the changes, but maybe the comment on line 40 should change to say something like this is for 10x linked read and not single cell 10x libraries.

We can certainly do something like pass in 6 or 10 bp for the UMI, but I think in order to avoid confusion the LinkedRead struct should be then renamed to something like TenXReads{} (or something a long that line)

craiglowe (commented):
Expanded on TODO comment concerning scRNA-seq data from 10x.

edotau (approved):
Perfect, thanks!

DanielSnellings (approved):

cfauci (approved):

PR #317, opened 2020-11-13
New simulateVCF with bounded rejection sampling

rimangan:


popgen/afsSimulate.go
```diff
@@ -31,22 +31,21 @@ func SimulateAFS(alpha float64, n int, k int) AFS {
...
-	var freq, r float64
-	var count int
-	var RightBound float64 = 1.0 - AlleleFrequencyBound
+	var maxIteration int = 10000
+	var r, x float64
+	var count, i int
 
-	//stepSize is the function support divided by the number of bins
-	stepSize := (RightBound-AlleleFrequencyBound)/float64(len(binHeights))
+	bound := numbers.ScaledBetaSampler(0.001, 0.5, 5000)
```
craiglowe, line 40: I played around with this code a little bit over the last few days and it seems like this function (at very small x values) bounds the f function in the sense that bound(x) >= f(x) for all x, but for the small x values bound(x) == f(x) == +Infinity.  This is not really the bounding behavior we are hoping for, but I don't think there is an easy way around it since f(x) goes to infinity and I am not sure of an easy way to deal with it in log space.  I think it is not a big deal because the x (allele freq in population) is so small that it never appears in the samples from the population.

craiglowe (approved):
Looks great!

cfauci (approved):

edotau (approved):
Excellent!

To satisfy my curiosity, what is the benefit of returning func() (float64, float64) vs return (float64, float64) ?

PR #320, opened 2020-11-18
0^0 should return 0 == log(1)

craiglowe:
small bug in power calc for 0^0


rimangan (approved):
Great catch!

cfauci (approved):

edotau (approved):
Great! Basic math

PR #321, opened 2020-11-23
Vcf Lift with Reference Fasta

rimangan:


cmd/lift/lift.go
```diff
@@ -27,6 +39,24 @@ func lift(chainFile string, inFile string, outFile string) {
 			log.Fatalf("Multiple overlaps???")
 		}
 		i.UpdateLift(interval.LiftIntervalWithChain(overlap[0].(*chain.Chain), i))
+		//special check for lifting over VCF files
+		if faFile != "" {
+			//faFile will be given if we are lifting over VCF data.
```
craiglowe, line 44: Do we need a check for this to ensure we get an error if an fa is given with a non-vcf, or if a vcf is given without an fa?

cmd/lift/lift.go
```diff
@@ -27,6 +39,24 @@ func lift(chainFile string, inFile string, outFile string) {
...
+			currVcf = i.(*vcf.Vcf)
+
+			//first question: does the "Ref" match the destination fa at this position.
+			if fasta.QuerySeq(records, currVcf.Chr, int(currVcf.Pos - 1), dna.StringToBases(currVcf.Ref)) {
+				//second question: does the "Alt" also match. Can occur in corner cases such as Ref=A, Alt=AAA. Currently we don't invert but write a verbose log print.
+				if fasta.QuerySeq(records, currVcf.Chr, int(currVcf.Pos - 1), dna.StringToBases(currVcf.Alt)) && Verbose > 0 {
```
craiglowe, line 50: This looks like it only works with bi-allelic sites.  Is there a check for that somewhere that gives an error to the user?

craiglowe (commented):
I know this is a work in progress, but I took a look through and had two very small comments

cmd/lift/lift.go
edotau: What is the reasoning for restricting this to only bi-allelic sites?

bed/methods.go
```diff
@@ -48,7 +48,7 @@ func (b BedSlice) Write(file string) {
 
 func (b *Bed) WriteToFileHandle(file *fileio.EasyWriter) {
 	//TODO: write max fields that are non-nil?
-	WriteToFileHandle(file, b, 7)
```
cshoben, line 51: Why was it "7" before?

fasta/compare.go
```diff
@@ -60,3 +60,15 @@ func SortByName(seqs []*Fasta) {
...
+
+//QuerySeq takes in a slice of fastas and a position (name and index) and returns true if a query sequence of bases matches the fasta at this position.
+func QuerySeq(records []*Fasta, chr string, index int, query []dna.Base) bool {
+	chrIndex := GetChromIndex(records, chr)
+	return dna.CompareSeqsIgnoreCaseAndGaps(query, records[chrIndex].Seq[index:index+len(query)]) == 0
+}
+
+
+
+
```
DanielSnellings, line 74: go fmt?

DanielSnellings (approved):

bed/methods.go
rimangan: The Bed format has a number of optional fields after the main positional information which may or may not be initialized in the data structure. The old version of WriteToFileHandle specifies the number of fields to write as a hardcoded number but here we actually check how many fields are initialized in the struct (I added a Fields component to the Bed struct to track this) so we can have a more flexible write function.

bed/bed.go
```diff
@@ -21,6 +21,7 @@ type Bed struct {
 	Name       string
 	Score      int64
 	Strand     bool
+	Fields int
```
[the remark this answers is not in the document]
rimangan: None I think.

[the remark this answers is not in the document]
DanielSnellings: Sweet

DanielSnellings (dismissed):

cfauci (approved):

cmd/lift/lift.go
```diff
@@ -18,30 +28,96 @@ func lift(chainFile string, inFile string, outFile string) {
...
+			i.WriteToFileHandle(un)
+		} else {
+			//DEBUG: interval.PrettyPrint(i)
+			i.UpdateLift(interval.LiftIntervalWithChain(overlap[0].(*chain.Chain), i))//now i is 1-based and we can assert VCF.
+			//DEBUG: interval.PrettyPrint(i)
+			//special check for lifting over VCF files
+			if faFile != "" {
+				//faFile will be given if we are lifting over VCF data.
+				currVcf = i.(*vcf.Vcf)
+				if utf8.RuneCountInString(currVcf.Ref) > 1 {
```
craiglowe, line 71: Do we also need to check Alt?  I see the log fatal says we don't handle indels (which is fine), but I am picture the ref is A and the alt is ACGT, which I don't think get caught by this check, even though it is an indel

cmd/lift/lift.go
```diff
@@ -18,30 +28,96 @@ func lift(chainFile string, inFile string, outFile string) {
...
+				}
+			}
+			i.WriteToFileHandle(out)
 		}
-		i.UpdateLift(interval.LiftIntervalWithChain(overlap[0].(*chain.Chain), i))
-		i.WriteToFileHandle(out)
 	}
 }
 
+//minMatchPass returns true if the interval/chain has over a 95 percent base match, false otherwise.
```
craiglowe, line 96: Should 95 be changed to minMatch?

interval/lift.go
```diff
@@ -11,3 +17,95 @@ type Lift interface {
...
+	answer := make(chan Lift, 1000)
+	go ReadToLiftChan(inputFile, answer)
+	return answer
+}
+
+func ReadToLiftChan(inputFile string, send chan<- Lift) {
+	// How the file is read is dependent on the file extension
+	filetype := path.Ext(inputFile)
+
+	if filetype == ".gz" {
```
craiglowe, line 32: This is cool.  Seems like it could be useful other places.  Maybe in the future we could have a standard function in fileio that takes a filename and returns the type?

interval/lift.go
```diff
@@ -11,3 +17,95 @@ type Lift interface {
...
+	var currPos int = c.TStart//starting with strand +/+ case for now.
+	if !c.TStrand {
+		log.Fatalf("Please format chain files for lift with the target in the positive strand.")
+	}
+	for j := 0; j < len(c.Alignment); j++ {
+		match += matchOverlapLen(currPos, currPos + c.Alignment[j].Size, i.GetChromStart(), i.GetChromEnd())
+		currPos += c.Alignment[j].Size
+		dT += matchOverlapLen(currPos, currPos + c.Alignment[j].TBases, i.GetChromStart(), i.GetChromEnd())
+		if matchOverlapLen(currPos, currPos + c.Alignment[j].TBases, i.GetChromStart(), i.GetChromEnd()) > 0 {
+			dQ += c.Alignment[j].QBases//this handles a special case where both TBases and QBases are non-zero, as in competing non-aligned bases are present at the same location in target and query.
```
craiglowe, line 101: Minor comment, but these long comment lines can sometimes be a little difficult to read.  Might be good to break it into multiple lines

craiglowe (commented):
Great Work

cmd/lift/lift.go
```diff
@@ -18,30 +28,96 @@ func lift(chainFile string, inFile string, outFile string) {
...
+	un := fileio.EasyCreate(unMapped)
+	defer un.Close()
+
+	var records []*fasta.Fasta
+	var currVcf *vcf.Vcf
+	var a, b float64
+
+	if faFile != "" {
+		if !vcf.IsVcfFile(inFile) {
+			log.Fatalf("Fasta files should be provided when lifting VCF files. inFile is not a valid VCF file and must end in either .vcf or .vcf.gz.")
```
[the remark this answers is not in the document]
rimangan: In this case, Fa is provided but we are not lifting a VCF, so the fasta is unnecessary. This throws an error in this case. I have clarified the language in the next commit.

cmd/lift/lift.go
```diff
@@ -18,30 +28,96 @@ func lift(chainFile string, inFile string, outFile string) {
...
+
 	//second task, read in intervals, find chain, and convert to new interval
 	inChan := interval.GoReadToLiftChan(inFile)
 	var overlap []interval.Interval
 	for i := range inChan {
-		overlap = interval.Query(tree, i, "any")
+		overlap = interval.Query(tree, i, "any")//TODO: verify proper t/q contains
 		if len(overlap) > 1 {
-			log.Fatalf("Multiple overlaps???")
+			fmt.Fprintf(un, "Record below maps multiple chains:\n")
```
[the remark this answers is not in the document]
rimangan: yep

cmd/lift/lift.go
```diff
@@ -18,30 +28,96 @@ func lift(chainFile string, inFile string, outFile string) {
...
+			i.WriteToFileHandle(un)
+		} else if len(overlap) == 0 {
+			fmt.Fprintf(un, "Record below has no ortholog in new assembly:\n")
+			i.WriteToFileHandle(un)
+		} else if !minMatchPass(overlap[0].(*chain.Chain), i, minMatch) {
+			a, b = interval.MatchProportion(overlap[0].(*chain.Chain), i)
+			fmt.Fprintf(un, "Record below fails minMatch with a proportion of %f:\n", numbers.MinFloat64(a, b))
```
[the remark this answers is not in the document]
rimangan: great idea! Fixed in next commit

interval/lift.go
rimangan: yes I think this should be reworked in the future for a more general solution that switches on type, as you said

LCBartelt (approved):

craiglowe (approved):
Great work!  Thanks for thinking about all the comments!

DanielSnellings (approved):

PR #323, opened 2020-11-30
VCF Unification Project

rimangan:
Not ready for review. Updating functions for a proposed VCF format using DNA Ref and Alt fields. This may be renamed to a Variant struct if we decide to reserve the Vcf name to a struct with string Ref and Alt fields.


vcf/vcf.go
```diff
@@ -14,15 +15,16 @@ import (
 //slices of slice
 type Vcf struct {
 	Chr    string
-	Pos    int64
+	Pos    int
 	Id     string
-	Ref    string
-	Alt    string
+	Ref    []dna.Base
+	Alt    [][]dna.Base
```
edotau, line 21: Something to consider for the 2D dna.Base slice is to instead use a bytes.Buffer and convert the bytes.Buffer into the [][]dna.Base only when necessary because I think in general most SNPs would only occupy one byte and the number of Alt alleles don't extend out too far either.

dna/examine.go
```diff
@@ -44,6 +44,16 @@ func CountMask(seq []Base) (unmaskedCount int, maskedCount int, gapCount int) {
 	return unmaskedCount, maskedCount, gapCount
 }
 
+func CountNs(seq []Base) int {
```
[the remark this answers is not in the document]
rimangan: oh must have missed this function. This is way better thanks

convert/convert_test.go
```diff
@@ -5,16 +5,17 @@ import (
...
 
 var seqA []dna.Base = dna.StringToBases("--TTTC--ATGAATAA")
 var seqB []dna.Base = dna.StringToBases("CCATTCCAA--CAGA-")
 var inputFa []*fasta.Fasta = []*fasta.Fasta{{"eggplant", seqA}, {"squash", seqB}}
-var var1 vcf.Vcf = vcf.Vcf{Chr: "chr1", Pos: 1, Id: ".", Ref: "T", Alt: "A", Qual: 100.0, Filter: "PASS", Info: ".", Format: ".", Notes: "."}
-var var2 vcf.Vcf = vcf.Vcf{Chr: "chr1", Pos: 4, Id: ".", Ref: "C", Alt: "CCA", Qual: 100.0, Filter: "PASS", Info: ".", Format: ".", Notes: "."}
-var var3 vcf.Vcf = vcf.Vcf{Chr: "chr1", Pos: 5, Id: ".", Ref: "ATG", Alt: "A", Qual: 100.0, Filter: "PASS", Info: ".", Format: ".", Notes: "."}
-var var4 vcf.Vcf = vcf.Vcf{Chr: "chr1", Pos: 8, Id: ".", Ref: "A", Alt: "C", Qual: 100.0, Filter: "PASS", Info: ".", Format: ".", Notes: "."}
-var var5 vcf.Vcf = vcf.Vcf{Chr: "chr1", Pos: 10, Id: ".", Ref: "T", Alt: "G", Qual: 100.0, Filter: "PASS", Info: ".", Format: ".", Notes: "."}
+var var1 vcf.Vcf = vcf.Vcf{Chr: "chr1", Pos: 1, Id: ".", Ref: "T", Alt: strings.Split("A", ","), Qual: 100.0, Filter: "PASS", Info: "."}
```
edotau, line 14: Not sure if this is the best way to handle the alt. Another "type" of Vcf you can have is a "structure variance" vcf file and that is not separated by commas. See page pages 12-18 [https://samtools.github.io/hts-specs/VCFv4.2.pdf](https://samtools.github.io/hts-specs/VCFv4.2.pdf)

vcf/vcfTools.go
```diff
@@ -69,19 +70,13 @@ func NoVcfOverlap(vcfs []*Vcf) {
 }
 
 func Snp(v *Vcf) bool {
```
edotau, line 72: This is for genome graph stuff, so can we leave these the way it is.

vcf/vcf_test.go
```diff
@@ -40,10 +40,11 @@ func TestWriteAndRead(t *testing.T) {
 		if !AllEqual(alpha, beta) {
 			t.Errorf("Error: Read and write files do not match\n")
 		}
-		os.Remove(tempFile)
+		//os.Remove(tempFile)
 	}
 }
 
+/*
```
edotau, line 47: Why is this commented out?

edotau (commented):
Yeah my biggest concern is to make sure you run `go test ./...` to make sure you are not breaking other dependencies

rimangan:
> Yeah my biggest concern is to make sure you run `go test ./...` to make sure you are not breaking other dependencies

yes this is my current objective. Currently we pass tests within VCF but there are plenty of external functions that use the old VCF struct components.

axt/callSnps.go
```diff
@@ -21,33 +21,33 @@ func AxtVcfToFile(filename string, axtList []*Axt, fa []*fasta.Fasta) {
...
 		if axtFile.RSeq[i] == dna.Gap {
 
 			qCount++
-			curr = &vcf.Vcf{Chr: axtFile.RName, Pos: rCount, Id: axtFile.QName, Ref: dna.BaseToString(dna.ToUpper(axtFile.RSeq[i-1])), Alt: dna.BaseToString(dna.ToUpper(axtFile.QSeq[i-1])), Qual: 24, Filter: "PASS", Info: fmt.Sprintf("query=%d", qCount), Format: "SVTYPE=INS", Notes: AxtInfo(axtFile)}
+			curr = &vcf.Vcf{Chr: axtFile.RName, Pos: rCount, Id: axtFile.QName, Ref: dna.BaseToString(dna.ToUpper(axtFile.RSeq[i-1])), Alt: []string{dna.BaseToString(dna.ToUpper(axtFile.QSeq[i-1]))}, Qual: 24, Filter: "PASS", Info: fmt.Sprintf("query=%d;SVTYPE=SNP;%s", qCount, AxtInfo(axtFile))}
 
 			for j := i; j < len(axtFile.RSeq); j++ {
 				if dna.ToUpper(axtFile.RSeq[j]) == dna.Gap {
-					curr.Alt += dna.BaseToString(dna.ToUpper(axtFile.QSeq[j]))
+					curr.Alt = append(curr.Alt, dna.BaseToString(dna.ToUpper(axtFile.QSeq[j])))
```
edotau, line 44: not sure if we should append this one by one. What is happening is we are looping forward and checking if there are bases for the insertion/deletion. It should follow the same `+=` logic.

edotau (approved):
LGTM! Lets start getting this checked in and we can start to take down the bugs together.

cmd/vcfToBed/vcfToBed.go
```diff
@@ -7,21 +7,18 @@ import (
...
 	out := fileio.EasyCreate(outfile)
 	defer out.Close()
 
 	for v := range ch {
-		words := strings.Split(v.Notes, "\t")
-		note = words[0]
-		for i := 0; i < len(words); i++ {
-			note = note + delimiter + words[i]
+		for i := 0; i < len(v.Samples); i++ {
+			note = note + delimiter + vcf.HelperSamplesToString(v.Samples, i)
```
craiglowe, line 21: Maybe SamplesToString?  A helper function would usually be a non-public function that is called by a more user-friendly function, often as the recursive step.
Does "note" need to get reset before entering this loop?  Otherwise it seems like "note" would get bigger and bigger as it goes over all the values coming out of the channel

convert/convert_test.go
craiglowe: I think the structure variance alleles are still comma separated.  You can see what I believe is an example at the bottom of section 5.4.3 in the document linked above.

gtf/annotation.go
```diff
@@ -39,7 +40,7 @@ func cDnaToString(v *vcfEffectPrediction, seq map[string][]dna.Base) string {
 func nonCodingToString(v *vcfEffectPrediction, seq map[string][]dna.Base) string {
 	var answer string = v.RefId + ":c."
 	ref := dna.StringToBases(v.Ref)
-	alt := dna.StringToBases(v.Alt)
+	alt := dna.StringToBases(v.Alt[0])//TODO: (riley) This looks like it was written for biallelic positions originally, I have doubled down on this for now but polyallelic bases should be handled.
```
craiglowe, line 43: Should we also add this expectation (biallelic only) to the main function comment?

DanielSnellings (approved):

craiglowe (commented):
Wonderful work!  Exciting to see this.  I had a few comments.  Made it through the first half and still going.

popgen/stationarity.go
```diff
@@ -53,24 +53,23 @@ func MultiFaToAFS(aln []*fasta.Fasta) AFS {
...
+	for i := range alpha {
 		currentSeg = &SegSite{i: 0, n: 0}
 		//gVCF converts the alt and ref to []DNA.base, so structural variants with <CN0> notation will fail to convert. This check allows us to ignore these cases.
-		if !strings.ContainsAny(i.Alt, "<>") {
-			g := vcf.VcfToGvcf(i)
-			for j = 0; j < len(g.Genotypes); j++ {
-				if g.Genotypes[j].AlleleOne != -1 && g.Genotypes[j].AlleleTwo != -1 { //check data for both alleles exist for sample.
+		if !strings.ContainsAny(i.Alt[0], "<>") {//By definition, segregting sites are biallelic, so we only check the first entry in Alt.
+			for j = 0; j < len(i.Samples); j++ {
+				if i.Samples[j].AlleleOne != -1 && i.Samples[j].AlleleTwo != -1 { //check data for both alleles exist for sample.
```
craiglowe, line 67: This is the second time I am seeing -1 come up.  I am thinking it means "empty," but I am not sure if it is an "allowable" value, or if it is an error.

vcf/filter.go
```diff
@@ -27,59 +28,74 @@ func Filter(v *Vcf, chrom string, minPos int64, maxPos int64, ref string, alt st
...
 	if v.Qual < minQual {
 		return false
 	}
 	return true
 }
 
-func FilterAlt(v *Vcf, alt string) bool {
-	if alt != "" && v.Alt != alt {
+//FilterAlt returns true if the Alt field of a Vcf entry matches a desired input Alt field, false otherwise
+func FilterAlt(v *Vcf, alt []string) bool {
```
craiglowe, line 40: Would a name like "AltMatches" or "AltEquals" be better?  It looks like an empty alt being passed in would match everything?  It also looks like the alt fields need to be in the same order, so C,G does not equal G,C?  Maybe a comment that order matters?

craiglowe (commented):
Few more comments.

vcf/filter.go
```diff
@@ -174,27 +208,36 @@ func FilterVcfPos(vcfs []*Vcf) []*Vcf {
...
 		for i = 0; i < len(curr); i++ {
 			if encountered[curr[i].Pos] == true {
 				//do not add
 			} else {
 				encountered[curr[i].Pos] = true
+				containsN = false
 				ref = dna.StringToBases(curr[i].Ref)
-				alt = dna.StringToBases(curr[i].Alt)
-				if dna.CountBaseInterval(ref, dna.N, 0, len(ref)) == 0 && dna.CountBaseInterval(alt, dna.N, 0, len(alt)) == 0 {
+				alt = GetAltBases(curr[i].Alt)
```
craiglowe, line 228: Not for now, but this seems to be similar code to what I saw in a function above.  I bet we could reduce code by having a ContainsN(vcf) bool
function that would check for Ns and give a true/false.

vcf/fix.go
```diff
@@ -19,18 +19,22 @@ func FixVcf(query *Vcf, ref map[string][]dna.Base) {
...
 // according to VCF specs
 func fixDash(query *Vcf, ref map[string][]dna.Base) {
-	if strings.Compare(query.Alt, "-") == 0 {
-		// query.Pos is -2, one for zero base, and one for previous base
-		prevBase := dna.BaseToString(ref[query.Chr][query.Pos-2])
-		query.Pos--
-		query.Ref = prevBase + query.Ref
-		query.Alt = prevBase
+	for i := 0; i < len(query.Alt); i++ {//TODO: (Riley) right now we're just checking if any of the alts have dashes. Not sure if this is the best way to handle polyallelic sites for fix. 
+			if strings.Compare(query.Alt[i], "-") == 0 {
```
craiglowe, line 23: go fmt?

vcf/fix.go
```diff
@@ -19,18 +19,22 @@ func FixVcf(query *Vcf, ref map[string][]dna.Base) {
...
+			// query.Pos is -2, one for zero base, and one for previous base
+			prevBase := dna.BaseToString(ref[query.Chr][query.Pos-2])
```
craiglowe, line 25: can query.Pos-2 be negative?

vcf/fix.go
```diff
@@ -19,18 +19,22 @@ func FixVcf(query *Vcf, ref map[string][]dna.Base) {
...
+			query.Pos--
+			query.Ref = prevBase + query.Ref
+			query.Alt[i] = prevBase
+		}
 	}
 	if strings.Compare(query.Ref, "-") == 0 {
 		// query.Pos is -2, one for zero base, and one for previous base
 		prevBase := dna.BaseToString(ref[query.Chr][query.Pos-2])
```
craiglowe, line 33: Can query.Pos-2 be negative?

vcf/gVcf.go
```diff
@@ -192,63 +191,47 @@ func PhasedToString(phased bool) string {
...
-		return "NoData "
-	} else {
-		if i == len(sample.Genotypes)-1 {
-			return fmt.Sprintf("%d%s%d=%s%s%s", sample.Genotypes[i].AlleleOne, PhasedToString(sample.Genotypes[i].Phased), sample.Genotypes[i].AlleleTwo, dna.BasesToString(sample.Seq[sample.Genotypes[i].AlleleOne]), PhasedToString(sample.Genotypes[i].Phased), dna.BasesToString(sample.Seq[sample.Genotypes[i].AlleleTwo]))
-		} else {
-			return fmt.Sprintf("%d%s%d=%s%s%s\t", sample.Genotypes[i].AlleleOne, PhasedToString(sample.Genotypes[i].Phased), sample.Genotypes[i].AlleleTwo, dna.BasesToString(sample.Seq[sample.Genotypes[i].AlleleOne]), PhasedToString(sample.Genotypes[i].Phased), dna.BasesToString(sample.Seq[sample.Genotypes[i].AlleleTwo]))
-		}
-	}
+	vReorder := ReorderSampleColumns(v, samples)
+	log.Printf("%s\t%d\t%s\t%s\t%s\n", v.Chr, v.Pos, v.Ref, v.Alt, SamplesToString(vReorder.Samples))
```
craiglowe, line 206: Is this for debugging?

vcf/gVcf.go
```diff
@@ -192,63 +191,47 @@ func PhasedToString(phased bool) string {
...
 	return answer
 }
 
 //helperGenotypeToStringNew uses just an array of GenomeSample structs to write to a string for simple gVCFs with just the allele info in notes.
-func helperGenotypeToStringNew(sample []GenomeSample, i int) string {
+func HelperSamplesToString(sample []GenomeSample, i int) string {
+	var answer string
 	if sample[i].AlleleOne < 0 {
-		return "noData "
+		answer = "noData "
```
craiglowe, line 221: No space at end?

vcf/vcf.go
```diff
@@ -111,6 +174,21 @@ func ReadHeader(er *fileio.EasyReader) *VcfHeader {
 	return &header
 }
 
+//FormatToString converts the []string Format struct into a string by concatenating with a colon delimiter.
+func FormatToString(format []string) string {
+	if len(format) == 0 {
+		return ""
+	}
+	if len(format) == 1 {
```
craiglowe, line 182: Do we need this if statement and return?  Seems like len(format) == 1 is handled ok by the below code, which also deals with larger lengths?

craiglowe (commented):
Wonderful work!

convert/convert_test.go
rimangan: I think we should table this issue for now until someone has need of a structure variance allele parser, at which point it makes sense to rework some of this.

vcf/filter.go
rimangan: I'll keep the name consistent with the other filter functions for now but add a comment that order matters.

vcf/fix.go
rimangan: I only changed castings here, this logic is not something I wrote.

vcf/gVcf.go
rimangan: Yes this is a debugging function

vcf/gVcf.go
rimangan: Eric set this up with the space, not sure if it has significance. I'll remove.

vcf/vcf.go
rimangan: great catch!

cfauci (dismissed):

vcf/vcf.go
```diff
@@ -8,27 +13,38 @@ import (
...
 	Qual   float64
 	Filter string
 	Info   string
-	Format string
-	Notes  string
+	Format []string
+	Samples []GenomeSample
 }
 
+//GenomeSample is a substruct of Vcf, and contains information about each individual represented in a VCF line.
```
DanielSnellings, line 34: can you add a comment noting that -1 is nil for AlleleOne and AlleleTwo

craiglowe (dismissed):
Great work!

edotau (approved):
Lets try this again...so close!!

DanielSnellings (approved):

edotau (approved):

craiglowe (approved):
Last few changes look good

PR #324, opened 2020-12-02
GoGene draft

DanielSnellings:
For awhile I have been thinking about making an internal data structure for gonomics that reduces the amount of bookkeeping required to traverse and manipulate genes. This is a draft data structure. I would be interested in hearing everyone's opinion. 

The basic idea is to process a GTF record into a struct that holds the genomic sequence, the cDNA sequence, and creates an easy way to convert coordinates between the two sequences (see featureArray). This has several benefits most of which revolve around reducing the number of corner cases that you need to consider when doing something like variant effect prediction (e.g. you don't need to worry about strand, exon hopping, genomic -> cDNA position conversion, and probably more I haven't thought about). Ultimately I would like to refactor my variant effect prediction code to use GoGenes instead of GTF records, but I wanted to run this by everyone to solicit comments and suggestions before I build too much off of this code.


edotau (commented):
This is pretty cool. I think you should connect to Christi. The feature array aspect of the struct is basically what is going on the genePred format. I think the difference is that it stores the calculated base positions rather than the increments you programed.

goGene/goGene.go
```diff
@@ -0,0 +1,137 @@
...
+// traversal and manipulation of genes on the genomic and mRNA levels.
+type GoGene struct {
+	id           string
+	startPos     int
+	strand       bool
+	cdsStarts    []int
+	genomeSeq    []dna.Base
+	cdnaSeq      []dna.Base
+	featureArray []Feature
+}
```
craiglowe, line 38: From looking ahead in the code I see that genomeSeq points into the fasta sequence, while cdnaSeq has its own memory allocated.  If I am understanding everything correctly, that means that genomeSeq is memory efficient since it is really only a pointer into the fasta sequence.  I think the downside of this could be some difficult to trackdown bugs and possibly mysterious behavior.  For example, if someone calls myGene.genomeSeq[100] = dna.A, they will not only be setting that base to A in the gene "myGene", but will also be editing the genome and editing that base position in any overlapping isoforms or genes (but not changing any cdnaSeq).  I wonder if storing a transcription end (or transcription length) would give you the same info as having genomeSeq, but make it more clear that the genome sequence was only being stored once (the fasta file).

craiglowe (commented):
Very cool!  I think it is a great idea to come up with a nice struct to better handle genes!

goGene/goGene.go
```diff
@@ -0,0 +1,137 @@
...
+	var queryPos int
+	if g.strand { // Positive Strand
+		queryPos = genomePos - g.startPos
+	} else { // Negative Strand
+		queryPos = g.startPos - genomePos
+	}
+
+	feature := g.featureArray[queryPos]
+	switch feature {
+	case -1:
```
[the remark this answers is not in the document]
DanielSnellings: Nice catch, I agree. Change in next push.

goGene/goGene.go
DanielSnellings: I thought that reslicing to a new variable created a copy rather than a pointer, but after testing in Go Playground I see that you are right. My intention was to make a copy so the sequence could be easily changed. I have switch to using `copy`

PR #325, opened 2020-12-03
generalized linked read struct name and funcs to handle rnaseq tenx too

craiglowe:
First pass at generalizing the linked read funcs

### DIFF
--- a/fastq/tenXGenomics.go
+++ b/fastq/tenXGenomics.go
@@ -6,42 +6,33 @@ import (
 	"github.com/vertgenlab/gonomics/fileio"
 )
 
-type LinkedRead struct {
+type TenXPair struct {
 	Fwd *Fastq
 	Rev *Fastq
 	Bx  []dna.Base
 	Umi []dna.Base
 }
 
-func ReadToChanLinked(fileOne string, fileTwo string, tenXg chan<- *LinkedRead) {
+func ReadToChanTenXPair(fileOne string, fileTwo string, barcodeLength int, umiLength int, tenXg chan<- *TenXPair) {
 	fwd, rev := fileio.EasyOpen(fileOne), fileio.EasyOpen(fileTwo)
 	defer fwd.Close()
 	defer rev.Close()
 
-	for curr, done := NextSeq(fwd, rev); !done; curr, done = NextSeq(fwd, rev) {
-		tenXg <- curr
+	for curr, done := NextFastqPair(fwd, rev); !done; curr, done = NextFastqPair(fwd, rev) {
+		tenXg <- FastqPairToTenXPair(curr, barcodeLength, umiLength)
 	}
 	close(tenXg)
-}
-
-func NextSeq(fwdReader *fileio.EasyReader, revReader *fileio.EasyReader) (*LinkedRead, bool) {
-	curr, done := NextFastqPair(fwdReader, revReader)
-	if done {
-		return nil, true
-	}
-	return FastqPairLinked(curr), false
 }
 
 //10x linked-read library construct are modifications only on forward or read one, read two remains unchanged:
 //1) first 16 bases: 10x barcode labeled either Bx or Rx tags
 //2) Next 6 bases is a 6 base Umi
 //3) finally adaptors and genomic sequence
+//10x scRNA-seq is also a 16bp barcode (for the cell), and then UMIs of various lengths
 //TODO: consider trimming off adapter sequences too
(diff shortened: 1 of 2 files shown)


fastq/tenXGenomics.go
```diff
@@ -6,42 +6,33 @@ import (
 	"github.com/vertgenlab/gonomics/fileio"
 )
 
-type LinkedRead struct {
+type TenXPair struct {
 	Fwd *Fastq
 	Rev *Fastq
 	Bx  []dna.Base
```
cshoben, line 12: What does “Bx” mean or stand for?

fastq/tenXGenomics.go
```diff
@@ -6,42 +6,33 @@ import (
...
 
-func ReadToChanLinked(fileOne string, fileTwo string, tenXg chan<- *LinkedRead) {
+func ReadToChanTenXPair(fileOne string, fileTwo string, barcodeLength int, umiLength int, tenXg chan<- *TenXPair) {
 	fwd, rev := fileio.EasyOpen(fileOne), fileio.EasyOpen(fileTwo)
 	defer fwd.Close()
 	defer rev.Close()
 
-	for curr, done := NextSeq(fwd, rev); !done; curr, done = NextSeq(fwd, rev) {
-		tenXg <- curr
+	for curr, done := NextFastqPair(fwd, rev); !done; curr, done = NextFastqPair(fwd, rev) {
```
cshoben, line 21: Can you help me understand what the “!done” is doing?

fastq/tenXGenomics.go
```diff
@@ -93,12 +84,12 @@ func fastqStats(fq *Fastq) string {
 	return fmt.Sprintf("%s\t%d\n%s\n", fq.Name, len(fq.Seq), dna.BasesToString(fq.Seq))
```
cshoben, line 84: Just wondering, is there a reason we tab between the name in length versus putting it on a new line? Is there a goal maybe to keep each entry only two lines? And I assume the new line at the end of this is to aid in human legibility downstream?

fastq/tenXGenomics.go
```diff
@@ -6,42 +6,33 @@ import (
...
-type LinkedRead struct {
+type TenXPair struct {
 	Fwd *Fastq
 	Rev *Fastq
 	Bx  []dna.Base
 	Umi []dna.Base
 }
 
-func ReadToChanLinked(fileOne string, fileTwo string, tenXg chan<- *LinkedRead) {
+func ReadToChanTenXPair(fileOne string, fileTwo string, barcodeLength int, umiLength int, tenXg chan<- *TenXPair) {
```
DanielSnellings, line 16: I am approving as this may be beyond the scope of this pull request, however, from Eric's testing it seems the consensus is that chans sending pointers is generally bad. How do you feel about slowly adjusting this function and the rest of the `ReadToChan` functions in gonomics to passing structs instead of pointers?

DanielSnellings (approved):
One broad scope question, but approving.

fastq/tenXGenomics.go
edotau: I can't remember the reason why, but if we are tending in this way, i think the same is true for sending a struct with pointers as data fields.

edotau: As I clicked enter I started to remember, pointers to structs (`*LinkedRead`) and pointers fields (`*Fwd, *Rev`) will be allocated on the heap by default for the duration of the program. We would only consider using sending a pointer through channels if we need to later modify the struct that it originally pointed to (which is really uncommon).

edotau (approved):
LGTM! Let's get this checked in.

rimangan (approved):

RavenYL (approved):

fastq/tenXGenomics.go
cfauci: Seems like fastq format is hardcoded to read and expect the seq on the second line of the record (with name being the first line) to change that would probably be a problematic task and that would only be useful for TenX reads, this way the name can be the first and the seq the second and you can store the length as well (that what it seems like is going on here). The way we have every file type coded they all need to end in a new line. If there are more fastqs to put in the file, it will write to this new line created in the record before it, but at the end of the file there should be an empty new line, and every reader we've built (at least that i've encountered or written) will fail if the file doesnt end in a new line character.

cfauci (approved):

cshoben (approved):

PR #327, opened 2020-12-08
Multi line tree

cfauci:
Adding ReadMultiLineTree to both tree.go and expandedTree.go. Processes the multi line tree with \n into a single line tree that can be read as normal by ReadNewick. EDIT: function was useful for any file type, so it's been moved to fileio and that is now called for the tree and expandedTree packages to read a multi-line tree.


expandedTree/expandedTree.go
```diff
@@ -36,24 +36,35 @@ func ReadTree(newickFilename string, fastasFilename string) *ETree {
 
 //read in tree from filename
 func ReadNewick(filename string) (*ETree, error) {
-	var line string
+	var singleLineTree string
```
craiglowe, line 39: Just an FYI that you can put this all on one line if you like:
var singleLineTree string = ReadMultiLineTree(filename)

expandedTree/expandedTree.go
```diff
@@ -36,24 +36,35 @@ func ReadTree(newickFilename string, fastasFilename string) *ETree {
...
-			return parseNewick(line[strings.Index(line, "("):(1 + strings.LastIndex(line, ";"))])
-		}
+	if !strings.HasPrefix(singleLineTree, "#") {
+		return parseNewick(singleLineTree[strings.Index(singleLineTree, "(") : 1+strings.LastIndex(singleLineTree, ";")])
 	}
+
 	return nil, errors.New("Error: tree file is either empty or has no non-comment lines")
 }
 
+func ReadMultiLineTree(input string) string {
```
craiglowe, line 55: This function looks great!
I think you should consider changing the name or adding a little to the function.  What this function really does is more general that just reading a tree file into a single string (with new lines removed); the function reads *any* file into a string (with line breaks removed).  One option would be to move this into the fileio package and rename it to something like ReadFileToSingleLine.  Another option would be to run the checks and call parseNewick within this function so that it is specific to trees and gives a tree back.

simulate/simulate_test.go
```diff
@@ -24,7 +24,7 @@ func TestRandGene(t *testing.T) {
 		if len(a[0].Seq) != test.length {
 			t.Errorf("expected RandGene to give %v, gave %v", test.length, len(a[0].Seq))
 		}
-		//fmt.Print(dna.BasesToString(a[0].Seq), "\n")
+		fmt.Print(dna.BasesToString(a[0].Seq), "\n")
```
craiglowe, line 27: Can we leave this commented out?  When we scroll the screen during tests, I think it is harder to notice what fails.  That way we have a test, when it fails someone can go into the file and uncomment this to try and debug it.

tree/tree.go
```diff
@@ -117,6 +116,19 @@ func splitNameAndLength(input string) (string, float64, bool, error) {
 	return "", 0, false, fmt.Errorf("Error: %s should only have one or two colons\n", input)
 }
 
+func ReadMultiLineTree(input string) string {
```
craiglowe, line 119: I think this also handles single line tree files, right?  I think this should actually replace the old version of ReadNewick, rather than become its own function.  I think this because it more improves something that already exists.  That way we don't end up with two read functions for trees that we need to maintain.

craiglowe (commented):
Great work!  A few small comments

expandedTree/expandedTree.go
cfauci: I like the idea of moving it to a larger package a lot, I hadn't thought of that, can we get a general consensus on that at update Thursday this week then and then I'll push that change rather than this one?

tree/tree.go
cfauci: I think moving this to the fileio package makes sense and then calling it in readNewick makes the most sense.

expandedTree/expandedTree.go
craiglowe: I think you can take a shot at the solution you like better.  You can either:
1) keep it general (any file type to a string of everything in that file with no line breaks) and put it in fileio with a general name, or
2) you can leave it here, but make it specific to reading trees by putting the conversion from string to tree inside this function.
You can ask some other people to review this pull request and get their opinion as well.

tree/tree.go
craiglowe: You can move this to fileio and name it with a general name like ReadFileToSingleLineString, but I would also update ReadNewick so that it calls your newly created fileio.ReadFileToSingleLineString and therefore handles both multi-line and single-line trees.

craiglowe (commented):
Added a couple quick responses

tree/tree.go
cfauci: Yes, just to clarify though, these functions are identical (the tree.go and expandedTree.go functions) so I think I'm going to keep the general option and call the new fileio general function in both cases.

edotau:
We have to deal with a few merging conflicts but otherwise LGTM!

tree/tree.go
```diff
@@ -168,20 +167,17 @@ func parseNewickHelper(input string) (*Tree, error) {
...
 	}
 	defer file.Close()
-	scanner := bufio.NewScanner(file)
 
-	for scanner.Scan() {
-		line = scanner.Text()
-		if !strings.HasPrefix(line, "#") {
-			return ParseNewick(line[strings.Index(line, "("): 1+strings.LastIndex(line, ";")])
-		}
+		if !strings.HasPrefix(singleLineTree, "#") {
```
craiglowe, line 179: Looks like this could use a "go fmt"

craiglowe (approved):
Wonderful!  I made a small formatting comment, but that is not holding up my approval.

expandedTree/expandedTree.go
rimangan: I'm happy with either solution, but I also quite like the idea of this as a fileio function. I think reading files to single strings might be useful in other contexts.

rimangan (approved):

rimangan:
Conflict with tree.go has yet to be resolved.

PR #328, opened 2020-12-08
GoGenes

DanielSnellings:
This pull request adds GoGenes which are an effort to simplify evaluation and manipulation of genes. GoGenes can be created from any input GTF record. The GoGene struct is intentionally designed to hold only private fields. Instead of directly manipulating the struct, manipulation of GoGenes should be done through public functions which handle all of the minutia of working with a GTF or genePred file (strand, introns, coordinate systems, etc.). Current functions provide easy conversion between genomic and cDNA coordinates, as well as making point mutations, insertions, and deletions. As designed, the PointMutation, Insertion, and Deletion functions will all return an EffectPrediction struct that will record how the mutation affected the resulting protein. The EffectPrediction functionality has been completed for point mutations, however effect prediction for indels is still a WIP. 

This is a relatively long pull request (sorry!), but more than half the added lines are from test files so it shouldn't be too bad. 

Suggestions to improve the struct are welcome!


dna/modify.go
```diff
@@ -96,6 +96,11 @@ func Complement(bases []Base) {
 	}
 }
 
+func ComplementSingleBase(base Base) Base {
```
craiglowe, line 99: This looks like a public version of the "complement" function.  Can we just rename the existing "complement" function to "ComplementSingleBase"?

goGene/fromGtf.go
```diff
@@ -0,0 +1,122 @@
+package goGene
+
+import (
+	"github.com/vertgenlab/gonomics/dna"
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/gtf"
+)
+
+// GtfToGoGene converts a gtf record into a GoGene data structure
+// WARNING: If multiple isoforms are present, only the longest is used
```
craiglowe, line 10: Longest before or after splicing?

goGene/fromGtf.go
```diff
@@ -0,0 +1,122 @@
...
+func GtfToGoGene(g *gtf.Gene, ref []*fasta.Fasta) *GoGene {
+	answer := new(GoGene)
+	gtf.MoveCanonicalToZero(g)
+	transcript := g.Transcripts[0]
+
+	answer.id = g.GeneID
+	answer.strand = transcript.Strand
+
+	if transcript.Strand { // If on the positive strand
```
craiglowe, line 19: Is transcript.PosStrand more clear?

craiglowe (commented):
Very cool!  Only had a new minutes to start looking at this.  I only made it about 1/4 of the way through, but figured I would send my comments now because I am about to jump into a stretch of meetings.

goGene/fromGtf.go
```diff
@@ -0,0 +1,122 @@
...
+		answer.featureArray = make([]Feature, len(answer.genomeSeq))
+		answer.cdsStarts = make([]int, 0, len(transcript.Exons))
+		answer.cdsEnds = make([]int, 0, len(transcript.Exons))
+
+		var i int
+		var prevExonEnd int = answer.startPos
+		var currCdsPos int32
+		for _, val := range transcript.Exons {
+			for i = prevExonEnd - answer.startPos; i < val.Start-1-answer.startPos; i++ {
+				answer.featureArray[i] = Intron
```
craiglowe, line 35: Do we need to make sure that there are no overlapping genes in the gtf?  It seems like otherwise later gtf records overwrite earlier ones.

goGene/goGene.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+type Feature int32
+type MutationType byte
+
+const (
+	Intron   Feature = -1
+	UtrThree Feature = -3
+	UtrFive  Feature = -5
+	// All positive values refer to cDNA position
```
craiglowe, line 16: Just curious what happened to -2 and -4

goGene/goGene.go
```diff
@@ -0,0 +1,176 @@
...
+	Intronic     MutationType = 5
+	Splice       MutationType = 6 // +/- 1-2 of E-I boundary
+	FarSplice    MutationType = 7 // +/- 3-10 of E-I boundary
+	DisruptStart MutationType = 8
+	DisruptStop  MutationType = 9
+)
+
+// GoGene is a processed version of a gtf record that enables easy
+// traversal and manipulation of genes on the genomic and mRNA levels.
+type GoGene struct {
```
craiglowe, line 32: I think these fields might each need a sentence of documentation.  I am trying to imagine writing code to turn data from a file in genePred into a GoGene, and I don't think I would always know what to do.
What about Gene instead of GoGene?  We have been using the Go prefix to mean a function that will launch its own thread.  Or EasyGene, InternalGene?

goGene/fromGtf.go
```diff
@@ -0,0 +1,122 @@
...
+				answer.cdnaSeq = append(answer.cdnaSeq, answer.genomeSeq[answer.startPos-(val.Cds.End-1):answer.startPos-(val.Cds.Start-2)]...)
+
+				for i = answer.startPos - (val.Cds.End - 1); i < answer.startPos-(val.Cds.Start-2); i++ {
+					answer.featureArray[i] = Feature(currCdsPos)
+					currCdsPos++
+				}
+			}
+		}
+	}
+	answer.orig.startPos = answer.startPos
```
craiglowe, line 110: I am not sure why many fields seem to be stored twice.

goGene/goGene.go
```diff
@@ -0,0 +1,176 @@
...
+type EffectPrediction struct {
+	Consequence MutationType
+	CdnaPos     int // zero base
+	CdnaOffset  int
+	AaPos       int // zero base
+	AaRef       []dna.AminoAcid
+	AaAlt       []dna.AminoAcid
+}
+
+// GenomicToCdna converts genomic coordinates to cDNA coordinates. The return format is c.100+10.
```
craiglowe, line 69: I tried to google for the c.100+10 format, but I think it must be something obvious that I am missing.

goGene/goGene.go
```diff
@@ -0,0 +1,176 @@
...
+// The first int return is the nearest position in the coding sequence.
```
craiglowe, line 70: Is the position in the coding seq in genomic coordinates or cDNA coordinates?

goGene/goGene.go
```diff
@@ -0,0 +1,176 @@
...
+// The second int return is the distance from the nearest coding exon
+// (>0 if after cds; <0 if before cds, ==0 if inside coding sequence).
```
craiglowe, line 72: Are "before" and "after" in terms of genomic coordinates, or in terms of direction of transcription?  For example, if the gene is on the minus strand, would an intronic position with a greater position number than its closest exon be before or after that exon?  Might also be good to mention what happens with ties.

craiglowe (commented):
Few more comments

goGene/fromGtf.go
edotau: Not sure I'm following, couldnt all this information be scattered to different lines of a gtf file?

edotau: Actually i think i remember vaguely there was some functionality that logically handles this, is that what `gtf.Gene` is?

DanielSnellings: Yes. Gtf.Gene holds all the same information. The goal here is to make that information more accessible. If I gave you a gtf.Gene and asked what is the cDNA position for chr3:45877890 you would need to write a block of code to get that answer. Here you just need to do featureArray[45877890]. It seems like overkill at first. But it greatly simplifies complex tasks like predicting the effect of an insertion or deletion.

edotau (dismissed):
LGTM!

gene/gene.go
```diff
@@ -0,0 +1,178 @@
...
+type diff struct {
+	genomePos int
+	removed   []dna.Base
+	added     []dna.Base
+}
+
+type EffectPrediction struct {
+	Consequence MutationType
+	CdnaPos     int // zero base
+	CdnaOffset  int
```
craiglowe, line 65: I am not sure what CdnaOffset holds.  Maybe a quick comment here?

gene/gene.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+		}
+		return int(g.featureArray[queryPos+reverseOffset]), reverseOffset * -1, nil
+
+	case UtrFive:
+		var forwardOffset int = 1
+		for g.featureArray[queryPos+forwardOffset] < 0 {
+			forwardOffset++
+			if queryPos+forwardOffset > len(g.featureArray) {
+				return 0, 0, errors.New("no coding sequence found after 5'UTR")
```
craiglowe, line 119: How do noncoding RNAs fit into this?  It is ok if they don't; I am just curious.  In the GTF are they coded as all 5'UTR?  Is the Gene struct only for protein coding genes, or is there a way to have ncRNAs use a subset of the features?

gene/mutate.go
```diff
@@ -0,0 +1,478 @@
...
+	var log diff
+	var genomeIndexPos int
+	var err error
+
+	// Fill changeLog
+	log.genomePos = genomePos
+	log.added = make([]dna.Base, len(alt))
+	copy(log.added, alt)
+
+	for _, val := range alt {
```
craiglowe, line 103: This could be split out as a separate function (maybe in the dna package) that is something like:
func IsSeqOfACGT([]dna.Base) bool {

gene/mutate.go
```diff
@@ -0,0 +1,478 @@
...
+
+	if genomeIndexPos > len(g.genomeSeq)-1 {
+		return answer, errors.New("input genomePos is not in the gene")
+	}
+
+	// If no error, then append changeLog
+	g.changeLog = append(g.changeLog, log)
+
+	// Update genome sequence
+	g.genomeSeq = append(g.genomeSeq, alt...)                                     // just to increase cap, will get overwritten next line
```
craiglowe, line 132: Could we use dna.Insert here?

gene/mutate.go
```diff
@@ -0,0 +1,478 @@
...
+		//answer.CdnaPos, answer.CdnaOffset, err = GenomicPosToCdna(g, genomePos + 1)
+		//frame := (cdnaPos + 1) % 3
+		//var currCodon dna.Codon
+		//
+		//if frame != 0 { // if insertion disrupts a codon, note the codon being changed
+		//	currCodon, err = CdnaPosToCodon(g, cdnaPos)
+		//	answer.AaRef = []dna.AminoAcid{dna.TranslateCodon(&currCodon)}
+		//}
+
+		// Update cDNA
```
craiglowe, line 185: Maybe dna.Insert could be used here too?

gene/mutate.go
```diff
@@ -0,0 +1,478 @@
...
+	var deletedCdnaBases int
+	var cdnaDelStart, cdnaDelEnd int = -1, -1
+	deletionLen := genomeIndexEndPos - (genomeIndexStartPos + 1)
+	for i := 0; i < len(g.cdsStarts); i++ {
+
+		switch {
+		// if cds is before deletion
+		case genomeIndexStartPos >= g.cdsEnds[i]:
+			cdnaDelStart = int(g.featureArray[g.cdsEnds[i]])
+			continue
```
craiglowe, line 334: Is this continue needed/used?

craiglowe (commented):
Looks great!  Lots of amazing work here!  I am ready to approve.  I just added a few comments/thoughts.
In the future we could maybe break this into 2 or even 3 checkins?  It took me a few tries to get through it all.
Wonderful addition!  Maybe you can give a little overview of the functionality and how it can be used next code review?

numbers/numbers.go
```diff
@@ -196,3 +196,11 @@ func DigitsBaseTen(x int64) int {
 	}
 	return count
 }
+
+func Abs(x int) int {
```
rimangan, line 200: Perhaps AbsInt? math.Abs does the same already for floats. Also needs Abs(+ or - Inf) to equal +Inf as a special case and perhaps the ability to handle NaN.

gene/mutate.go
DanielSnellings: Yes, didn't know that existed.

gene/mutate.go
DanielSnellings: Changed

gene/mutate.go
DanielSnellings: Nope, nice catch

numbers/numbers.go
DanielSnellings: Changed it to AbsInt. Not sure how to implement the rest. Might need your help on that.

rimangan (approved):

cfauci (approved):

edotau (approved):
LGTM!

craiglowe (approved):
Great work!

PR #329, opened 2020-12-09
fastqFormat, fastqFilter UMI collapse, and fastq package pointer cleanup

LCBartelt:
skeleton for fastq reformat


cshoben (approved):

rimangan:
Here are the new features:
fastqFormat is a command that currently only supports single-cell formatting in the 10x style, takes in pairedEnd fastqs, parses out the UMI and barcodes, and adds these sequences as annotations to the 'name' field of the fastq, and then trims the UMI and barcode sequences away from the R1 fastq. For 10x scRNA-seq reads, this will produce a blank sequence for R1, but R1 will contain additional sequence for linked reads in the 10x style. Passes tests.
fastqFilter now has an additional option 'collapseUmi', in which removes reads from a fastq in 10x format that contains a barcode and UMI that has already been observed. Currently only removes exact matches for UMIs and barcodes, but we may consider expanding this to remove UMIs and barcodes that are off by 1 from something already observed. Passes tests.
Finally, I removed the []*Fastq style slices and replaced them with []Fastq, which led to many cascading changes to other packages. Everything covered by testing functions appears to work fine. 
Also added a Compare.go file for the Fastq package with documentation.

fastq/singleCell.go
```diff
@@ -0,0 +1,65 @@
...
+	Reads PairedEnd
+	Bx    []dna.Base
+	Umi   []dna.Base
+}
+
+//ReadToChanSingleCellPair takes two files, corresponding to R1 and R2, and returns a channel of SingleCellPair structs. It parses the barcode and UMI to user-specified lengths.
+func ReadToChanSingleCellPair(fileOne string, fileTwo string, barcodeLength int, umiLength int, tenXg chan<- SingleCellPair) {
+	fwd, rev := fileio.EasyOpen(fileOne), fileio.EasyOpen(fileTwo)
+	defer fwd.Close()
+	defer rev.Close()
```
craiglowe, line 20: I would think about closing these files before the "close(tenXg)".  Closing the channel signals to other go routines that this function is done, but the way it is written now it still needs to close two files after that channel is closed.

fastq/singleCell.go
```diff
@@ -0,0 +1,65 @@
...
+	sc := SingleCellPair{Reads: fqPair, Bx: GetBarcode(fqPair.Fwd, 0, barcodeLength), Umi: GetBarcode(fqPair.Fwd, barcodeLength, barcodeLength+umiLength)}
+	sc.Reads.Fwd = TrimFastq(fqPair.Fwd, barcodeLength+umiLength, len(fqPair.Fwd.Seq))
+	bxTag := fmt.Sprintf("BX:%s", dna.BasesToString(sc.Bx))
+	umiTag := fmt.Sprintf("UMI:%s", dna.BasesToString(sc.Umi))
+	sc.Reads.Fwd.Name = fmt.Sprintf("%s_%s_%s", fqPair.Fwd.Name, umiTag, bxTag)
+	sc.Reads.Rev.Name = fmt.Sprintf("%s_%s_%s", fqPair.Rev.Name, umiTag, bxTag)
+	return sc
+}
+
+//TrimFastq trims fastq bases and quals given a start and end position zero base
```
craiglowe, line 39: might be good to mention it is left-closed right-open

craiglowe (commented):
Fantastic work!  Could approve now, but made a few quick comments that might be improvements needing only a few line edits.

cmd/fastqFormat/fastqFormat.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+func fastqFormat(s Settings) {
+	if !s.PairedEnd {
+		log.Fatalf("fastqFormat is still under development. Currently, the only formatting options available are for paired end reads. Select 'pairedEnd' from options.")
+	} else if s.SingleCell {
+		scPairChan := make(chan fastq.SingleCellPair)
+		go fastq.ReadToChanSingleCellPair(s.R1InFile, s.R2InFile, s.BarcodeLength, s.UmiLength, scPairChan)
+		outR1 := fileio.EasyCreate(s.R1OutFile)
+		defer outR1.Close()
```
dasnellings, line 18: Just a reminder that deferring file closure is ok with os.File since the most serious error is "this file has already been closed", but the fileio writer has some serious errors that need to be handled.

fastq/compare.go
```diff
@@ -0,0 +1,135 @@
...
+		return false
+	}
+	if CompareQuals(alpha, beta) != 0 {
+		return false
+	}
+	return true
+}
+
+//CompareName compares two Fastq names for sorting or equality testing.
+func CompareName(alpha Fastq, beta Fastq) int {
```
dasnellings, line 51: It might be worth re-evaluating if all these compare functions need to be public functions.

fastq/compare.go
```diff
@@ -0,0 +1,135 @@
...
+	return strings.Compare(alpha.Name, beta.Name)
```
dasnellings, line 52: I think we should move away from using strings.Compare. From the godoc for strings.Compare: 

"Compare is included only for symmetry with package bytes. It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on."

dasnellings (approved):

craiglowe (approved):
Great work!

cfauci (approved):

PR #331, opened 2020-12-14
Sim speed up

cfauci:
Major speed up in Simulate at BaseExtToBases using sort.Slice. Go fmt caught bed.go and vcf.go in simulate package.


simulate/simulate.go
```diff
@@ -335,22 +336,36 @@ func BasesToBaseExt(seq []dna.Base) []BaseExt {
...
-	var i, j int
+	var newSequence []dna.Base = make([]dna.Base, len(seq))
 
-	for i = 0; i < len(seq); i++ { //counter to put things back in order
-		for j = 0; j < len(seq); j++ { //counter to check all seq bases before incrementing to look for the next base in the newSeq
-			if seq[j].SeqPos == i {
-				newSequence = append(newSequence, seq[j].Base)
-			}
-		}
+	SortBaseExtBySeqPos(seq)
```
DanielSnellings, line 341: No problem here, but for my own understanding why is a sort necessary?

simulate/simulate.go
```diff
@@ -335,22 +336,36 @@ func BasesToBaseExt(seq []dna.Base) []BaseExt {
...
 	if len(newSequence) != len(seq) {
 		log.Fatal("Cannot find order of bases")
 	}
 	return newSequence
 }
 
+//SortBaseExtBySeqPos orders a string of BaseExt by seq position
+func SortBaseExtBySeqPos(unordered []BaseExt) {
+	sort.Slice(unordered, func(i, j int) bool {
+		return Compare(unordered[i], unordered[j]) == 0
```
DanielSnellings, line 354: Should the check here be `Compare(...) < 0` instead of `Compare(...) == 0`?

simulate/simulate.go
cfauci: this was the only part of simulate that wasn't essentially instantaneous, sort is to put the bases back in order before appending and returning the mutating sequence.

simulate/simulate.go
cfauci: I'm doing a test to check this is actually functioning, but I'm not totally sure, I haven't really fully grasped how this function works yet.

cfauci:
Updated the things we spoke about in code review. Left the sorting code (which now passes tests on out of frame codons) in case of future use. Speed up went from 2 hours to complete simulate, it now takes 15 seconds and is linear. To run the 600 way tree on 100kb it takes about 10 min or so

rimangan (dismissed):

simulate/simulate.go
```diff
@@ -336,21 +336,34 @@ func BasesToBaseExt(seq []dna.Base) []BaseExt {
...
-			}
-		}
-	}
-	if len(newSequence) != len(seq) {
-		log.Fatal("Cannot find order of bases")
+//SortBaseExtBySeqPos orders a string of BaseExt by seq position
+func SortBaseExtBySeqPos(unordered []BaseExt) []dna.Base{
+	var ordered = make([]dna.Base, len(unordered))
+
+	for i := 0; i < len(unordered); i++ {
```
DanielSnellings, line 347: Not sure I'm following. It doesn't look like this is actually reordering anything. I thought the idea was to do something like `ordered[unordered[i].SeqPos] = unordered[i].Base`?

simulate/simulate.go
```diff
@@ -336,21 +336,34 @@ func BasesToBaseExt(seq []dna.Base) []BaseExt {
...
-		for j = 0; j < len(seq); j++ { //counter to check all seq bases before incrementing to look for the next base in the newSeq
-			if seq[j].SeqPos == i {
-				newSequence = append(newSequence, seq[j].Base)
-			}
-		}
-	}
-	if len(newSequence) != len(seq) {
-		log.Fatal("Cannot find order of bases")
+//SortBaseExtBySeqPos orders a string of BaseExt by seq position
+func SortBaseExtBySeqPos(unordered []BaseExt) []dna.Base{
```
craiglowe, line 344: Since this is not sorting the slice, but making a new slice, maybe a name like, BaseExtSliceToBaseSlice?

craiglowe (commented):
Looks good!  One typo/bug

edotau (approved):
LGTM!

craiglowe (approved):
Great work!

DanielSnellings (approved):

edotau (approved):

PR #332, opened 2020-12-14
fix for graphRecon error (argument missing in test call of simulate)

cfauci:
fix for the graphReconstruct error when testing all packages


edotau (approved):

craiglowe (approved):
Nice!  Good to get these tests running again.

DanielSnellings (approved):

PR #335, opened 2020-12-14
fixed compile errors for allele split

edotau:
TODO: make sure to test before using alleleSplit cmd

### DIFF
--- a/cmd/alleleSplit/toSam.go
+++ b/cmd/alleleSplit/toSam.go
@@ -9,10 +9,10 @@ import (
 )
 
 func SnpSearch(samfile string, genotypeVcf string, fOne string, parentOne string, parentTwo, prefix string) {
-	reader := vcf.GoReadGVcf(genotypeVcf)
-	sampleHash := vcf.HeaderToMaps(reader.Header)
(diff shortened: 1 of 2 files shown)


cfauci (approved):

cmd/alleleSplit/toSam.go
```diff
@@ -93,5 +93,13 @@ func SnpSearch(samfile string, genotypeVcf string, fOne string, parentOne string
 			sam.WriteAlnToFileHandle(childTwo, read)
 		}
 	}
-	reader.File.Close()
+}
+
+func getOldSeq(v *vcf.Vcf) [][]dna.Base {
```
craiglowe, line 98: Maybe getSeq?  I think the "Old" part makes sense to you because you are thinking about the previous version of the structs, but I think it may be confusing to other people.

craiglowe (approved):
Good to get this compiling again!  Does not have to be in this check-in, but would be good to make a point to go through your code and try to add a sentence of documentation to functions and structs and also add testing functions.

DanielSnellings (approved):

LCBartelt (approved):

rimangan (approved):

PR #337, opened 2020-12-16
Psl package coming soon to a coding session on zoom!

edotau:
new psl package, reading io utilities and a small test and benchmark


psl/compare.go
```diff
@@ -0,0 +1,81 @@
...
+		return false
+	}
+	if EqualSliceInt(x.TList, y.TList) {
+		return false
+	}
+	return true
+}
+
+// EqualSliceInt is a function what will check two slices of int if the values are the same.
+func EqualSliceInt(x []int, y []int) bool {
```
craiglowe, line 71: This will likely be useful beyond just dealing with PSLs.  Maybe the numbers package would make it easier to find?

psl/psl.go
```diff
@@ -0,0 +1,171 @@
...
+	Reader  *fileio.SimpleReader
+	curr    Psl
+	columns []string
+	err     bool
+}
+
+// NewPslReader will open a given text file and return a pointer to a PslReader struct.
+func NewPslReader(filename string) *PslReader {
+	var fields []string
+	var done bool
```
craiglowe, line 47: Are these variables outside the struct used?  Seems like they just hold default values, which are then assigned to the variables within the struct.  For example, could this line/variable be removed then then 'false' put for the initial value of err in the struct?

psl/psl.go
```diff
@@ -0,0 +1,171 @@
...
+	work := strings.Split(column, ",")
+	var answer []int = make([]int, len(work))
+	for i := 0; i < len(work)-1; i++ {
+		answer[i] = common.StringToInt(work[i])
+	}
+	return answer
+}
+
+// intListToString will process a slice of type int as an input and return a each value separated by a comma as a string.
+func intListToString(nums []int) string {
```
craiglowe, line 116: Maybe useful beyond PSLs?  Maybe the numbers package?

psl/psl.go
```diff
@@ -0,0 +1,171 @@
...
+		reader.curr.QList = stringToInts(reader.columns[19])
+		reader.curr.TList = stringToInts(reader.columns[20])
+		return &reader.curr, false
+	} else {
+		return nil, true
+	}
+}
+
+// stringToInts will process a column of bytes and convert the slice into a slice of type int.
+func stringToInts(column string) []int {
```
craiglowe, line 106: Likely useful beyond PSLs.  Maybe the numbers package or fileio?

psl/psl.go
```diff
@@ -0,0 +1,171 @@
...
+	for curr, err := pslLine(reader); !err; curr, err = pslLine(reader) {
+		ans = append(ans, *curr)
+	}
+	return ans
+}
+
+// pslLine is a helper function which assigns data fields and return a pointer to a Psl.
+func pslLine(reader *PslReader) (*Psl, bool) {
+	reader.Reader.Buffer, reader.err = fileio.ReadLine(reader.Reader)
+	if !reader.err {
```
craiglowe, line 69: Maybe this should be called something like "doneReading"? or "fileFinished"?  From the name it looks like the function silently returns nothing when there is an error, but from jumping through to fileio.ReadLine, it seems like this is a bool saying whether io.EOF was encountered and other errors cause fatal to be called.

psl/psl.go
```diff
@@ -0,0 +1,171 @@
...
+func Read(filename string) []Psl {
+	var ans []Psl
+	reader := NewPslReader(filename)
+	for curr, err := pslLine(reader); !err; curr, err = pslLine(reader) {
+		ans = append(ans, *curr)
+	}
+	return ans
+}
+
+// pslLine is a helper function which assigns data fields and return a pointer to a Psl.
```
craiglowe, line 66: Would be good to also mention the bool that it returns, which I think is a flag for when it has completed reading the file.

psl/psl.go
```diff
@@ -0,0 +1,171 @@
...
+	QList       []int
+	TList       []int
+}
+
+// PslReader is a struct that contains a SimpleReader with an embedded bufioReader and a 2-D byte slice to reduce memory allocation when reading each line.
+type PslReader struct {
+	Reader  *fileio.SimpleReader
+	curr    Psl
+	columns []string
+	err     bool
```
craiglowe, line 41: This is called error, but later in the file it seems to be mostly used as a flag for when the end of the file has been reached.

craiglowe (commented):
Great to see this functionality added!

psl/psl.go
edotau: You have a good point after thinking about this I dont think these variable have that much upside.

psl/psl.go
edotau: Yep I added that as a comment

fileio/simpleio.go
```diff
@@ -88,7 +89,32 @@ func CatchErrThrowEOF(err error) {
...
-		common.ExitIfError(err)
+		common.ExitIfError(reader.close())
 	}
 }
+
+// StringToIntSlice will process a column data separated by commas, convert the slice into a slice of type int.
+func StringToIntSlice(column string) []int {
+	work := strings.Split(column, ",")
+	var answer []int = make([]int, len(work))
+	for i := 0; i < len(work)-1; i++ {
```
DanielSnellings, line 100: Why does this ignore the last element in `work`? If this is to deal with data types with a trailing comma it might be good to make a note in the comment, otherwise someone may try to use it and have the end of their data cut off.

fileio/simpleio.go
```diff
@@ -88,7 +89,32 @@ func CatchErrThrowEOF(err error) {
...
+		answer[i] = common.StringToInt(work[i])
+	}
+	return answer
+}
+
+// IntListToString will process a slice of type int as an input and return a each value separated by a comma as a string.
+func IntSliceToString(nums []int) string {
+	ans := strings.Builder{}
```
DanielSnellings, line 108: Nice seeing strings.Builder!

DanielSnellings (approved):
One small comment but otherwise looks great!

fileio/simpleio.go
```diff
@@ -20,9 +20,9 @@ const (
 // and a pointer to os.File for closeure when reading is complete.
 type SimpleReader struct {
 	*bufio.Reader
-	file   *os.File
+	close  func() error
```
craiglowe, line 23: When using this struct, what is the difference between calling .Close(), like at the end of ReadLine below, or calling the .close() function on this line?

fileio/simpleio.go
craiglowe: Maybe we could check to see if the last character is a comma and then handle both cases?

fileio/simpleio.go
```diff
@@ -88,7 +89,32 @@ func CatchErrThrowEOF(err error) {
...
+func StringToIntSlice(column string) []int {
+	work := strings.Split(column, ",")
+	var answer []int = make([]int, len(work))
+	for i := 0; i < len(work)-1; i++ {
+		answer[i] = common.StringToInt(work[i])
+	}
+	return answer
+}
+
+// IntListToString will process a slice of type int as an input and return a each value separated by a comma as a string.
```
craiglowe, line 106: Might be good to say it adds a trailing comma at the end?

numbers/compare.go
```diff
@@ -0,0 +1,16 @@
+package numbers
+
+import ()
+
+// EqualSliceInt is a function what will check two slices of int if the values are the same.
```
craiglowe, line 5: Might be good to mention that order matters

craiglowe (commented):
Added a couple more comments after taking another pass.  Looks really good!

cfauci (approved):

LCBartelt (dismissed):

fileio/simpleio.go
rimangan: could someone explain a bit about the strings.Builder usage in code review tomorrow? I haven't seen this before

psl/psl.go
```diff
@@ -0,0 +1,155 @@
...
+package psl
+
+import (
+	"github.com/vertgenlab/gonomics/common"
+	"github.com/vertgenlab/gonomics/fileio"
+	"log"
+	"strings"
+)
+
+// Psl struct declares data fields and types for Psl struct.
```
rimangan, line 11: Sort of a side note, but Craig and I were discussing how LiftOver can use a PSL instead of a chain file for matching positions in the old and new assembly. I suppose this might be theoretically possible with Axt and MultiFa as well. perhaps we could consider some sort of "alignment" interface between these formats in the future

psl/psl.go
```diff
@@ -0,0 +1,155 @@
...
+			log.Fatalf("Error: number of columns do not equal 21...\n")
+			return nil, true
+		}
+	} else {
+		return nil, true
+	}
+}
+
+// ToString will print a Psl struct to a human readable string.
+func ToString(p *Psl) string {
```
rimangan, line 111: This is so much easier to read than the sprintf runon \t lines I'm used to seeing. Love this!

rimangan (approved):

fileio/simpleio.go
edotau: Yeah that was why I originally had it as a small private function in PSL, since it was specific to some of UCSC's file type anomalies. I have a better idea as a proposed change.

fileio/simpleio.go
edotau: You got it.

fileio/simpleio.go
edotau: I'm not too sure if there are differences other than small style preferences. I didn't want to keep *os.File around only to close the reader, so I defined a closure function to run the same behavior as os.File.close.
If i have `reader := NewSimpleReader(filename)`
I wanted to run `defer reader.Close()` or `reader.Close()` and close it at the end, without returning the error and mimic the close function in easyio reader.

rimangan (approved):

numbers/compare.go
edotau: Added that into the GoDocs

RavenYL (dismissed):

fileio/simpleio.go
edotau: It's basically like a bytes.Buffer but optimized for strings and the most desirable whenever you want to concatenate strings . In a slightly less complicated version, you can think about a string being a slice of bytes `[]byte`. When you want to append data you can to resize the `len([]byte)` and perform memory allocations. String builder optimizes the speed and performance of all this going on under the hood when you build simple strings.

cfauci (dismissed):
I'd also really like to hear more about strings.Builder in code review if the toString funcs can be made that way more often

cfauci (approved):

DanielSnellings (approved):

rimangan (approved):

LCBartelt (approved):

PR #338, opened 2020-12-18
http reading functions and small bug in ReadLine() when closing files

edotau:
Tried to close the reader inside the read line function, but that was creating more problems than was helpful.


cfauci (approved):

cmd/vimUrl/vimUrl.go
```diff
@@ -0,0 +1,38 @@
...
+		output := fileio.EasyCreate(*toFile)
+		defer output.Close()
+		reader := fileio.HttpReader(url)
+		for i, err := fileio.ReadLine(reader); !err; i, err = fileio.ReadLine(reader) {
+			i.WriteByte('\n')
+			output.Write(i.Bytes())
+		}
+	}
+
+	fileio.VimUrl(url)
```
craiglowe, line 37: I was expecting this to be in an "else", like you would either print it to a file, or print it to stdout.  I think this is setup that it always prints to the screen, but you can also have it print to a file, more like "tee" than a redirect.  If you want the behavior of "tee", maybe it makes sense to merge these two functions so that you only read the file once.  Right now I think it would read the file once and print it to the file, then read the file a second time and print it to the screen.

craiglowe (commented):
Cool functionality!  Is it possible to merge the functionality of easyio and simpleio?  I am wondering if simpleio is "better" and should be merged with easyio, or if it is "different" and we need to keep both.

rimangan (approved):

edotau:
came up with the name goFetchCat for the cmd exe. let me know if it should be something else. And i thought about @craig's comment about merging easy reader and simple reader. What do people think about the simple reader becoming the new easy ready? SimpleReader can basically do everything EasyReader does, but SimpleReader interfaces with the io.Reader library more elegantly.

craiglowe (approved):
Great work!

rimangan (approved):

cfauci (approved):

PR #344, opened 2021-01-05
dna.InsertStable

DanielSnellings:
This one is a bit technical. For my effect prediction code I want an insert function in the DNA package that maintains the underlying memory of the sequence slice. This push adds dna.InsertStable which does just that. 

The normal dna.Insert function takes the two input slices and outputs a brand new slice with the insertion (i.e. the input slides are not changed). The dna.InsertStable function actually inserts one slice into another without creating a new slice (i.e. the input slices are changed).

### DIFF
--- a/dna/modify.go
+++ b/dna/modify.go
@@ -130,6 +130,17 @@ func Delete(seq []Base, delStart int64, delEnd int64) []Base {
 	return append(seq[:delStart], seq[delEnd:]...)
 }
 
+// DeleteStable is identical to Delete, but conserves the
+// underlying memory of the input destSeq slice
+func DeleteStable(seq *[]Base, delStart int, delEnd int) error {
+	if delStart >= delEnd || delStart < 0 || delEnd > len(*seq) {
+		return ErrInvalidPosition
+	}
+	*seq = (*seq)[:len(*seq)-(delEnd-delStart)]                // decrease the LEN of the slice to by num of deleted bases
+	copy((*seq)[delStart:cap(*seq)], (*seq)[delEnd:cap(*seq)]) // shift bases in slice to remove deleted bases
+	return nil
+}
+
 // base position is zero-based, insertion happens before specified base
 // giving the length of the sequence puts the insertion at the end
 func Insert(seq []Base, insPos int64, insSeq []Base) []Base {
@@ -141,19 +152,18 @@ func Insert(seq []Base, insPos int64, insSeq []Base) []Base {
 
 // InsertStable is identical to Insert, but conserves the
 // underlying memory of the input destSeq slice if possible
-func InsertStable(destSeq []Base, insPos int, insSeq []Base) ([]Base, error) {
-	if insPos < 0 || insPos > len(destSeq) {
-		return nil, ErrInvalidPosition
+func InsertStable(destSeq *[]Base, insPos int, insSeq []Base) error {
+	if insPos < 0 || insPos > len(*destSeq) {
+		return ErrInvalidPosition
 	}
-	if len(destSeq)+len(insSeq) > cap(destSeq) {
-		return Insert(destSeq, int64(insPos), insSeq), ErrExceedCapNewSliceCreated
+	if len(*destSeq)+len(insSeq) > cap(*destSeq) {
+		*destSeq = Insert(*destSeq, int64(insPos), insSeq)
+		return ErrExceedCapNewSliceCreated
 	}
-	destSeq = destSeq[:len(destSeq)+1]                                           // extend the LEN of the slice to include the inserted bases
-	copy(destSeq[insPos+len(insSeq):cap(destSeq)], destSeq[insPos:cap(destSeq)]) // shift bases in slice by len(insSeq) at the position of the insertion
-	for i := 0; i < len(insSeq); i++ {                                           // add in the inserted bases
-		destSeq[insPos+i] = insSeq[i]
-	}
-	return destSeq, nil
+	*destSeq = (*destSeq)[:len(*destSeq)+len(insSeq)]                                    // extend the LEN of the slice to include the inserted bases
+	copy((*destSeq)[insPos+len(insSeq):cap(*destSeq)], (*destSeq)[insPos:cap(*destSeq)]) // shift bases in slice by len(insSeq) at the position of the insertion
+	copy((*destSeq)[insPos:insPos+len(insSeq)], insSeq)                                  // add in the inserted bases
+	return nil
 }
 
 // all base positions are zero based and left closed, right open
--- a/dna/modify.go
+++ b/dna/modify.go
@@ -1,9 +1,15 @@
 package dna
 
 import (
+	"errors"
 	"fmt"
 	"github.com/vertgenlab/gonomics/common"
 	"log"
+)
+
+var (
+	ErrInvalidPosition          = errors.New("Input position is invalid")
+	ErrExceedCapNewSliceCreated = errors.New("Capacity of input slice was exceeded, output slice was created at a new memory address")
 )
 
 func ToUpper(b Base) Base {
@@ -131,6 +137,23 @@ func Insert(seq []Base, insPos int64, insSeq []Base) []Base {
 		log.Fatalf("Error: insertion location of %d is not valid.\n", insPos)
 	}
 	return append(seq[:insPos], append(insSeq, seq[insPos:]...)...)
+}
+
+// InsertStable is identical to Insert, but conserves the
+// underlying memory of the input destSeq slice if possible
+func InsertStable(destSeq []Base, insPos int, insSeq []Base) ([]Base, error) {
+	if insPos < 0 || insPos > len(destSeq) {
+		return nil, ErrInvalidPosition
+	}
+	if len(destSeq)+len(insSeq) > cap(destSeq) {
+		return Insert(destSeq, int64(insPos), insSeq), ErrExceedCapNewSliceCreated
+	}
+	destSeq = destSeq[:len(destSeq)+1]
+	copy(destSeq[insPos+len(insSeq):cap(destSeq)], destSeq[insPos:cap(destSeq)])
+	for i := 0; i < len(insSeq); i++ {
+		destSeq[insPos+i] = insSeq[i]
+	}
+	return destSeq, nil
 }
 
 // all base positions are zero based and left closed, right open
--- a/dna/modify.go
+++ b/dna/modify.go
@@ -131,7 +131,7 @@ func Delete(seq []Base, delStart int64, delEnd int64) []Base {
 }
 
 // DeleteStable is identical to Delete, but conserves the
-// underlying memory of the input destSeq slice
+// underlying memory of the input destSeq slice.
 func DeleteStable(seq *[]Base, delStart int, delEnd int) error {
 	if delStart >= delEnd || delStart < 0 || delEnd > len(*seq) {
 		return ErrInvalidPosition
--- a/dna/modify.go
+++ b/dna/modify.go
@@ -148,9 +148,9 @@ func InsertStable(destSeq []Base, insPos int, insSeq []Base) ([]Base, error) {
 	if len(destSeq)+len(insSeq) > cap(destSeq) {
 		return Insert(destSeq, int64(insPos), insSeq), ErrExceedCapNewSliceCreated
 	}
-	destSeq = destSeq[:len(destSeq)+1]
-	copy(destSeq[insPos+len(insSeq):cap(destSeq)], destSeq[insPos:cap(destSeq)])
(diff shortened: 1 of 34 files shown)


dna/modify.go
```diff
@@ -133,6 +139,23 @@ func Insert(seq []Base, insPos int64, insSeq []Base) []Base {
...
+// underlying memory of the input destSeq slice if possible
+func InsertStable(destSeq []Base, insPos int, insSeq []Base) ([]Base, error) {
+	if insPos < 0 || insPos > len(destSeq) {
+		return nil, ErrInvalidPosition
+	}
+	if len(destSeq)+len(insSeq) > cap(destSeq) {
+		return Insert(destSeq, int64(insPos), insSeq), ErrExceedCapNewSliceCreated
+	}
+	destSeq = destSeq[:len(destSeq)+1]                                           // extend the LEN of the slice to include the inserted bases
+	copy(destSeq[insPos+len(insSeq):cap(destSeq)], destSeq[insPos:cap(destSeq)]) // shift bases in slice by len(insSeq) at the position of the insertion
```
craiglowe, line 152: I would consider using len instead of cap.
copy(destSeq[insPos+len(insSeq):len(destSeq)+len(insSeq)], destSeq[insPos:len(destSeq)])
I think the way you have it now you will be potentially copying a lot of values that are not in the length of destSeq, but are in the capacity.  For example, picture destSeq having a huge capacity (1 billion bases), but only using 1kb right now.  You will still be copying all the memory values between 1kb and 1billion.

dna/modify.go
```diff
@@ -133,6 +139,23 @@ func Insert(seq []Base, insPos int64, insSeq []Base) []Base {
...
+	for i := 0; i < len(insSeq); i++ {                                           // add in the inserted bases
+		destSeq[insPos+i] = insSeq[i]
```
craiglowe, line 154: Nothing wrong with this for-loop, but this loop could be replaced by another copy command:
copy(destSeq[insPos:insPos+len(insSeq)], insSeq)

craiglowe (commented):
Cool addition!  I had one or two comments.

dna/modify.go
```diff
@@ -133,6 +139,23 @@ func Insert(seq []Base, insPos int64, insSeq []Base) []Base {
 	return append(seq[:insPos], append(insSeq, seq[insPos:]...)...)
 }
 
+// InsertStable is identical to Insert, but conserves the
+// underlying memory of the input destSeq slice if possible
+func InsertStable(destSeq []Base, insPos int, insSeq []Base) ([]Base, error) {
```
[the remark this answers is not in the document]
DanielSnellings: I could also go either way. I definitely see the benefit (and prefer the style) of not returning a `[]Base`. Just to lay out my logic, the reason I built it to have a return was so that if the insertion exceeded the dest slice capacity then it would automatically generate a new slice and return an error. Do you think this tries to predict the users needs too much? I could just return the error and leave it to the user to generate a copy if they want. Interested in hearing your opinion.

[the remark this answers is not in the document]
craiglowe: I think your idea is a good one to allocate new memory for the user (if needed) and then use this new slice if needed.  I think you can do this seamlessly by passing in a pointer to the slice instead of a slice.  The trick is that at the end of the function when you would normally return the slice, you instead set the memory pointed to by the pointer to the values that reference the new slice.  This usually takes the form of:
*in = answer
I made a toy example you can see here:
https://play.golang.org/p/negQb7k_jkd

craiglowe (commented):
Added a response to a comment

dna/modify.go
```diff
@@ -1,11 +1,17 @@
...
 	"fmt"
 	"github.com/vertgenlab/gonomics/common"
 	"log"
 )
 
+var (
+	ErrInvalidPosition          = errors.New("Input position is invalid")
+	ErrExceedCapNewSliceCreated = errors.New("Capacity of input slice was exceeded, output slice was created at a new memory address")
+)
+
```
cshoben, line 14: I actually have a question about the code starting at line 15. For the if...else if...else statement; why bother if the b == Gap or b == Dot returning "b" when the else statement will also just return "b"? Wouldnt this code do exactly the same thing? :
if b > N {
     return b - 5
} else {
     return b
}

dna/modify.go
```diff
@@ -133,6 +150,22 @@ func Insert(seq []Base, insPos int64, insSeq []Base) []Base {
...
+func InsertStable(destSeq *[]Base, insPos int, insSeq []Base) error {
+	if insPos < 0 || insPos > len(*destSeq) {
+		return ErrInvalidPosition
+	}
+	if len(*destSeq)+len(insSeq) > cap(*destSeq) {
+		*destSeq = Insert(*destSeq, int64(insPos), insSeq)
+		return ErrExceedCapNewSliceCreated
+	}
+	*destSeq = (*destSeq)[:len(*destSeq)+len(insSeq)]                                    // extend the LEN of the slice to include the inserted bases
+	copy((*destSeq)[insPos+len(insSeq):cap(*destSeq)], (*destSeq)[insPos:cap(*destSeq)]) // shift bases in slice by len(insSeq) at the position of the insertion
```
cshoben, line 164: I'm trying to wrap my head around how copy() works. Can you point me in the direction of some documentation please?

dna/modify.go
DanielSnellings: In the definition of Base in dna/dna.go each base is a byte with a numerical value A=0, C=1... N=4... Gap=10, Dot=11. Since the numerical value assigned to Gap and Dot is greater than N, they would be caught by the first else statement and be changed to "a" and "c" respectively. To avoid this, they are handled in a separate if statement

dna/modify.go
DanielSnellings: See https://yourbasic.org/golang/copy-explained/

Basically copy just copies the content from one slice into another slice. Here i use copy to "shift" the elements in the slice to make room for more elements to be added. Its not a very intuitive use of the copy function, but it is very handy.

craiglowe, line 164: I think this is copying more than you need to.  You only need to copy the bases from the original seq that come after the insertion:
copy((*destSeq)[insPos+len(insSeq):len(*destSeq)], (*destSeq)[insPos:len(*destSeq)-len(insSeq)])

craiglowe (dismissed):
Great work!  I commented about some possible speed improvements for when len and cap are pretty different.

cfauci (dismissed):

dna/modify.go
```diff
@@ -124,6 +130,17 @@ func Delete(seq []Base, delStart int64, delEnd int64) []Base {
...
 }
 
+// DeleteStable is identical to Delete, but conserves the
+// underlying memory of the input destSeq slice
+func DeleteStable(seq *[]Base, delStart int, delEnd int) error {
+	if delStart >= delEnd || delStart < 0 || delEnd > len(*seq) {
+		return ErrInvalidPosition
+	}
+	*seq = (*seq)[:len(*seq)-(delEnd-delStart)]                // decrease the LEN of the slice to by num of deleted bases
+	copy((*seq)[delStart:cap(*seq)], (*seq)[delEnd:cap(*seq)]) // shift bases in slice to remove deleted bases
```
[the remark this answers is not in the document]
DanielSnellings: Consider the following:
`Cdna := make([]Base, 4, 5)`
`Cdna = dna.StringToBases("ACTG") // cDNA sequence is: ACTG`
`CodingSeq = Cdna[1:3] // CDS is: CT`

The goal of this function is the following
`InsertStable(CodingSeq, 1, []Base{A})`
Expected cDNA sequence is: ACATG
Expected CDS sequence is: CAT

If I only copy to the len and not the cap then the result would be
Actual cDNA sequence is: ACATA <- last base is nil value
Actual CDS sequence is: CAT

The start/end of the `Cdna` slice will need to be adjusted after the insertion, but the fidelity of the entire sequence array is maintained.

dna/modify.go
cshoben: So here we only want to return b-5 if b is greater than 4 (N) but is not equal to 10 (Gap) or 11 (Dot). And I assume you use "N", "Gap", and "Dot" instead of 4, 10, 11, since values are arbitrary and can be changed?

dna/modify.go
cshoben: That helped a lot, thanks Dan! I think I was getting caught up on the fact that the insert was beginning at the end point of the insertion but I realize it doesn't matter since the total number of slices will increase which is the ultimate goal.

cshoben: Craig, when I follow that code it looks like the number of slices is staying the same. Wouldnt subtracting the len(insSeq) defeat the purpose of increase by the length of the insertion in the destination argument?
assuming: copy (dst, src)
I made an example to help me think through these with destSeq [1:100] insPos = 50 and len(insSeq) = 10
With Dan's code I get: 
copy (60:100, 50:100) 
but with your code I'm getting:
copy(60:100, 50:90). 
Maybe I'm thinking about this all wrong as we can chat during code review?

dna/modify_test.go
```diff
@@ -70,6 +72,20 @@ func TestDel(t *testing.T) {
 	}
 }
 
+func TestDelStable(t *testing.T) {
+	destSeq := []Base{A, C, G, T}
+	origFirstBase := destSeq[0:1]
+	expectedSeq := []Base{A, G, T}
+	err := DeleteStable(&destSeq, 1, 2)
```
cshoben, line 79: Can you remind me why we would use "&destSeq" instead of just destSeq here? I understand you are trying to do things with memory address here but I'm not sure much past that.

cshoben: Maybe because this is an error and we don't want to actually change the original variable?

cfauci: DeleteStable takes a pointer to a slice of bases so as the first argument, so that has to be the format. destSeq is a slice of bases here and so DeleteStable is fed a pointer to it.

dna/modify.go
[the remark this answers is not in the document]
craiglowe: I think I understand the use case now.  Maybe this is specific to an insertion in your gene struct and does not smoothly generalize to any dna seq?
Does it work smoother if the insertion would be executed on the cDNA instead of the CDS?

dna/modify.go
craiglowe: I think the trick is that Dan is re-sizing the slice on the line above, so when the copy is executed, the length is no longer 100, but 110:
destSeq is originally a slice of length 100 (index 0 to index 99)
We are going to insert a 10bp, so now we change it to a slice of length 110, where the first 100 bases are the starting seq and the last 10 are random values
Then the copy command says take the values in destSeq[50 to (110-10)] aka destSeq[50 to 100] and put them in destSeq[50+10 to 110] aka destSeq[60 to 110]. This serves to "shift" all the bases in the original sequence that come after the insertion, to make room for the insertion.

craiglowe (commented):
Couple more comments

dna/modify.go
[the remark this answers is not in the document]
DanielSnellings: It does work smoothly if done on the cDNA and not the CDS. However, I think the use case for the IndelStable functions will be manipulating a slice where different variables hold a re-slice of the same underlying array. With multiple re-slices of an array the only way to safely make an indel would be to perform the function on an a slice that is a superset of all other slices (cDNA in this case). To maintain the ability to manipulate a subset slice I think we need to copy to the cap of the slice. I am not sure whether this generalizes to all []dna.Base. I think it depends on the use case. I would be happy to move these to private functions within the gene package if people think these are too specific.

RavenYL (dismissed):

dna/modify.go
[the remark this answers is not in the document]
craiglowe: Maybe this is worth a quick discussion at our 4pm meeting?
I am thinking that preserving data between len and cap is pretty specific to this use case and could be computationally expensive.  Maybe there is a way to call the insert function on cDNA?  It seems like even if the underlying array of cDNA was changed by calling insert on CDS, we would still need to modify the cDNA slice to understand that its length needed to be updated, since the last base of the cDNA is now either closer or farther from the start base.  Since we need access to that struct anyway, maybe there is a way to call this function on that one?
I don't want to hold up your progress though.  I think this is a relatively minor point.  Maybe just make a detailed note in InsertStable documentation that it copies all seq between len and cap as well?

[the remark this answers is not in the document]
cshoben: Thank you Craig! That makes much more sense. I was thinking about it wrong, totally forgot we already made the total number of slices longer in the code on the line directly above.

dna/modify_test.go
cshoben: Yay! This helped me so much Christi. And you read my mind about lines 80-81, I've been stuck there trying to wrap my head around it.

PR #348, opened 2021-01-07
Update checks

DanielSnellings:
Fixed the GOPATH problem in the checks. Unfortunately the only way to test it was to make a push so there are like 20 commits. Some things in main still dont build or pass tests so the checks still fail, but they are now working as intended. If you look at the logs in the 'Checks' panel, the errors should be the same as they are if you run them on your local copy of gonomics main.


craiglowe (approved):
Excited to have this!

cfauci (approved):

RavenYL (approved):

PR #350, opened 2021-01-09
Raven human insertion

RavenYL:
humanInsertion.go:
-takes a pairwise alignment maf
-finds lines that start with e and end with C belonging to species_del (e.g. panTro6)
-finds corresponding lines (the line before) that start with s belonging to species_ins (e.g. hg38)
-writes to 2 bed files, 1 for species_del and 1 for species_ins


cmd/humanInsertion/humanInsertion.go
```diff
@@ -0,0 +1,79 @@
...
+	var bedList_ins []*bed.Bed      //initialize 2 bed files
+	var bedList_del []*bed.Bed      //1 bed file for ins, 1 bed file for del
+
+	//go through each line
+	for i, _ := range mafRecords { //each i is a block
+		for k, _ := range mafRecords[i].Species { //each k is a line
+
+			//convert maf to bed, start with getting assembly because it is needed to verify species_ins and species_del
+			assembly_del, chrom_del := maf.SrcToAssemblyAndChrom(mafRecords[i].Species[k].Src)   //start with species_del, get assembly (e.g. panTro6), chrom (e.g. chrI)
+			assembly_ins, chrom_ins := maf.SrcToAssemblyAndChrom(mafRecords[i].Species[k-1].Src) //then find corresponding species_ins line, same block, 1 line above since pairwise maf
```
RavenYL, line 23: When I tried to run humanInsertion.go I got an index out of range error that seems to be related to this line and line 79. Does Species[k-1] not work? Why not?

go run humanInsertion/humanInsertion.go ../../LoweLab/Results/hg38.panTro6.synNet_head.maf ../../LoweLab/Results/outBed_hg38.bed ../../LoweLab/Results/outBed_panTro6.bed hg38 panTro6

panic: runtime error: index out of range

goroutine 1 [running]:
main.humanInsertion(0x7fffc967a18e, 0x32, 0x7fffc967a1c1, 0x25, 0x7fffc967a1e7, 0x28, 0x7fffc967a210, 0x4, 0x7fffc967a215, 0x7)
        /mnt/c/Users/raven/Desktop/gonomics/cmd/humanInsertion/humanInsertion.go:23 +0x973
main.main()
        /mnt/c/Users/raven/Desktop/gonomics/cmd/humanInsertion/humanInsertion.go:79 +0xf4
exit status 2

craiglowe: Yes, I think the k-1 is triggering the error.  Index out of range means you are trying to access a position in the slice (like arrays in many languages) that does not exist.  This can either be asking for a position with a negative index, or one beyond the last element of the slice.  You have two places where you use an index on that line and it could be the "i" or the "k-1".  I get used the line before and doesn't give an error, so k-1 looks like it could be it.
The for loop on line 19 sends k over all indices for the species in the maf block.  In this case we will generally have hg38 and panTro6 in the block, so k will first be 0 and then 1.  The first time through the loop k will be zero, or look at the first species in the block (probably hg38), so that makes k-1 equal to negative one, which is not allowed.
To verify this, I added the line:
log.Printf("score:%f k:%d k-1:%d\n", mafRecords[i].Score, k, k-1)
as line 20.
We can talk through this more at code review.  How you want to code this may depend on if you want to make it specific to pairwise mafs, where you could might be able to remove your 'k' loop and use hardcoded values of 0 and 1.  Otherwise I think you will have to use your k loop to find the index value of species you are looking for.

craiglowe (commented):
Quick comment about k-1 evaluating to -1

cmd/humanInsertion/humanInsertion.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	//write out bed files
+	bed.Write(outBed_del, bedList_del, 5) //bed file has 5 fields
+	bed.Write(outBed_ins, bedList_ins, 5)
+}
+
+func usage() {
+	fmt.Print(
+		"humanInsertion - takes pairwise alignment maf and finds insertions in species_ins not present in species_del but flanked by continuous alignments\n" +
```
craiglowe, line 55: Great job making this program more general than you need for your project.  i would also think about giving it a more general name since it can find insertions in all sorts of species (and also deletions).  Maybe "mafIndels"?

cmd/humanInsertion/humanInsertion.go
```diff
@@ -0,0 +1,81 @@
...
+			"Usage:\n" +
+			" humanInsertion mafFile outBed_ins outBed_del species_ins species_del\n" +
```
craiglowe, line 57: It is not always followed, but a general convention is to place inputs before outputs.  I would consider making outBed_ins and outBed_del at the end.  I would also consider names like "in.maf" and "outIns.bed" to make it more clear what is a file name and what is a direct input (e.g. species_ins).

maf/maf.go
```diff
@@ -140,6 +140,8 @@ func parseMafEStatus(s string) rune {
 		return 'M'
 	case "n":
 		return 'n'
+	case "T": //raven added the "T" case because maf lines can now start with e and end with T
```
craiglowe, line 143: Great!  Thank you for updating this!

cmd/humanInsertion/humanInsertion.go
```diff
@@ -0,0 +1,81 @@
...
+
+func humanInsertion(mafFile string, outBed_ins string, outBed_del string, species_ins string, species_del string) {
+	//initialize variables
+	mafRecords := maf.Read(mafFile) //Read entire mafFile. mafRecords has type Maf
+	var bedList_ins []*bed.Bed      //initialize 2 bed files
+	var bedList_del []*bed.Bed      //1 bed file for ins, 1 bed file for del
+
+	//go through each line
+	for i, _ := range mafRecords { //each i is a block
+		for k, _ := range mafRecords[i].Species { //each k is a line
```
craiglowe, line 19: We are looping k over all species in the alignment here, but later on there is a built in assumption that k=0 corresponds to  species_ins and k=1 corresponds to species_del.  Maybe it makes sense to start the k loop at 1 instead of 0 and then change the k-1 indices to 0?
The loop would be something like for k=1; k<len(mafRecords[i].Species); k++ {

craiglowe (commented):
Wonderful work!  I included a view minor comments.

cmd/humanInsertion/humanInsertion.go
```diff
@@ -0,0 +1,81 @@
...
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/maf"
+	"log"
+)
+
+func humanInsertion(mafFile string, outBed_ins string, outBed_del string, species_ins string, species_del string) {
+	//initialize variables
+	mafRecords := maf.Read(mafFile) //Read entire mafFile. mafRecords has type Maf
```
cshoben, line 13: I'm try to wrap my head around type Maf and was wondering if someone can tell me what the "S", "I", and "E" stand for in MafSLine, MafILine, and MafELine?

RavenYL: That's the letter that begins the line. In a maf file, within each alignment block, each species has at least 1 line describing its status in the alignment block. For example, if the line for human starts with s, that means the human sequence is aligned; if the line for chimp starts with e, that means the chimp sequence is "empty" or does not align
Official description of the Maf format: https://genome.ucsc.edu/FAQ/FAQformat.html#format5

cmd/humanInsertion/humanInsertion.go
```diff
@@ -0,0 +1,81 @@
...
+	var bedList_ins []*bed.Bed      //initialize 2 bed files
+	var bedList_del []*bed.Bed      //1 bed file for ins, 1 bed file for del
+
+	//go through each line
+	for i, _ := range mafRecords { //each i is a block
+		for k, _ := range mafRecords[i].Species { //each k is a line
+
+			//convert maf to bed, start with getting assembly because it is needed to verify species_ins and species_del
+			assembly_del, chrom_del := maf.SrcToAssemblyAndChrom(mafRecords[i].Species[k].Src) //start with species_del, get assembly (e.g. panTro6), chrom (e.g. chrI)
```
cshoben, line 22: I'm wondering why we use "_del" here? This variable is holding the assembly and the chrom before we even checked if it was a deletion or insertion, right? This took me a while to figure out because at first it seems like we somehow already knew we were dealing with a deletion based on the variable names, however my interpretation is that we don't really check for "Eline" until line 25. I understand that this variable continually is reset and only used if in fact that there is a deletion, but I still find it kinda confusing to also label insertions as "del" even if we junk them. Could we call it like assembly_k and chrom_k to just make it clear that it is the assembly and chrom being captured as we loop through k? Please help me if I'm thinking about this all wrong.

cshoben (changes requested):
I really appreciate all the explicit comments through out your code! I had a question about the k numbering that might need changed before the merge?

cmd/mafIndels/mafIndels.go
```diff
@@ -0,0 +1,84 @@
...
+	var bedList_ins []*bed.Bed     //initialize 2 bed files
+	var bedList_del []*bed.Bed     //1 bed file for ins, 1 bed file for del
+
+	//go through each line
+	for i, _ := range mafRecords { //each i is a block
+		for k := 1; k < len(mafRecords[i].Species); k++ { //each k is a line. Start loop at k=1, so that checking line k-1 starts at 0 and does not index out of range
+
+			//convert maf to bed, start with getting assembly because it is needed to verify species_ins and species_del
+			//here I assume only pairwise alignment, not >2 species
+			//here I assume species_ins is target (1st line in the block, k=0); species_del is query (2nd line in the block, k=1)
```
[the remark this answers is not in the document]
RavenYL: Yes you're right! Just changed this. Thank you!

cmd/mafIndels/mafIndels.go
```diff
@@ -0,0 +1,84 @@
...
+	//write out bed files
+	bed.Write(outDel_bed, bedList_del, 5) //bed file has 5 fields
+	bed.Write(outIns_bed, bedList_ins, 5)
+}
+
+func usage() {
+	fmt.Print(
+		"mafIndels - takes pairwise alignment maf and finds insertions in species_ins not present in species_del but flanked by continuous alignments\n" +
+			"here I assume only pairwise alignment, not >2 species\n" + //note my assumptions here
+			"here I assume species_ins is target (1st line in the block, k=0); species_del is query (2nd line in the block, k=1)\n" + //note my assumptions here
```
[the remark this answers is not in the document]
RavenYL: Good point! Just changed this

cshoben (approved):

DanielSnellings (approved):

rimangan (approved):

cfauci (approved):

craiglowe (approved):
Great work!

PR #351, opened 2021-01-09
changing axt from int64 to int

craiglowe:
Early on when we first started coding gonomics, I thought it would be a good idea to use int64 instead of int so that people running on small/cheaper computers could still index to the end of long chromosomes and deal with alignment scores that got very large.  In retrospect, I think this was a bad idea of mine: 1) 64 bit computers are not only the standard now, but it is actually hard to find a computer with a 32 bit processor these days.  I think that even small teaching universities, or even high schools, can find a 64 bit machine.  2) there are lots of times when it becomes a hassle when programming to have both ints and int64s bouncing around in a function.  I didn't realize how annoying this would be.
As a first attempt, I changed the axt struct to be int instead of int64.  I chose this because it is not used very much and would be a small change to the code base as a first attempt to discussing during code review.

### DIFF
--- a/axt/axt.go
+++ b/axt/axt.go
@@ -7,7 +7,6 @@ import (
 	"github.com/vertgenlab/gonomics/dna"
 	"github.com/vertgenlab/gonomics/fileio"
 	"log"
-	"strconv"
 	"strings"
 	"sync"
 )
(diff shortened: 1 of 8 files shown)


rimangan (commented):
Very happy with this change. Should we consider for bed?

DanielSnellings (approved):
Excited to see this shift happen!

cshoben (approved):

rimangan (approved):

RavenYL (approved):

PR #354, opened 2021-01-12
cleaning untracked testing files

rimangan:
Many test functions in the repository currently make tmp files for Read/Write tests that are never removed. Therefore, if you clone a clean copy of gonomics and run go test ./..., you wind up with a bunch of untracked files. This is a bit messy, so I went through all of these packages and added a fileio.EasyRemove() on tmp files. Everything should still pass tests.


RavenYL (dismissed):

fileio/easyio.go
```diff
@@ -60,6 +60,13 @@ func EasyNextRealLine(file *EasyReader) (string, bool) {
 	return NextRealLine(file.BuffReader)
 }
 
+func EasyRemove(filename string) {
+	err := os.Remove(filename)
+	if err != nil {
```
craiglowe, line 65: You don't need this "if" statement wrapped around common.ExitIfError(), inside the function it checks for nil and then dies if it is not nil (and otherwise the program carries on).

craiglowe (approved):
Looks great!  Thank you!  One very small comment on an extra "if" statement.

fileio/easyio.go
rimangan: great catch! fixed in next commit

DanielSnellings (approved):
Thanks! This is much nicer. Sorry for being one of the main offenders :-)

cfauci (approved):

craiglowe (approved):
Great!

PR #356, opened 2021-01-13
Gene Refactor 1

DanielSnellings:
I have just finished a significant refactor of the `gene` package to add in several new fields to the struct handling coding and UTR sequences. I have broken up the pull request to make each one easier to review.


rimangan (dismissed):

gene/fromGtf.go
```diff
@@ -40,11 +40,9 @@ func GtfToGene(g *gtf.Gene, ref []*fasta.Fasta) *Gene {
 				for i = val.FiveUtr.Start - 1 - answer.startPos; i < val.FiveUtr.End-answer.startPos; i++ {
 					answer.featureArray[i] = UtrFive
 				}
-			}
-
-			if val.ThreeUtr != nil {
-				for i = val.ThreeUtr.Start - 1 - answer.startPos; i < val.ThreeUtr.End-answer.startPos; i++ {
-					answer.featureArray[i] = UtrThree
+				answer.cdnaSeq = append(answer.cdnaSeq, fastaMap[transcript.Chr][val.FiveUtr.Start-1:val.FiveUtr.End]...)
+				if val.FiveUtr.End-answer.startPos > answer.utrFive.end {
```
craiglowe, line 44: I am not totally sure what is happening here.  Not saying it is a bug, just not sure I completely follow it.  I am thinking we are changing coordinate systems where "val" uses zero as the first position in the chromosome and answer uses zero as the first position of transcription.  I think a 1-line comment might help me.

gene/fromGtf.go
```diff
@@ -58,6 +56,16 @@ func GtfToGene(g *gtf.Gene, ref []*fasta.Fasta) *Gene {
 					currCdsPos++
 				}
 			}
+
+			if val.ThreeUtr != nil {
+				for i = val.ThreeUtr.Start - 1 - answer.startPos; i < val.ThreeUtr.End-answer.startPos; i++ {
+					answer.featureArray[i] = UtrThree
+				}
+				if answer.utrThree.start == 0 {
```
craiglowe, line 64: Again, not a bug, I think I am must having trouble telling what is happening.  I am not why/how a 3' UTR would have a start position of zero.  I guess there would be no 5' UTR and no coding seq.  Is this catching ncRNA genes, which are encoded as all 3' UTR?  I am also not sure why in the "if" we then set the 3' UTR to start the base after the cDNA ends; it seems like that would give an index-out-of-range when used.

gene/fromGtf.go
```diff
@@ -67,7 +75,7 @@ func GtfToGene(g *gtf.Gene, ref []*fasta.Fasta) *Gene {
 		answer.genomeSeq = make([]dna.Base, transcript.End-(transcript.Start-1))
 		copy(answer.genomeSeq, fastaMap[transcript.Chr][transcript.Start-1:transcript.End])
 		dna.ReverseComplement(answer.genomeSeq)
-		answer.cdnaSeq = make([]dna.Base, 0, gtf.CdsLength(transcript))
+		answer.cdnaSeq = make([]dna.Base, 0, len(answer.genomeSeq))
```
craiglowe, line 78: Not a bug, but it seems like we are allocating more memory than needed.  Seems like this would be the max amount of memory needed (the case where there are no introns).

gene/fromGtf.go
```diff
@@ -87,11 +95,9 @@ func GtfToGene(g *gtf.Gene, ref []*fasta.Fasta) *Gene {
 				for i = answer.startPos - (val.FiveUtr.End - 1); i < answer.startPos-(val.FiveUtr.Start-2); i++ {
 					answer.featureArray[i] = UtrFive
 				}
-			}
-
-			if val.ThreeUtr != nil {
-				for i = answer.startPos - (val.ThreeUtr.End - 1); i < answer.startPos-(val.ThreeUtr.Start-2); i++ {
-					answer.featureArray[i] = UtrThree
+				answer.cdnaSeq = append(answer.cdnaSeq, answer.genomeSeq[answer.startPos-(val.FiveUtr.End-1):answer.startPos-(val.FiveUtr.Start-2)]...)
+				if answer.utrFive.end < answer.startPos-(val.FiveUtr.Start-2) {
```
craiglowe, line 99: Similar to above, I am having trouble with the "if"s dealing with UTRs.  Not saying it is a bug, just that I am having trouble following everything.  Maybe either more comments, or shorter functions (where the function names then act as comments for what is happening)?

gene/fromGtf.go
```diff
@@ -105,8 +111,27 @@ func GtfToGene(g *gtf.Gene, ref []*fasta.Fasta) *Gene {
 					currCdsPos++
 				}
 			}
+
+			if val.ThreeUtr != nil {
+				for i = answer.startPos - (val.ThreeUtr.End - 1); i < answer.startPos-(val.ThreeUtr.Start-2); i++ {
+					answer.featureArray[i] = UtrThree
+				}
+				if answer.utrThree.start == 0 {
+					answer.utrThree.start = len(answer.cdnaSeq)
```
craiglowe, line 120: Again, this seems like it could cause index-out-of-range problems; however, I might not be seeing the big picture.

gene/gene.go
```diff
@@ -31,16 +31,27 @@ const (
...
-	cdsStarts    []int        // The start position of each CDS. This value is stored in Gene genomic coordinates (slice index of genomeSeq)
-	cdsEnds      []int        // The end position of each CDS. This value is stored in Gene genomic coordinates (slice index of genomeSeq)
-	genomeSeq    []dna.Base   // The genomic sequence of the gene from 5' to 3'. NOTE: This field is reverse complemented relative to reference file if the gene is on the negative strand
-	cdnaSeq      []dna.Base   // The cDNA sequence of the gene from 5' to 3'.
-	featureArray []Feature    // FeatureArray is a slice with len(featureArray) = len(genomeSeq). The index of featureArray corresponds to the same index of genomeSeq. featureArray denotes the features listed above as negative values, or the cDNA pos using all values > 0.
-	orig         goGeneBackup // Copy of initial Gene state to enable the Reset() function.
-	changeLog    []diff       // Log of any mutations that have been performed on the Gene to enable the Reset() function.
+	id           string          // Identifier for the transcript the Gene is derived from. In GTF this is the GeneID field.
```
craiglowe, line 34: Is it worth having a "symbol" field, where we could store the "human-readable" gene name (e.g. RUNX2)?

gene/gene.go
```diff
@@ -134,7 +156,7 @@ func CdnaPosToGenomic(g *Gene, cdnaPos int) (int, error) {
 	if cdnaPos < 0 {
 		return 0, errors.New("input cDNA position must be positive")
 	}
-	if cdnaPos > len(g.cdnaSeq)-1 {
+	if cdnaPos > len(g.codingSeq.seq)-1 {
```
craiglowe, line 159: I am not sure why this changed from cdna to codingSeq.  From the error message, cdna seems like what I would expect.

gene/gene.go
```diff
@@ -156,17 +178,17 @@ func CdnaPosToCodon(g *Gene, cdnaPos int) (dna.Codon, error) {
 	if cdnaPos < 0 {
 		return answer, errors.New("input cDNA position must be positive")
 	}
-	if cdnaPos > len(g.cdnaSeq)-1 {
+	if cdnaPos > len(g.codingSeq.seq)-1 {
```
craiglowe, line 181: I am not sure why this changed from cdna to codingSeq.  From the error message, cdna seems like what I would expect.

gene/gene.go
```diff
@@ -156,17 +178,17 @@ func CdnaPosToCodon(g *Gene, cdnaPos int) (dna.Codon, error) {
...
 		return answer, errors.New("input position is greater than the length of the cDNA")
 	}
 
 	switch cdnaPos % 3 {
 	case 0:
-		return dna.Codon{Seq: g.cdnaSeq[cdnaPos : cdnaPos+3]}, nil
+		return dna.Codon{Seq: g.codingSeq.seq[cdnaPos : cdnaPos+3]}, nil
```
craiglowe, line 187: This relates to last two comments.  cdna seems to make sure sense than codingSeq.  Maybe I am not understanding what the codingSeq is?

craiglowe (commented):
Great work!  Didn't see any bugs, but there were places that I had trouble fully understanding.  I think some combo of shorter functions or more comments might help me.

gene/compare.go
```diff
@@ -0,0 +1,180 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"github.com/vertgenlab/gonomics/dna"
+)
+
+func equal(alpha, beta *Gene) (bool, error) {
+	if alpha.id != beta.id {
+		return false, errors.New("id")
```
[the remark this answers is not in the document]
DanielSnellings: This was really just intended as a testing function that will return the difference. Not sure what the utility of an equal function would be when actually using gene. In most cases I imagine it would be more useful just to compare the part you were interested in.

gene/fromGtf.go
```diff
@@ -22,7 +22,7 @@ func GtfToGene(g *gtf.Gene, ref []*fasta.Fasta) *Gene {
 		fastaMap := fasta.FastaMap(ref)
 		answer.genomeSeq = make([]dna.Base, transcript.End-(transcript.Start-1))
 		copy(answer.genomeSeq, fastaMap[transcript.Chr][transcript.Start-1:transcript.End])
-		answer.cdnaSeq = make([]dna.Base, 0, gtf.CdsLength(transcript))
+		answer.cdnaSeq = make([]dna.Base, 0, len(answer.genomeSeq))
```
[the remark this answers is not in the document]
DanielSnellings: Correct, this gives the cdna plenty of room to make insertions which, in theory, should reduce the number of full array copies that need to be done.

gene/fromGtf.go
DanielSnellings: Correct, the UTR sequences are being recorded relative to the cDNA such that the first base of the 5' UTR is base zero. I added a comment.

gene/fromGtf.go
DanielSnellings: So this relies on the fact that GTF is organized such that the 5' UTR comes before the CDS comes before the 3' UTR. By the time a 3' UTR line is being read the 5' UTR and CDS sequences have already been recorded. This if statement says "if the 3' UTR start is unset (==0), then this must be the first 3'UTR sequence we encountered, therefore the start will be equal to wherever we currently are in the cDNA encoding." I added a couple comments to hopefully help clear this up.

gene/fromGtf.go
DanielSnellings: Correct, this was just my quick & dirty way of adding space for indels.

gene/fromGtf.go
DanielSnellings: Everything here is working to get them relative to the cDNA. I will see what I can do to break up the functions.

gene/fromGtf.go
DanielSnellings: Same as above, this is relying on the fact that the cDNA is read in order.

gene/gene.go
DanielSnellings: added

gene/gene.go
DanielSnellings: In the previous push I was calling this cDNA when what I really ment was CDS. This change is correct, but I changed the rest of the function to properly refer to this as `CodingPosToGenomic`.

gene/gene.go
DanielSnellings: In the previous push I was calling this cDNA when what I really ment was CDS. This change is correct, but I changed the rest of the function to properly refer to this as CodingPosToGenomic.

gene/gene.go
DanielSnellings: Before I was using cDNA to refer to the actual coding sequence of the gene, but of course the cDNA also includes the UTR sequences, so I made several changes to make cdna and codingSeq different.

DanielSnellings:
I did some much needed simplification of fromGtf. It should be a bit easier to follow now.

cfauci (approved):

craiglowe (approved):
Wonderful work!

rimangan (approved):

PR #357, opened 2021-01-13
Gene Refactor 2

DanielSnellings:
See PR Gene Refactor 1 before reviewing

Adds point mutations

New Files:
`pointMutation.go`
`pointMutation_test.go`
`reset.go`

### DIFF
--- a/gene/pointMutation.go
+++ b/gene/pointMutation.go
@@ -1,0 +1,129 @@
+package gene
+
+import (
+	"github.com/vertgenlab/gonomics/dna"
+	"github.com/vertgenlab/gonomics/numbers"
+)
+
+// PointMutation changes a single nucleotide to the desired base, predicts the effect,
+// and updates the Gene struct to reflect the change.
+// The position of the mutation should be given in base-zero genomic coordinates.
+func PointMutation(g *Gene, genomePos int, alt dna.Base) (EffectPrediction, error) {
+	var answer EffectPrediction
+	answer.StopDist = -1
+	var err error
+
+	err = pointPreRunChecks(g, genomePos, &alt)
+	if err != nil {
+		return EffectPrediction{}, err
+	}
+
+	genomeIndexPos := numbers.AbsInt(genomePos - g.startPos) // abs needed to handle negative posStrand
+	g.genomeSeq[genomeIndexPos] = alt
+
+	cdnaIndexPos := int(g.featureArray[genomeIndexPos])
+
+	if cdnaIndexPos >= 0 { // mutation is coding
+		err = pointCoding(g, cdnaIndexPos, alt, &answer)
+	} else { // mutation is noncoding
+		err = pointNonCoding(g, genomePos, &answer)
+	}
+
+	if err != nil {
+		return EffectPrediction{}, err
+	}
+
+	g.protSeq = dna.TranslateSeq(g.codingSeq.seq) // TODO improve efficiency by updating the protein as changes are made
+	return answer, nil
+}
+
+// checkSplice inputs a cDNA offset value and determines if a variant at this site may affect splicing.
+func checkSplice(CdnaOffset int) MutationType {
+	if numbers.AbsInt(CdnaOffset) <= 2 {
+		return Splice
+	}
+
+	if numbers.AbsInt(CdnaOffset) <= 10 {
+		return FarSplice
+	}
+
+	return Intronic
+}
+
+// pointPreRunChecks performs error checking and appends the diff log prior to making a point mutation.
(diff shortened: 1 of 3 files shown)


rimangan (approved):

cfauci (approved):

RavenYL (approved):

gene/pointMutation.go
```diff
@@ -0,0 +1,129 @@
...
+	var answer EffectPrediction
+	answer.StopDist = -1
+	var err error
+
+	err = pointPreRunChecks(g, genomePos, &alt)
+	if err != nil {
+		return EffectPrediction{}, err
+	}
+
+	genomeIndexPos := numbers.AbsInt(genomePos - g.startPos) // abs needed to handle negative posStrand
```
craiglowe, line 21: I am having trouble following this conversion.  I think we are changing an index where 0 is the first base of the chrom, to the corresponding index where 0 is the first base of the cDNA.  Now that I am typing this, I think what is throwing me off is that you are assuming that genomePos is within the gene, but I think I would feel more comfortable if we check that.  Correctly if they give a genomePos on the "wrong side" of the cDNA start, it will just fix it for them and mutate the position the same amount to the "right side" of the cDNA start.  While this works perfectly when the input is reasonable, it makes me nervous that unreasonable input is converted to something reasonable without an error.

gene/pointMutation.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return EffectPrediction{}, err
+	}
+
+	g.protSeq = dna.TranslateSeq(g.codingSeq.seq) // TODO improve efficiency by updating the protein as changes are made
+	return answer, nil
+}
+
+// checkSplice inputs a cDNA offset value and determines if a variant at this site may affect splicing.
+func checkSplice(CdnaOffset int) MutationType {
```
craiglowe, line 41: I am not totally sure what CdnaOffset is.  From the code in the function it seems like it is the position of a mutation minus the position of a splice site.  Maybe it should be called "distanceToSpliceSite"?  Maybe it should take the mutation index and the splice site index as two arguments instead of one?

gene/pointMutation.go
```diff
@@ -0,0 +1,129 @@
...
+	if numbers.AbsInt(CdnaOffset) <= 2 {
+		return Splice
+	}
+
+	if numbers.AbsInt(CdnaOffset) <= 10 {
+		return FarSplice
+	}
+
+	return Intronic
```
craiglowe, line 50: It seems like all we know is that it is not close to a splice site.  Maybe it could be coding or in a utr?  Not sure why it would need to be an intron.

gene/pointMutation.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// pointPreRunChecks performs error checking and appends the diff log prior to making a point mutation.
+func pointPreRunChecks(g *Gene, genomePos int, alt *dna.Base) error {
+	var log diff
+
+	genomeIndexPos := numbers.AbsInt(genomePos - g.startPos) // abs needed to handle negative posStrand
```
craiglowe, line 57: Same comment as above for this index conversion.  Maybe it makes sense to make this a function that can do some checking and also the conversion?

gene/pointMutation.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if !dna.IsSeqOfACTG([]dna.Base{*alt}) {
+		return ErrNonACGTBase
+	}
+
+	if genomePos < 0 {
+		return ErrNegativeInputValue
+	}
+	if g.posStrand {
```
craiglowe, line 77: I am seeing the checks here that I was expecting at the conversion step.  This addresses my previous comments.  I do wonder if making this coordinate conversion its own function (since I have already seen it called a couple times) that included the checks would be more readable.

gene/pointMutation.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// if no error then  append log
+	g.changeLog = append(g.changeLog, log)
+	return nil
+}
+
+// pointNonCoding determines the effects of a non-coding point mutation
+func pointNonCoding(g *Gene, genomePos int, answer *EffectPrediction) error {
+	var err error
+	answer.CdnaPos, answer.CdnaDist, err = GenomicPosToCdna(g, genomePos)
```
craiglowe, line 100: Maybe GenomicPosToCdna is the function I was thinking about in my last comment?  Why not use it in the calls above like you use it here?

craiglowe (approved):
Great work!  I had some comments about coordinate conversions (checks and doing them in a function vs doing them in line with some math).

PR #358, opened 2021-01-13
Gene Refactor 3

DanielSnellings:
See PR Gene Refactor 2 before reviewing

Adds deletions

New Files:
`delete.go`
`delete_test.go`


rimangan (dismissed):

RavenYL (dismissed):

dna/codon.go
```diff
@@ -297,6 +297,20 @@ func BasesToCodons(b []Base) []Codon {
 	return answer
 }
 
+// BasesToCodonsIgnoreRemainder converts a slice of DNA bases into a slice of Codons.
+// Any bases remaining after all 3-base codons have been assembled will be ignored.
+func BasesToCodonsIgnoreRemainder(b []Base) []Codon {
+	frame := len(b) % 3
```
craiglowe, line 303: After this line, could we just:
return BasesToCodons(b[0:len(b)-frame])

dna/codon.go
```diff
@@ -342,6 +356,22 @@ func TranslateSeq(b []Base) []AminoAcid {
...
+// The translation will end after the first stop codon is reached and the function will return
+// the protein sequence including the trailing stop codon. Any bases beyond the stop codon,
+// or remaining bases after all 3-base codons have been made will be ignored.
+func TranslateSeqToTer(b []Base) []AminoAcid {
+	answer := make([]AminoAcid, 0, len(b)/3)
+	codons := BasesToCodonsIgnoreRemainder(b)
+	for i := range codons {
+		answer = append(answer, TranslateCodon(codons[i]))
+		if answer[i] == Stop {
+			break
```
craiglowe, line 369: Is it more clear to just return right here?

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+	"github.com/vertgenlab/gonomics/dna"
+)
+
+var (
+	ErrNoStopFound = errors.New("frameshift was unable to find a stop before the end of the gene")
+)
+
+//TODO EffectPrediction
+// Deletion removes bases from the Gene, predicts the effect, and updates the Gene struct to reflect the change.
+// genomeStartPos should be the base directly BEFORE the deleted bases. genomeEndPos should be the base directly AFTER the deleted bases.
```
craiglowe, line 14: Is it easier if we keep the deletion bounds to our usual left-close right-open, instead of left-open right-open?

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+	origCdsStart := g.cdsStarts[0]
+	origCdsEnd := g.cdsEnds[len(g.cdsEnds)-1]
+
+	// Delete from Genome Sequence
+	copy(g.genomeSeq[genomeIndexStartPos+1:], g.genomeSeq[genomeIndexEndPos:])
+	g.genomeSeq = g.genomeSeq[:len(g.genomeSeq)-(genomeIndexEndPos-(genomeIndexStartPos+1))]
+
+	// Update startPos if necessary
+	if genomeIndexStartPos == -1 {
+		g.startPos += genomeIndexEndPos - (genomeIndexStartPos + 1)
```
craiglowe, line 35: I am struggling with this if statement.  I am thinking that genomeIndexStartPos is the genomic start of the deletion mapped into a coordinate system where the TSS is 0 and the second transcripted base (in whatever direction) is 1.  So now we have an index of -1, which would be the base just before the TSS.  Because we are left-open, we are starting the deletion by deleting the TSS, and we do:
g.startPos += genomeIndexEndPos - (-1 + 1)
evaluate to:
g.startPos += genomeIndexEndPos
Am I thinking about this right?  Is it more clear without the extra math that evals to zero?

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+	var cdsIndexesToDelete []int
+	var deletedCodingBases int
+	var codingDelStart, codingDelEnd int = -1, -1
+	deletionLen := genomeIndexEndPos - (genomeIndexStartPos + 1)
+	for i := 0; i < len(g.cdsStarts); i++ {
+
+		switch {
+		// if cds is before deletion
+		case genomeIndexStartPos >= g.cdsEnds[i]:
+			codingDelStart = int(g.featureArray[g.cdsEnds[i]])
```
craiglowe, line 48: I am not sure what codingDelStart is saving/recording here.  I am thinking that g.cdsEnds[i] is the base after the coding exon being looked at, and then featureArray is going to tell me the function/classification of that base position.  Can a gene end with a coding exon (no 3' UTR)?  Would this give an index out of range.  I am not sure why we are saving either intron or 3' UTR here.

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+
+		// If cds exon is fully deleted
+		case genomeIndexStartPos < g.cdsStarts[i] && genomeIndexEndPos > g.cdsEnds[i]:
```
craiglowe, line 51: Should the > be >=?  I am thinking that you need the < because the deletion is left-open and the cdsStart is left-closed?

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+			// mark cds for deletion
+			cdsIndexesToDelete = append(cdsIndexesToDelete, i)
+			deletedCodingBases += (g.cdsEnds[i] + 1) - g.cdsStarts[i]
+			if codingDelStart == -1 {
```
craiglowe, line 55: I am not sure what this variable holds

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+				codingDelStart = int(g.featureArray[g.cdsStarts[i]]) - 1
+			}
+			codingDelEnd = int(g.featureArray[g.cdsEnds[i]]) + 1
+
+		// If deletion falls within a cds
+		case genomeIndexStartPos >= g.cdsStarts[i] && genomeIndexStartPos < g.cdsEnds[i] &&
+			genomeIndexEndPos > g.cdsStarts[i] && genomeIndexEndPos <= g.cdsEnds[i]:
```
craiglowe, line 62: I think the <= should be < so that it falls within?

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+			// mark cds for deletion
+			cdsIndexesToDelete = append(cdsIndexesToDelete, i)
+			deletedCodingBases += (g.cdsEnds[i] + 1) - g.cdsStarts[i]
+			if codingDelStart == -1 {
+				codingDelStart = int(g.featureArray[g.cdsStarts[i]]) - 1
+			}
+			codingDelEnd = int(g.featureArray[g.cdsEnds[i]]) + 1
+
+		// If deletion falls within a cds
+		case genomeIndexStartPos >= g.cdsStarts[i] && genomeIndexStartPos < g.cdsEnds[i] &&
```
craiglowe, line 61: Because the deletion interval is left-open, I am thinking that "genomeIndexStartPos <" should have a "+1" in there?

gene/delete.go
```diff
@@ -0,0 +1,338 @@
...
+			genomeIndexEndPos > g.cdsStarts[i] && genomeIndexEndPos <= g.cdsEnds[i]:
+			g.cdsEnds[i] -= deletionLen
+			deletedCodingBases += deletionLen
+			codingDelStart = int(g.featureArray[genomeIndexStartPos])
+			codingDelEnd = int(g.featureArray[genomeIndexEndPos])
+
+		// If cds exon is partially deleted on right
+		case genomeIndexStartPos >= g.cdsStarts[i] && genomeIndexStartPos < g.cdsEnds[i]:
```
craiglowe, line 69: I am again thinking that there should be a "+1" before the "<".  If g.cdsEnds[i] is 10, that means that position 9 is the last base of the exon (right-open).  However, if genomeIndexStartPos is 9, that means that 10 is the first base deleted (left-open).  So genomeIndexStartPos=9 and g.cdsEnds[i]=10 satisfies the "<", but that deletion will not touch the exon.

craiglowe (commented):
Really cool functionality, but I was not able to keep up with the coordinate system.  I kept thinking I was finding off-by-one errors.  Maybe a few more comments would help me.

gene/delete.go
DanielSnellings: Yeah this was a holdover from the days when this was intended to take VCFs. I've changed it to left closed right open. This greatly reduced the number of +/-1 shifts so hopefully the coordinates are a bit easier to follow now.

gene/delete.go
DanielSnellings: You are correct, that simplifies things since in this case we already know that genomeIndexStartPos is zero (since updating to left closed).

gene/delete.go
DanielSnellings: Yeah... to confuse things even more the cdsStart/cdsEnds are left closed right closed so this featureArray lookup actually gives the cds position. I figured that a fully closed cdsStart/End would reduce the number of shifts needed, but maybe that should be changed too. Thoughs?

gene/delete.go
DanielSnellings: since the index positions were fully open and the cds positions were fully closed this logic was correct, however it has been changed since making the index positions left closed right open

gene/delete.go
DanielSnellings: the position of deletion start in the coding sequencing (as opposed to genomeIndexStartPos which is the genomic coordinates)

gene/delete.go
DanielSnellings: cds ends is closed, sorry for the confusion.

gene/delete.go
DanielSnellings: again, cds ends is closed

gene/delete.go
DanielSnellings: cds ends is closed

DanielSnellings (commented):
I've made a few updates that hopefully make the coordinates a bit easier to follow. One open question is whether cdsStart/End should be changed to left-closed right-open instead of being fully closed as it is now. Being fully closed simplifies some of the logic, but as we are used to left-closed right-open I think it may generate more confusion than it simplifies.

gene/delete.go
craiglowe: I understand now.  I think many of my comments are explained by this misunderstanding of mine.  I would vote for zero-based left-closed right-open coordinates to be used everywhere besides specific formats/structs like gtf or vcf where we didn't have control over how that file format was originally defined.  I would also suggest we try to keep those index/interval deviations as isolated to those structs/files as possible.  I think at times other coordinate/interval systems can make things a little easier to write, but my concern is that I think it adds a level of complexity to reading/reviewing/modifying the code that can make it more difficult for someone to jump right in.  The reader needs to hold in their head the coordinate/interval system being used for each variable as well as the high-level idea of what it is holding.  I think it also adds a layer of complexity to math that involves indices/intervals that are off different systems, but being compared or used to update one of them.  I think it is also nice that golang uses the left-closed right-open in a way with how they reslice slices.

craiglowe (commented):
I think most of my comments were due to not understanding the interval conventions being used.

rimangan (approved):

craiglowe (approved):
Great work!

RavenYL (approved):

PR #359, opened 2021-01-13
Gene Refactor 4 (final)

DanielSnellings:
See PR Gene Refactor 3 before reviewing

Adds insertions

New Files:
`insertion.go`
`insertion_test.go`
`dna/compare.go`
`dna/codon.go`


RavenYL (approved):

dna/codon.go
```diff
@@ -365,6 +365,19 @@ func TranslateSeqToTer(b []Base) []AminoAcid {
 	return answer
 }
 
+//TranslateToTer takes a sequence of in-frame DNA bases and translates to the first stop codon. Returns a slice of amino acids the distance to the stop codon in base zero.
```
craiglowe, line 368: This comment talks about a distance to the stop codon, but I don't see that being calculated.

dna/codon.go
```diff
@@ -365,6 +365,19 @@ func TranslateSeqToTer(b []Base) []AminoAcid {
...
+func TranslateToTer(b []Base) []AminoAcid {
+	var answer []AminoAcid
+	codons := BasesToCodons(b)
+	for i := 0; i < len(codons); i++ {
+		answer = append(answer, TranslateCodon(codons[i]))
+		if answer[len(answer)-1] == Stop {
+			break
```
craiglowe, line 375: Is it more clear to return right here rather than breaking to then end up at the later return?

gene/insert.go
```diff
@@ -0,0 +1,326 @@
...
+	var err error
+
+	genomeIndexPos, err = insertionPreRunChecks(g, genomePos, alt)
+
+	// Update genome sequence
+	g.genomeSeq = dna.Insert(g.genomeSeq, genomeIndexPos+1, alt)
+
+	// Update CDS starts and ends
+	for idx, val := range g.cdsStarts {
+		if val <= genomeIndexPos && g.cdsEnds[idx] > genomeIndexPos {
```
craiglowe, line 25: Is it more natural for this 'if' to come after the one below as an "else if" and lose the first clause?  In english it would be: if the insertion comes before the start, do this, else if it comes before the end, do this.

gene/insert.go
```diff
@@ -0,0 +1,326 @@
...
+			}
+		}
+
+		answer.CdnaPos, answer.CdnaDist, err = GenomicPosToCdna(g, genomePos)
+		frame := (codingPos + 1) % 3
+		var currCodon dna.Codon
+
+		if frame != 0 { // if insertion disrupts a codon, note the codon being changed
+			currCodon, err = CdnaPosToCodon(g, codingPos)
+			if err != nil { //TODO REMOVE
```
craiglowe, line 66: It is not clear to me why this error check should be removed.

gene/insert.go
```diff
@@ -0,0 +1,326 @@
...
+
+		switch fillVal {
+		case UtrFive:
+			utrFiveOffset := 0
+
+			for i := 0; i <= genomeIndexPos; i++ {
+				if g.featureArray[i] == UtrFive {
+					utrFiveOffset++
+				}
+				if i > len(g.featureArray) { // case of no codingSeq
```
craiglowe, line 138: Can this if evaluate to true?  Three lines above we access g.featureArray[i] and I don't see anything that updates 'i' between there and here.  It seems like if I were to be > len(g.featureArray) on this line, we would have hit an index out of range three lines before.

gene/insert.go
```diff
@@ -0,0 +1,326 @@
...
+			}
+
+		case UtrThree:
+			utrThreeOffset := 0
+
+			for i := genomeIndexPos; g.featureArray[i] < 0; i-- {
+				if g.featureArray[i] == UtrThree {
+					utrThreeOffset++
+				}
+				if i <= 0 { // case of no codingSeq
```
craiglowe, line 157: The everything annotated as 3' UTR when there is no CDS?

gene/insert.go
```diff
@@ -0,0 +1,326 @@
...
+		g.featureArray = append(g.featureArray, tmpSlice...)
+		copy(g.featureArray[genomeIndexPos+1+len(alt):], g.featureArray[genomeIndexPos+1:])
+		for i := 0; i < len(alt); i++ {
+			g.featureArray[genomeIndexPos+1+i] = fillVal
+		}
+
+		var endCdnaOffset int
+		answer.CdnaPos, answer.CdnaDist, err = GenomicPosToCdna(g, genomePos+1)
+		_, endCdnaOffset, err = GenomicPosToCdna(g, genomePos+1+(len(alt)-1))
+		if err != nil { //TODO REMOVE
```
craiglowe, line 180: I am not sure why this error check would be removed.

gene/insert.go
```diff
@@ -0,0 +1,326 @@
...
+	transitionCodon := append(shiftedCds[len(shiftedCds)-frameOffset:], utrThree[:3-frameOffset]...)
+	answer = append(answer, dna.TranslateSeq(transitionCodon)[0])
+	if answer[len(answer)-1] == dna.Stop {
+		return answer
+	}
+
+	remainingUtrThree := utrThree[3-frameOffset:]
+
+	for i := 0; i < len(remainingUtrThree); i += 3 {
+		if i+3 > len(remainingUtrThree) {
```
craiglowe, line 317: Should this be in the for loop condition instead of here?

craiglowe (commented):
Wonderful to have this functionality joining the code base!  I do struggle reasoning through the really long functions.  I had minor comments; no bugs, but a couple parts I didn't understand.

RavenYL (approved):

gene/insert.go
DanielSnellings: No, this is catching a corner case where the entire gene is annotated as a 3' UTR. It won't actually change the annotation at all, but it will let insertAdjust know that the insertion should be made at the start of the gene + the number of bases preceding the genomeIndexPos.

gene/insert.go
DanielSnellings: removed the todo

gene/insert.go
DanielSnellings: Yes

DanielSnellings (commented):
I have responded to the existing comments. I think simplifying this code will be a bit of a task so if it is alright with people I would like to merge this and work on refactoring this code again in a future PR as there are now several things that have accumulated through these 4 PRs that are not wrong, but could be simplified. This is definitely not a final product, but it is working and should give a good template for simplification. That said people are welcome to block this from merging prior to simplification if they feel strongly.

cfauci (approved):

craiglowe (approved):
Great to see this!

RavenYL (approved):

PR #360, opened 2021-01-15
Proposed Changes To DNA

DanielSnellings:
These are some changes I am proposing for the `dna` package. Some are stylistic, some are functional. All tests pass within the `dna` package, but these changes break most of gonomics (though mostly because it breaks fasta). I am still in the process of adding example functions and fixing the rest of gonomics, but I wanted to put this up for people to make comments on. What do you all think of these changes?


dna/codon.go
```diff
@@ -1,264 +1,245 @@
...
-		answer = append(answer, TranslateCodon(codons[i]))
+// TranslateSeq takes a sequence of DNA bases and translates it into a slice of Amino acids.
+func TranslateSeq(b []Base) ([]AminoAcid, error) {
+	answer := make([]AminoAcid, len(b)/3)
+	codons, err := BasesToCodons(b)
+	for i := range codons {
+		answer[i], err = TranslateCodon(codons[i])
 	}
-	return answer
+	return answer, err
```
craiglowe, line 187: I think we may have a bug in the error handling code here.  Right now it only returns the err value associated with the last call to TranslateCodon, but we ignore all previous calls to TranslateCodon as well as the call to BasesToCodons.  I see three possible ways to do it:
1) return first error encountered
2) return "worst" error encountered
3) return all errors

There would also be a few different ways to handle what gets returned for the []AminoAcid where there are errors at some locations but not others.  We could return all AminoAcids that could be translated without an error, just those up to the first error, or all that could be translated with some sort of "filler" codon in the places where there was an error.  Maybe similar to N or Gap in DNA.
I think this goes to a bigger question of what do we want the programs to do when there is an "upstream" error detected in either the input or in "upstream" code that has caused something that "does make sense" to appear at some point in the program.  I think we have a few choices:
1) Abort and make the user/programmer fix it
2) Try to process parts of the data that make sense and return the answer for those, ignoring the other parts
3) Take a best-guess as to what the parts that don't make sense might be.

There may also be other choices here that I have not thought of.  Right now we are mostly programming in the style of 1, but I am interested to hear everyone's thoughts.  2 and 3 make me nervous because I worry it can lead to either us, or people who use our code, publishing incorrect results.  I think 2 and 3 will also lead to a lot of extra thought (e.g. what should we do when we get X as input?) and extra code (implementation of handling different types of errors and doing the best we can) as well as additional testing code to make sure all the error handling is working as expected

dna/codon.go
```diff
@@ -1,264 +1,245 @@
...
+		s.WriteString(str)
 	}
-	return answer
+	return s.String(), finalErr
 }
 
 //TranslateToShortString converts a sequence of DNA bases into a string of one character amino acid symbols.
-func TranslateToShortString(b []Base) string {
+func TranslateToShortString(b []Base) (string, error) {
 	AllToUpper(b)
```
craiglowe, line 218: Does this modify the input?

dna/codon.go
```diff
@@ -1,264 +1,245 @@
...
+	a, err := TranslateSeq(b)
+	return PolypeptideToShortString(a), err
 }
 
-//TranslateToString converts a sequence of DNA bases into a string of three character amino acid symbols.
-func TranslateToString(b []Base) string {
+// TranslateToString converts a sequence of DNA bases into a string of three character amino acid symbols.
+func TranslateToString(b []Base) (string, error) {
+	var finalErr error
 	AllToUpper(b)
```
craiglowe, line 226: Same as above.  Does this modify the input?

dna/codon.go
```diff
@@ -1,264 +1,245 @@
...
+// PolypeptideToString converts a slice of AminoAcids into a string of three character amino acid symbols.
+func PolypeptideToString(a []AminoAcid) (string, error) {
+	var s strings.Builder
+	s.Grow(len(a) * 3)
+	var str string
+	var err, finalErr error
+	for i := range a {
+		str, err = AminoAcidToString(a[i])
+		if err != nil {
+			finalErr = err
```
craiglowe, line 209: This relates to a previous comment where we would need to think if the vision is to process data until the first error and then return the "error-free" part, or if we want to keep going and do the best with what we have.  One option would be to return right here.

dna/codon.go
```diff
@@ -1,264 +1,245 @@
...
+		}
+		s.WriteString(str)
```
craiglowe, line 211: When err is not nil, I think we are returning an asci zero and adding it to the string?  I think I am able to follow the values of the errors as errors get passed up between functions, but I am having a hard time following the data values being passed up.  I think the function comments will need to go more into what data is given back when errors are passed.  Right now I think we are passing up the zero value of variables when there is an error here?

dna/compare.go
```diff
@@ -5,20 +5,43 @@ import (
...
+	} else {
+		a = rune(alpha)
+		b = rune(beta)
 	}
-	if alpha < beta {
-		return -1
-	} else if alpha > beta {
-		return 1
+
+	if alpha == Gap || alpha == Dot || beta == Gap || beta == Dot {
```
craiglowe, line 20: I am not sure why we switch 1 and -1 when either alpha or beta are a dash.  Is it so that A < -?  Why not follow ascii convention?

craiglowe (commented):
Great to see this work to further improve a package that gets a ton of use!  I took a couple hours and made it about halfway through.  I think there are some high-level decisions we need to make as a group about error handling and the vision for what will get returned as the value of the function when there are errors.  Depending on how we answer these high-level questions, I think function comments will need to describe what gets returned during errors.

dna/convert.go
```diff
@@ -1,156 +0,0 @@
...
-	return rec[start:end]
-}
-
-//BaseToString converts a DNA base to a string by casting a BaseToRune result to a string.
-func BaseToString(b Base) string {
-	return string(BaseToRune(b))
-}
-
-//StringToBases parses a string into a slice of DNA bases
-func StringToBases(s string) []Base {
```
craiglowe, line 128: I think this function gets used a lot.  Is there a way to keep it?

dna/dna.go
```diff
@@ -1,38 +1,184 @@
...
-	T   Base = 3
-	N   Base = 4
-	a   Base = 5
-	c   Base = 6
-	g   Base = 7
-	t   Base = 8
-	n   Base = 9
-	Gap Base = 10
-	Dot Base = 11
+	A      Base = 'A'
```
craiglowe, line 10: What do we gain by using ascii values instead of values counting up from zero?  I think this going to have some ripple effects where conversion to using less bits are not as straight forward and that bases can not be used as indices into slices (like for a scoring matrix).  Not against this change if there are some strong benefits, but I think it will have some minor drawbacks.

dna/dna.go
```diff
@@ -1,38 +1,184 @@
...
+	T      Base = 'T'
+	N      Base = 'N'
+	LowerA Base = 'a'
+	LowerC Base = 'c'
+	LowerG Base = 'g'
+	LowerT Base = 't'
+	LowerN Base = 'n'
+	Gap    Base = '-'
+	Dot    Base = '.'
+	Nil    Base = '*'
```
craiglowe, line 22: What is the vision for Nil?  Different than N?

dna/dna.go
```diff
@@ -1,38 +1,184 @@
...
 )
 
-func CreateAllGaps(numGaps int64) []Base {
+// CreatAllGaps creates a DNA sequence of Gap with length of numGaps
+func CreateAllGaps(numGaps int) []Base {
 	answer := make([]Base, numGaps)
-	for i := 0; i < len(answer); i++ {
+	for i := range answer {
```
craiglowe, line 28: Is this change for style or performance?

dna/dna.go
```diff
@@ -1,38 +1,184 @@
...
 	return answer
 }
 
-func CreateAllNs(numGaps int64) []Base {
-	answer := make([]Base, numGaps)
-	for i := 0; i < len(answer); i++ {
+// CreatAllN creates a DNA sequence of N with length of numGaps
+func CreateAllNs(numN int) []Base {
+	answer := make([]Base, numN)
+	for i := range answer {
```
craiglowe, line 37: Same comment as above

dna/dna.go
```diff
@@ -1,38 +1,184 @@
...
 }
+
+//BaseToString converts a DNA base to a string by casting a BaseToRune result to a string.
+func BaseToString(b Base) string {
+	return string(b)
+}
+
+//StringToBases parses a string into a slice of DNA bases
+func StringToBases(s string) []Base {
+	return []Base(s)
```
craiglowe, line 50: I am worried about a lack of input checking here where almost any string would get turned into bases, but the []Base would have strange values in it.  Before I think we had error checking here to make sure someone didn't give us a protein sequence or whatever else they might do.

dna/dna.go
```diff
@@ -1,38 +1,184 @@
...
+		s.WriteByte(byte(bases[i]))
+	}
+	return s.String()
+}
+
+// ByteSliceToDnaBases will convert a slice of bytes into a slice of Bases.
+func ByteSliceToDnaBases(b []byte) []Base {
+	answer := make([]Base, len(b))
+	for i := range b {
+		answer[i] = Base(b[i])
```
craiglowe, line 67: I am again worried about a lack of checks on the input

dna/dna.go
```diff
@@ -1,38 +1,184 @@
...
+func CreateAllNs(numN int) []Base {
+	answer := make([]Base, numN)
+	for i := range answer {
 		answer[i] = N
 	}
 	return answer
 }
+
+//BaseToString converts a DNA base to a string by casting a BaseToRune result to a string.
+func BaseToString(b Base) string {
```
craiglowe, line 44: I think some of these functions are coming over from convert.go.  There might be some benefits you are thinking of to having less files in a package, but one thing I like about having more files is that there are less git collisions when multiple people are checking in changes.

dna/examine.go
```diff
@@ -20,15 +23,15 @@ func Count(seq []Base) (ACount int, CCount int, GCount int, TCount int, NCount i
 			TCount++
 		case N:
 			NCount++
-		case a:
+		case LowerA:
```
craiglowe, line 26: This name change is nice so that lowercase A can be seen outside of the dna package.

dna/examine.go
```diff
@@ -37,51 +40,63 @@ func Count(seq []Base) (ACount int, CCount int, GCount int, TCount int, NCount i
...
+func baseDist(a Base, b Base) int {
 	if a == b {
 		return 0
 	}
 	return 1
 }
 
-func Dist(a []Base, b []Base) int {
+// Dist returns the number of bases that do not match between the input sequences.
+// Input sequences must be the sample length.
```
craiglowe, line 71: typo: sample -> same

dna/examine.go
```diff
@@ -37,51 +40,63 @@ func Count(seq []Base) (ACount int, CCount int, GCount int, TCount int, NCount i
...
-	case t:
+	case LowerT:
 		return true
 	default:
 		return false
 	}
 }
 
+// DefineBase returns false if the input base is an N, Gap, Dot, or Nil.
 func DefineBase(b Base) bool {
```
craiglowe, line 100: Maybe this is an old typo?  Should it be "DefinedBase"?

dna/modify.go
```diff
@@ -1,139 +1,185 @@
...
 
-func Complement(bases []Base) {
-	for i := 0; i < len(bases); i++ {
-		bases[i] = ComplementSingleBase(bases[i])
+// complementSwap complements the input bases and swaps their positions in the return.
+func complementSwap(alpha, beta Base) (Base, Base, error) {
+	var err error
+	alpha, err = ComplementSingleBase(alpha)
+	if err != nil {
+		return beta, alpha, err
```
craiglowe, line 139: I think the functions that return errors will need  descriptions of what they return  upon errors.  This one will return the original bases/values swapped, but not complemented.

dna/modify.go
```diff
@@ -1,139 +1,185 @@
...
 	RangeToUpper(bases, 0, len(bases))
 }
 
+// AllToLower changes all bases in a sequence to lowercase.
 func AllToLower(bases []Base) {
 	RangeToLower(bases, 0, len(bases))
 }
 
-func ComplementSingleBase(b Base) Base {
+// ComplementSingleBase returns the nucleotide complementary to the input base.
```
craiglowe, line 75: I think we will need additional documentation about what happen when there is an error.  In this case I think they get the starting base/value back.  Maybe we also needs tests for error inputs?

dna/modify.go
```diff
@@ -1,139 +1,185 @@
...
 		}
 	}
 	return ans
 }
 
-// all base positions are zero based and left closed, right open
-func Delete(seq []Base, delStart int64, delEnd int64) []Base {
-	if delStart >= delEnd || delStart < 0 || delEnd > int64(len(seq)) {
-		log.Fatalf("Error: deletion interval from %d to %d is not valid.\n", delStart, delEnd)
+// Delete removes bases from and sequence of bases.
```
craiglowe, line 161: typo: and->a

craiglowe (commented):
Great work and nice to see a ton of improvements!

DanielSnellings (commented):
Craig made some really good comments that call for significant changes. I am closing this branch.

PR #361, opened 2021-01-17
Vcf samples bug fix

rimangan:
Small bug fix in VCF. The sample fields are separated by tabs in VCF files, but the version on main puts the tab in the wrong location for samples with multiple format fields. This is now fixed. Passes tests


craiglowe (dismissed):
Good catch!
For the future, I think it might be good to add a test where we have code that passes tests, but has a bug.  This would hopefully guard against the same or similar bugs coming back in the future.

DanielSnellings (dismissed):

rimangan:
> Good catch!
> For the future, I think it might be good to add a test where we have code that passes tests, but has a bug. This would hopefully guard against the same or similar bugs coming back in the future.

New commit includes a test for the case that was previously broken in VCF.

DanielSnellings (dismissed):

rimangan:
Newest commit 1/20/21 includes a fix for two critical issues. First, the current main vcf breaks when it encounters single haplotype data such as for chrX or chrY gVCFs because a check for err != nil should have read err == nil in vcf.go. Second, when format is blank (represented by "."), the program no longer attempts to parse subsequent fields, which can cause vcf to break if the line ends with a tab (such as data made from the unix cut function, the test.vcf file in annotation_test of the vcf package was one such program where this could break, but this is now resolved. Passes tests.

vcf/vcf.go
```diff
@@ -122,6 +122,12 @@ func ParseNotes(data string, format []string) []GenomeSample {
 	if format[0] != "GT" && format[0] != "." { //len(format) == 0 checks for when there is info in the notes column but no format specification
 		log.Fatalf("VCF format files with sample information must begin with \"GT\" as the first format column or be marked blank with a period. Here was the first format entry: %s.\nError parsing the line with this Notes column: %s.\n", format[0], data)
 	}
+
+	if format[0] == "." { //if the format column is blank, we do not need to parse further.
```
DanielSnellings, line 126: does something break without this check? If so, could you add a test (or at least a comment) for this so it doesn't get removed in the future?

rimangan: We actually do run into another issue here that is unresolved. The VCF file in the gtf package contains a space and then a "." in the format column, and thus can't be read and written with fidelity. We could check for format[0] contains "." to avoid this, but I'm not sure what is standard for VCF specifications. I plan to copy that VCF file into the VCF testdata and test it along with the new tests here to handle a case where there is no format or sample data.

cshoben: Could you do format[0] or format[1] ==  "." for situations where a space is added before the "."?

cshoben (approved):

vcf/vcf.go
cfauci: i don't know vcf that well, but we typically set up formatting as tab-based so a space followed directly by a period wouldn't be read as separate fields in the reader. format[1] would point to the next field (the field directly **after** the first tab in the file) if you go to func processVcfLine in vcf.go you can see that the lines of the vcf files are split on the tab (line 90) and then it identifies the fields based on which parts of the file is being processed (so data[0] which is the info directly **before** the first tab is always processed as the chr number)

cfauci (approved):

vcf/vcf.go
```diff
@@ -139,12 +145,11 @@ func ParseNotes(data string, format []string) []GenomeSample {
 			alleles = strings.SplitN(fields[0], "/", 2)
 			answer[i] = GenomeSample{AlleleOne: common.StringToInt16(alleles[0]), AlleleTwo: common.StringToInt16(alleles[1]), Phased: false, FormatData: fields}
 		} else {
-			//Deal with single haps. There might be a better soltuion, but I think this should work.
 			n, err = strconv.ParseInt(fields[0], 10, 16)
-			if err != nil && n < int64(len(text)) {
+			if err == nil && n < int64(len(text)) {
```
craiglowe, line 149: I am not sure I understand the comparison to Len(text). I think len(text) is the number of individuals with genotype data in this vcf.  It seems like n should be less than or equal to the number of alleles at this position, which maybe we don't know within this function?

craiglowe (approved):

PR #363, opened 2021-01-19
Raven global alignment 3

RavenYL:
-Expanded usage message to include usage command and the -faOut option
-Made main function more concise, and moved everything other than command line arguments into the function globalAlignment, which now calls other helper functions
-Test was successful

### DIFF
--- a/cmd/globalAlignment/globalAlignment.go
+++ b/cmd/globalAlignment/globalAlignment.go
@@ -108,7 +108,7 @@ func usage() {
 		"./globalAlignment - chelsea's global alignment\n" +
 			" Align 2 .fasta files, each with only 1 sequence\n" +
 			"Usage:\n" +
-			"	faTarget faQuery\n" +
+			"	globalAlignment target.fasta query.fasta\n" +
 			"options:\n" +
 			"	-faOut=fasta MSA output filename\n")
 	flag.PrintDefaults()
--- a/cmd/globalAlignment/globalAlignment.go
+++ b/cmd/globalAlignment/globalAlignment.go
@@ -28,6 +28,35 @@ func CountSeqIdx(inputFile *fileio.EasyReader) int {
 	return seqIdx
 }
 
+//raven did not put this helper function into the globalAlignment function because it is used in the globalAlignment function
+//faONe is target, faTwo is query
+func cigarToGraph(target *fasta.Fasta, query *fasta.Fasta, aln []align.Cigar) *simpleGraph.SimpleGraph {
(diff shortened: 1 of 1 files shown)


cmd/globalAlignment/globalAlignment.go
```diff
@@ -94,37 +68,62 @@ func main() {
...
+	simpleGraph.PrintGraph(genomeGraph)
+}
+
+//raven edited this block to specify only 1 sequnce is expected in each fasta file and add Usage nad options
+func usage() {
+	fmt.Print(
+		"./globalAlignment - chelsea's global alignment\n" +
+			" Align 2 .fasta files, each with only 1 sequence\n" +
+			"Usage:\n" +
+			"	faTarget faQuery\n" +
```
cshoben, line 103: faTarget and faQuery are the input variables. Would it make more sense to label line 102 as input variable instead of usage?

cshoben (approved):
The reorganization is very fitting. I'm really enjoying getting to see this code improve. Question about line 36 (I know it's old code), but maybe we can talk about it at code review.

RavenYL:
> Can you update the global alignment test file too? I looks like the test is calling cigarToGraph which is now being deleted.

Yes! I ended up adding the cigarToGraph function back to globalAlignment.go because output is easier to test that way, so now the test passes

cmd/globalAlignment/globalAlignment.go
RavenYL: Yes you are right. They are input variables. It seems from the other examples in the repository that "usage" is followed by what users type on the command line to run the program, so that's why I wrote the input variables on line 103

RavenYL:
> The reorganization is very fitting. I'm really enjoying getting to see this code improve. Question about line 36 (I know it's old code), but maybe we can talk about it at code review.

Thank you! Yes we can talk about line 36 at code review

cmd/globalAlignment/globalAlignment.go
cfauci: that's true, but it should also have the executable's name so it would be 'globalAlignment faTarget faQuery' (could also expand to be 'globalAlignment target.fasta query.fasta')

RavenYL: Got it! Just fixed. Thanks!

craiglowe (approved):
Great work!

edotau (approved):
LGTM!

cfauci (approved):

PR #366, opened 2021-01-25
Graph reconstruct

cfauci:
PathFinder (bestPath as helper) and BuildNodes function for graph reconstruction. I recommend looking at the files rather than the change log since i removed everything that was copied over from the reconstruct package, so it's very messy when it come to line numbering.


graphReconstruct/graphReconstruct.go
```diff
@@ -2,230 +2,91 @@ package graphReconstruct
...
-		scrap := []float64{0, 0, 0, 0} //checking one base at a time each time you call BubbleUp
-		scrap[i] = node.Stored[i]
-		if node.Up != nil {
-			//Bubble up the tree using the memory of the previous node in relation to changing position taking in probabilities of bases
-			//(node will be BubbleUp prevNode and node.Up will be the node being operated on)
-			BubbleUp(node.Up, node, scrap) //node becomes PrevNode and scrap is set to one value of prevNode.Stored in BubbleUp
-			ans[i] = root.Scrap            //root.Stored has previously assigned values (SetInternalState), you want to use whatever is returned by BubbleUp instead
-		} else if node.Up == nil {
-			ans[i] = root.Stored[i]
+	for n := 0; n < len(g.Nodes); n++ {
```
DanielSnellings, line 63: What happens if your graph has multiple starts/ends or is discontiguous?

graphReconstruct/graphReconstruct.go
```diff
@@ -2,230 +2,91 @@ package graphReconstruct
...
-		yHat := Yhat(fix)
-		internalNodes[k].Fasta.Seq = append(internalNodes[k].Fasta.Seq, []dna.Base{dna.Base(yHat)}...)
+//bestPath is the helper function for PathFinder, and recursively traverses the graph depth first to determine the most likely path from start to finish
+func bestPath(node *simpleGraph.Node, prevProb float32, prevPath []uint32) (prob float32, path []uint32) {
+	var tempProb float32 = 0
+	var tempPath []uint32
+	var finalProb float32
+	var finalPath []uint32
+
+	tempPath = append(tempPath, prevPath...)
```
DanielSnellings, line 78: may be a better way to do this than appending. I recommend either using the builtin copy function, or declaring the capacity of tempPath as len(prevPath)

DanielSnellings: On second thought, do you even need tempPath? Can you just append the next node to the prevPath (or just rename it "path") and input that to bestPath? Since you only append nodes to the path and don't change earlier nodes in the path, the underlying path should be memory stable... i think. Recursion hurts my brain so the only way to be sure would be to test it, but I think this can be done with just 1 underlying slice in memory.

DanielSnellings: I made a quick program to test this

https://play.golang.org/p/YJftVT3KJCH

I believe this will work, but in order to save yourself from constantly shifting memory during append functions you might consider setting the capacity of 'path' to len(graph.Nodes). It will be more than you need, but should make things alot faster.

cfauci: so i totally agree about the rest of it and i'm going to push an update, but i did a test to confirm and setting a length longer than necessary just adds a repeat of the sequence for the node labelled with 0 for its id.

graphReconstruct/graphReconstruct.go
DanielSnellings: Ok, thats fair. The circumstances would like a genome graph with multiple chromosomes. There is a BreakDiscontiguousGraph function, but if this is mainly for small scale stuff it is fine as is.

graphReconstruct/graphReconstruct.go
```diff
@@ -2,230 +2,89 @@ package graphReconstruct
...
-	return pos
-}
-
-func allZero(r []float64) bool {
-	for _, v := range r {
-		if v != 0 {
-			return false
-		}
+	for seq := range nodeInfo {
+		id += 1
```
craiglowe, line 25: I think I would increment id at the end of the loop instead of the beginning.  I think it fits style a little better to use the id you are given and pass back the id that the next function can begin using

graphReconstruct/graphReconstruct.go
DanielSnellings: Don't make length equal to len(graph.Nodes), make capacity equal to len(graph.Nodes). This would look like `path := make([]uint32, 0, len(graph.Nodes)). This way you start with an empty slice, but can add to it without needing to allocate more memory with append.

graphReconstruct/graphReconstruct.go
```diff
@@ -2,230 +2,89 @@ package graphReconstruct
...
+//seq to the Fasta field of the ancestors tree node
+//func FindAncSeq will loop through aligncolumns and build a single graph of all of the nodes that belong to the ancestor species after edges are created
+//run PathFinder on the graph for the anc, run seqOfPath, then turn that to a fasta for that node of the tree
+
+//seqOfPath takes in a graph and a path specified by the Node IDs and returns the seq of the path through the graph
+func seqOfPath(g *simpleGraph.SimpleGraph, path []uint32) []dna.Base {
+	var seq []dna.Base
+	for n := 0; n < len(g.Nodes); n++ {
+		for p := 0; p < len(path); p++ {
+			if g.Nodes[n].Id == path[p] {
```
craiglowe, line 49: I think we need a check to catch if this 'if' never evaluates to true for a given 'path[p]'.  This would then be an invalid path and should give an error.

graphReconstruct/graphReconstruct.go
cfauci: ah, ok i see, no i dont think we have to worry about that with this function, and I'm aware of the BreakDiscontiguous func too

craiglowe: For a while I have been wondering if our graph data struct (along with a slice of all the nodes) should have a list of all the "start" nodes (maybe the "end" nodes too).

craiglowe (commented):
Great work!  A few small comments.  Only change I would really like to see would be an error check in seqOfPath for an invalid path

graphReconstruct/graphReconstruct.go
cfauci: I think thats a great idea, i think no matter what you're doing with the graph that info is useful enough to add it to the existing struct

graphReconstruct/graphReconstruct.go
```diff
@@ -2,230 +2,91 @@ package graphReconstruct
...
-	n = 0
-	var pos int
-	for p, v := range r {
-		if v > n {
-			n = v
-			pos = p
-		}
-	}
-	return pos
+type graphColumn struct {
```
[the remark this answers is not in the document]
cfauci: I agree that's what we should do once we have it all settled

graphReconstruct/graphReconstruct_test.go
```diff
@@ -1,46 +1,117 @@
...
+	chimpNode2.Prev = []*simpleGraph.Edge{chimpEdge1}
+
+	gorillaNode1.Next = []*simpleGraph.Edge{gorillaEdge1}
+	gorillaNode2.Prev = []*simpleGraph.Edge{gorillaEdge1}
+
+	simpleGraph.PrintGraph(humanGraph)
+	simpleGraph.PrintGraph(chimpGraph)
+	simpleGraph.PrintGraph(gorillaGraph)
+
+	log.Print(nodeAlign0.AlignId)
```
craiglowe, line 76: Automating this test would be beneficial, even if the check is not perfect.  I worry that this test will always been seen as passing, even if something breaks your code.

graphReconstruct/graphReconstruct_test.go
```diff
@@ -1,46 +1,117 @@
...
-		for i := 0; i < len(leaves[0].Fasta.Seq); i++ {
-			LoopNodes(tr, i)
+func TestPathFinder(t *testing.T) {
+	path, prob := PathFinder(humanGraph)
+	log.Print("path")
+	log.Print(path)
+	log.Print("prob")
+	log.Print(prob)
+	seq := seqOfPath(humanGraph, path)
+	log.Print(dna.BasesToString(seq))
```
craiglowe, line 89: Can we compare seq to a hard-coded DNA sequence that it should match?

graphReconstruct/graphReconstruct_test.go
```diff
@@ -1,46 +1,117 @@
...
+			id = BuildNodes(iNodes[in], allAlign[i], id)
+			for _, nodes := range allAlign[i].AlignNodes {
+				for n := 0; n < len(nodes); n++ {
+					if nodes[n].Name == iNodes[in].Name {
+						speciesGraph.Nodes = append(speciesGraph.Nodes, nodes[n])
+					}
+				}
+			}
 		}
+		simpleGraph.PrintGraph(speciesGraph)
```
craiglowe, line 115: Can we also have an automated check here?

craiglowe (commented):
Code looks great!  I think we need to make sure we have tests that can fail so that we will know if something breaks these functions.

RavenYL (dismissed):

DanielSnellings (approved):

graphReconstruct/graphReconstruct.go
```diff
@@ -2,230 +2,95 @@ package graphReconstruct
...
+//seq to the Fasta field of the ancestors tree node
+//func FindAncSeq will loop through aligncolumns and build a single graph of all of the nodes that belong to the ancestor species after edges are created
+//run PathFinder on the graph for the anc, run seqOfPath, then turn that to a fasta for that node of the tree
+
+//seqOfPath takes in a graph and a path specified by the Node IDs and returns the seq of the path through the graph
+func seqOfPath(g *simpleGraph.SimpleGraph, path []uint32) []dna.Base {
+	var seq []dna.Base
+	var foundInGraph = false
+	for n := 0; n < len(g.Nodes); n++ {
+		for p := 0; p < len(path); p++ {
```
craiglowe, line 50: Right inside this loop over 'p', I think we need to set FoundInGraph to false each time so that foundInGraph is by default false for each element of path until that element is found

craiglowe (commented):
I think I found one tiny bug where a bool is not getting reset to false.  Otherwise looks great!

craiglowe (approved):
Great work!

RavenYL (approved):

DanielSnellings (approved):

PR #368, opened 2021-02-03
bedMaxWig update

rimangan:
This is a change I made to the bedMaxWig function ages ago but I never pushed it because the changes were made on the HARDAC. Rectifying that now. bedMaxWig is a function that takes bed files, compares them to a wig file, and adds the max wig value contained within each bed record as the "score" field for the bed. This is a fix we wanted for Raven's epigenomic annotation of human insertions.


RavenYL (approved):
Thank you Riley!

RavenYL (approved):

RavenYL (dismissed):

cfauci (dismissed):

cmd/sampleVcf/sampleVcf.go
```diff
@@ -0,0 +1,100 @@
...
+		for _, i := range sequentialSlice {
+			outHeaderSampleList = append(outHeaderSampleList, sampleList[i])
+		}
+
+		vcf.HeaderUpdateSampleList(header, outHeaderSampleList)
+
+		var outSamples []vcf.GenomeSample
+
+		for _, i := range records {
+			outSamples = make([]vcf.GenomeSample, 0)
```
DanielSnellings, line 48: Minor point, if you are sampling a large number of variants than this will slow you down a lot as you need to keep reshuffling memory as you append elements beyond the capacity of this slice. You could speed this up by doing `outSamples = make([]vcf.GenomeSample, 0, len(sequentialSlice)`. Giving it the cap will make append much faster (as i understand it it would be equivalent to doing outSamples[i] = i.Samples[j]). Better yet I think you could even declare outSamples outside of the loop and just reuse the memory each run.

That said, I suspect this is not a performance-intensive task so I will approve anyways.

cmd/sampleVcf/sampleVcf.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"github.com/vertgenlab/gonomics/common"
+	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/vcf"
+	"log"
+	"math/rand"
+)
+
+//sampleVcf takes a VCF file and returns a random subset of variants to an output VCF file. Can also retain a random subset of alleles from gVCF data (diploid, does not break allele pairs)
+func sampleVcf(inFile string, outFile string, numVariants int, numSamples int, randSeed bool, setSeed int64) {
```
DanielSnellings, line 14: This seems like a useful function. Maybe one day it would be worth moving this to the vcf package.

DanielSnellings (dismissed):

RavenYL (approved):

DanielSnellings (approved):

cfauci (approved):

PR #370, opened 2021-02-09
gtfToBed

rimangan:
I wanted to run overlapSelect on gene data but found it was perhaps easiest to just convert gtf data into bed format first. Here's a quick tool to do so. All gtf data other than the required bed fields are put in the annotations columns, so this program is lossless, and all gtf data can theoretically be recovered. 
Passes tests.

### DIFF
--- a/cmd/gtfToBed/gtfToBed.go
+++ b/cmd/gtfToBed/gtfToBed.go
@@ -1,13 +1,13 @@
 package main
 
 import (
-	"github.com/vertgenlab/gonomics/fileio"
-	"github.com/vertgenlab/gonomics/common"
-	"github.com/vertgenlab/gonomics/bed"
-	"strings"
-	"log"
 	"flag"
 	"fmt"
+	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/common"
+	"github.com/vertgenlab/gonomics/fileio"
+	"log"
+	"strings"
 )
 
 func gtfToBed(fileName string, outFile string) {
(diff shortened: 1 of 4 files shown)


DanielSnellings (dismissed):

cfauci (approved):

cmd/gtfToBed/testdata/testOut.bed
```diff
@@ -0,0 +1,7 @@
+chr1	173477335	173488815	transcript	0	+	refGene	.	+	.	gene_id "PRDX6"; transcript_id "NM_004905"; gene_name "PRDX6";
```
craiglowe, line 1: This might confuse some people because there is a set way to encode genes in bed format and this output we are testing against is not it.  Maybe this is a code review question?  I think we are starting to call things beds that are not beds.  Maybe we should come up with our own name and then keep beds strictly beds?
https://genome.ucsc.edu/FAQ/FAQformat.html#format1

cmd/gtfToBed/gtfToBed.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/common"
+	"github.com/vertgenlab/gonomics/fileio"
+	"log"
+	"strings"
+)
+
+func gtfToBed(fileName string, outFile string) {
+	file := fileio.EasyOpen(fileName)
+	defer file.Close()
```
craiglowe, line 15: Might be good to move the closes to the end and check the error that they return.

cmd/gtfToBed/gtfToBed.go
```diff
@@ -0,0 +1,62 @@
...
+
+	var line string
+	var annotationStringSlice []string
+	var currBed bed.Bed
+	var doneReading bool = false
+
+	for line, doneReading = fileio.EasyNextRealLine(file); !doneReading; line, doneReading = fileio.EasyNextRealLine(file) {
+		words := strings.Split(line, "\t")
+		annotationStringSlice = make([]string, 1) //annotation contains all fields except the position information and name, so four fields are not used here.
+		annotationStringSlice[0] = words[1]
```
craiglowe, line 28: I think this is a bigger question, but maybe we should think about keep bed to be strictly a bed and then creating a new file format/struct that is similar to bed, but allows fields of free annotation after the first few?  I am worried that we have invented our own file format that is similar to beds, but not a bed, yet we are calling it a bed.

craiglowe (commented):
I think people will really like this functionality.  I think we may be at a fork in the road where we need to think about creating a few format that is called something different from bed and allows free text fields at the end.

cfauci (approved):

DanielSnellings (approved):

craiglowe (approved):
Great to have this!

cshoben (approved):

RavenYL (approved):

PR #371, opened 2021-02-10
Fileio Update

DanielSnellings:
A few small updates to the `fileio` package. 

Major Changes:
`SimpleReader` has been renamed to `ByteReader` as I think the new name more explicitly describes the functionality. See the updated GoDoc for ByteReader for some more details on how it differs from EasyReader. 

`EasyReader` now has a `Read` method so it can implement `io.Reader` 

Bug fixed in PSL that added an extra empty element to an integer slice (PSL tests also had a bug that made it look like it passed)

`fileio` no longer imports `common`. Instead i implemented `ExitIfError` as `panicOnErr`. 

The `Close()` method for `EasyReader` now returns an error

### DIFF
--- a/fileio/fileio.go
+++ b/fileio/fileio.go
@@ -3,37 +3,38 @@ package fileio
 import (
 	"bufio"
 	"fmt"
-	"github.com/vertgenlab/gonomics/common"
 	"io"
 	"log"
 	"os"
 	"strings"
 )
 
+// MustCreate creates a file with the input name. Panics if not possible.
 func MustCreate(filename string) *os.File {
 	file, err := os.Create(filename)
-	common.ExitIfError(err)
+	panicOnErr(err)
 	return file
 }
 
+// MustOpen opens the input file. Panics if not possible.
 func MustOpen(filename string) *os.File {
 	file, err := os.Open(filename)
-	common.ExitIfError(err)
+	panicOnErr(err)
 	return file
 }
 
-// returns the next line of the file (might be a comment line)
-// returns true if the file is done
+// NextLine returns the next line of the file (might be a comment line).
+// Returns true if the file is done.
 func NextLine(reader *bufio.Reader) (string, bool) {
 	var line string
 	var err error
 	line, err = reader.ReadString('\n')
 	if err != nil && err != io.EOF {
-		common.ExitIfError(err)
+		panicOnErr(err)
 	}
 	if err == io.EOF {
 		if line != "" {
-			log.Fatalf("Error: last line of file didn't end with a newline character: %s\n", line)
+			log.Panicf("Error: last line of file didn't end with a newline character: %s\n", line)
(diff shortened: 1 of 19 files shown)


rimangan (approved):

fileio/fileio.go
```diff
@@ -128,3 +134,10 @@ func ReadFileToSingleLineString(filename string) string {
 	}
 	return catInput
 }
+
+// panicOnErr will call a blank panic if the input error != nil
+func panicOnErr(err error) {
```
craiglowe, line 139: I have grown to think that having a "common" package is not a good style, but does it make sense to have a public version of this function somewhere?

fileio/simpleio.go
```diff
@@ -27,24 +35,24 @@ type SimpleReader struct {
 
 // Read reads data into p and is a method required to implement the io.Reader interface.
 // It returns the number of bytes read into p.
-func (reader *SimpleReader) Read(b []byte) (n int, err error) {
+func (reader *ByteReader) Read(b []byte) (n int, err error) {
 	return reader.Read(b)
 }
 
 // NewSimpleReader will process a given file and performs error handling if an error occurs.
```
craiglowe, line 42: This comment line might need to change

fileio/simpleio.go
```diff
@@ -27,24 +35,24 @@ type SimpleReader struct {
...
-// SimpleReader will prcoess gzipped files accordinging by performing a check on the suffix
+// ByteReader will process gzipped files accordingly by performing a check on the suffix
 // of the provided file.
-func NewSimpleReader(filename string) *SimpleReader {
+func NewSimpleReader(filename string) *ByteReader {
```
craiglowe, line 45: Should this change to NewByteReader?

fileio/testdata/smallTest
```diff
@@ -0,0 +1,3 @@
+#shhhh this line is a secret
```
craiglowe, line 1: ha ha, always funny to see lines like this

craiglowe (dismissed):
Wonderful work here!  Thank you Dan!

fileio/fileio.go
DanielSnellings: I am not sure. I also tend to dislike the common package, especially since 90% of common imports are just to use ExitIfError. I don't know what is idiomatic in this case and I don't have very strong feelings either way but I am slightly leaning towards each package that wants to use this kind of function implementing their own. That said I am happy to have my mind changed if there are reasons a public function would be better.

fileio/simpleio.go
DanielSnellings: Nice catch, see updated PR

cfauci (approved):

craiglowe (approved):
Great work!

rimangan (approved):

PR #372, opened 2021-02-11
small bug fix for sampleVcf

rimangan:
In sampleVcf, we parse the VcfHeader to update the list of samples. If a Vcf did not have a header, the old version would fatal out. I think this is a bit limiting, so instead we handle this case. Passes tests.


craiglowe (approved):
Nice improvement!

cfauci (approved):

DanielSnellings (approved):

PR #374, opened 2021-02-13
VcfFilter by Group

rimangan:
New functionality in VcfFilter to select certain sample columns to be retained in the output using the popgen Group file format. Passes tests. New test for VcfFilter and Group in popgen.

### DIFF
--- a/cmd/vcfFilter/testdata/expectedOut.vcf
+++ b/cmd/vcfFilter/testdata/expectedOut.vcf
@@ -1,0 +1,3 @@
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Second	Fourth
+chr3	120	 TestingSamples	G	T	100.0	PASS	.	GT:DP	0|1:10	0|0:0
+chr3	250	TestingMultiAlt	A	T,A	100.0	PASS	.	GT:DP	1|0:250	0|0:0
--- a/cmd/vcfFilter/testdata/expectedOut.vcf
+++ b/cmd/vcfFilter/testdata/expectedOut.vcf
@@ -1,3 +1,2 @@
 #CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Second	Fourth
-chr3	120	 TestingSamples	G	T	100.0	PASS	.	GT:DP	0|1:10	0|0:0
-chr3	250	TestingMultiAlt	A	T,A	100.0	PASS	.	GT:DP	1|0:250	0|0:0
+chr3	120	TestingSamples	G	T	100.0	PASS	.	GT:DP	0|1:10	0|0:0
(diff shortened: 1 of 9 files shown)


cmd/vcfFilter/testdata/expectedOut.vcf
```diff
@@ -0,0 +1,3 @@
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Second	Fourth
+chr3	120	 TestingSamples	G	T	100.0	PASS	.	GT:DP	0|1:10	0|0:0
```
craiglowe, line 2: Is there an extra space before TestingSamples?

cmd/vcfFilter/testdata/test.vcf
```diff
@@ -0,0 +1,5 @@
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	First	Second	Third	Fourth
+chr3	120	 TestingSamples	G	T	100.0	PASS	.	GT:DP	0|0:44	0|1:10	1|0:34	0|0:0
```
craiglowe, line 2: Extra space before TestingSamples?

cmd/vcfFilter/vcfFilter.go
```diff
@@ -4,25 +4,61 @@ import (
...
 		if vcf.Filter(v, chrom, minPos, maxPos, ref, altSlice, minQual) {
+			v.Samples = filterRecordsSamplesToKeep(v.Samples, samplesToKeep)
 			vcf.WriteVcf(out.File, v)
 		}
 	}
 }
 
+func filterRecordsSamplesToKeep(recordSamples []vcf.GenomeSample, samplesToKeep []int) []vcf.GenomeSample {
+	var answer []vcf.GenomeSample
+	for i := 0; i < len(samplesToKeep); i++ {
```
craiglowe, line 47: Could use the "range" syntax here.  I think it would be:
for i := range samplesToKeep {

cmd/vcfFilter/vcfFilter.go
```diff
@@ -4,25 +4,61 @@ import (
...
+		answer = append(answer, recordSamples[samplesToKeep[i]])
+	}
+
+	return answer
+}
+
+func filterHeaderSamplesToKeep(samples []string, samplesToKeep []int) []string {
+	var answer []string
+	for i := 0; i < len(samplesToKeep); i++ {
```
craiglowe, line 56: range could be used here as well

vcf/filter.go
```diff
@@ -40,6 +40,9 @@ func FilterQual(v *Vcf, minQual float64) bool {
 //FilterAlt returns true if the Alt field of a Vcf entry matches a desired input Alt field, false otherwise. Order sensitive.
 func FilterAlt(v *Vcf, alt []string) bool {
 	if len(alt) > 0 && CompareAlt(v.Alt, alt) != 0 {
+		if alt[0] == "" {
```
craiglowe, line 43: I am not sure I understand this part.  It seems like if the first entry of the desired alt field is empty this is treated as a special value that overrides other logic to return true, but I am not sure why.  it also seems like passing in an alt value that is empty will cause this to always match

craiglowe (commented):
Great work!  Couple minor comments

vcf/filter.go
rimangan: So we have the case where the Alt option in the cmd is left empty, which currently is signified by "". Alt in this case is the user inputed slice of strings corresponding to the desired state of Alt in the output file, if this is empty, then we want this to pass no matter what.

cmd/vcfFilter/testdata/test.vcf
rimangan: thanks!

rimangan:
Newest commits responded to comments but also added two additional features: the ability to filter VCF files for only biallelic and substitution variants or a combination thereof.

rimangan:
One more addition: added another argument for segregatingSitesOnly, which filters out variants where all samples are 0 or all samples are 1, 2 etc.

RavenYL (approved):

cfauci (approved):

DanielSnellings (approved):

PR #376, opened 2021-02-17
Lift multiallelic and structural variants to unmapped instead of fatal.

rimangan:
In Lift, multiallelic and structural variants from VCF data cannot be transferred to new assemblies. In the main version of Lift, the program fatals if it encounters such a variant, requiring the user to filter their data before running lift. Instead, in this version, these variants are sent to the unmapped file and the program is allowed to continue. I think this is probably more convenient.


rimangan:
Added one more bug fix to this pull request: vcf.IsSubstitution is supposed to return whether a variant represents a substitution by returning false on insertions and deletions. The main version only checks for deletions, not insertions, so this is corrected here.

cfauci (approved):

DanielSnellings (approved):

craiglowe (approved):
Nice catch!

PR #378, opened 2021-02-17
single nucleotide deletions

olivier-m-boivin:
Checks sequence for deletions (represented by "-") and adds one at a random position (100% of the time) if none are found. Deletions are not generated in the start/stop codons and are limited to intronic regions for the time being (the TranslateCodon function is implemented in the simulation and is unable to handle codons which contain a deletion).


graphReconstruct/graphReconstruct.go
```diff
@@ -115,7 +115,17 @@ func SetState(node *expandedTree.ETree, position int) {
...
+			// if there is a deletion at current position, set Stored list to equal probabilities
+			isLeftDel := node.Left.Fasta.Seq[position]
+			isRightDel := node.Right.Fasta.Seq[position]
+			switch isLeftDel == 10 && isRightDel == 10 {
+			case true:
+				node.Left.Stored = []float64{0.25, 0.25, 0.25, 0.25}
+				node.Right.Stored = []float64{0.25, 0.25, 0.25, 0.25}
+			default:
+				log.Fatal("no Stored values passed to internal node")
+			}
```
cfauci, line 128: We should definitely discuss a solution that will allow stored to hold more than the four possibilities it currently holds. I'll need this going forward as well with graphReconstruct

simulate/simulate.go
```diff
@@ -501,3 +514,33 @@ func RemoveAncestors(filename string, tree *expandedTree.ETree, outputFilename s
 	outFile = outputFilename
 	fasta.Write(outFile, newFastas)
 }
+
```
cfauci, line 517: in seqContainsDels and codonContainsDels, you pass dna.Gaps as an argument. I'm not sure if there is a "correct" thing is to do here, but an alternative is to just get rid of the del argument on lines 521 and 531 and instead write " if base == dna.Gaps {" which will do the same thing without the argument being there.

cfauci (commented):
It looks good! a few comments for more general things

cfauci (dismissed):

graphReconstruct/graphReconstruct.go
```diff
@@ -115,7 +115,17 @@ func SetState(node *expandedTree.ETree, position int) {
 		SetState(node.Left, position)
 		SetState(node.Right, position)
 		if allZero(node.Left.Stored) && allZero(node.Right.Stored) {
-			log.Fatal("no Stored values passed to internal node")
+			// switch statement below accounts for potential deletions
+			// if there is a deletion at current position, set Stored list to equal probabilities
+			isLeftDel := node.Left.Fasta.Seq[position]
+			isRightDel := node.Right.Fasta.Seq[position]
+			switch isLeftDel == 10 && isRightDel == 10 {
```
DanielSnellings, line 122: I was initially confused what the 10 value was, but I now realize it is the integer value for a gap. It would be best to replace `== 10` with `== dna.Gap` to reduce confusion.

simulate/simulate.go
```diff
@@ -501,3 +514,33 @@ func RemoveAncestors(filename string, tree *expandedTree.ETree, outputFilename s
...
+// seqContainsDels checks whether a deletion is present in slice of bases.
+func seqContainsDels(seq []dna.Base, del dna.Base) bool {
```
DanielSnellings, line 519: This works, but if you are just checking for dels, do you need the dna.Base input? You could just have `if base == dna.Gap`

DanielSnellings (dismissed):
Small comments about readability, but otherwise great job.

graphReconstruct/graphReconstruct.go
cfauci: I was going to comment on this as well, i agree with dan

graphReconstruct/graphReconstruct.go
```diff
@@ -115,7 +115,17 @@ func SetState(node *expandedTree.ETree, position int) {
...
+			case true:
+				node.Left.Stored = []float64{0.25, 0.25, 0.25, 0.25}
```
craiglowe, line 124: Right now setting the base probabilities to 0.25 when there is a gap (missing data) is happening in internal nodes, but I was thinking that would happen at the leaf nodes and then the internal nodes (i.e. recursive step) would not change.

craiglowe: Does having a species with missing data (all bases set to 0.25) influence the predictions at ancestral nodes?  For example, if we predict the human chimp ancestor with human, chimp, and gorilla would the predicted human-chimp ancestor be different if I added bonobo with 0.25 for all bases?

craiglowe (commented):
Few comments on new code in reconstruction

simulate/simulate.go
```diff
@@ -501,3 +513,33 @@ func RemoveAncestors(filename string, tree *expandedTree.ETree, outputFilename s
...
+		if baseExt.Base == del {
+			return true
+		}
+	}
+	return false
+}
+
+// deleteBase replaces a single random base from provided sequence with deletion symbol(hyphen)
+func deleteBase(seq []dna.Base) []dna.Base {
+	delPos := rand.Intn(len(seq) - 3) // select random sequence position for deletion; avoid stop codon
```
craiglowe, line 539: This may be a good place to use numbers.RandIntInRange.  If not, you can use 3 + rand.Intn(len(seq)-6) to avoid both the start and stop codons

simulate/simulate.go
craiglowe: Yes, I second this suggestion.  There might be a typo in the comment above where I think it should be dna.Gap (not plural)

simulate/simulate.go
```diff
@@ -229,7 +235,13 @@ func MutateGene(inputSeq []dna.Base, branchLength float64, geneFile string) []dn
 		}
 
 	}
-	finalSequence = BaseExtToBases(newSequence)
+	mutatedSequence := BaseExtToBases(newSequence)
+	delFound := seqContainsDels(mutatedSequence, dna.Gap)
+	if delFound == false && deletions == true {
+		finalSequence = deleteBase(mutatedSequence)
```
craiglowe, line 241: I think this always puts in a deletion in the first branch in the tree.  Can we use branch length to control the likelihood of this happening on different branches?

simulate/simulate.go
```diff
@@ -182,19 +182,25 @@ func MutateGene(inputSeq []dna.Base, branchLength float64, geneFile string) []dn
...
+						newSequence = append(newSequence, thisCodon.Seq[2])
+					}
 					if stop == true {
 						newCodon = PickStop(thisCodon)
 						newSequence = append(newSequence, newCodon.Seq[0])
 						newSequence = append(newSequence, newCodon.Seq[1])
 						newSequence = append(newSequence, newCodon.Seq[2])
 					}
-					if start == false && stop == false {
+					if start == false && stop == false && delFound == false {
```
craiglowe, line 203: It seems like once a deletion happens to the sequence, it no longer goes into this if statement and therefore stops evolving with substitutions.  Is that true?

craiglowe (commented):
Great work so far.  Exciting to see the beginnings of this functionality.  I included a few comments

simulate/simulate.go
olivier-m-boivin: I believe the way I've set it up here, each codon is checked for deletions and those that contain a dna.Gap are not evolved further. I had to do this because the TranslateCodons function is called during codon evolution and it cannot handle dna.Gap characters. Codons without deletions should still proceed through the function as they did previously.

simulate/simulate.go
olivier-m-boivin: Great suggestion, that looks much cleaner. Thanks for the feedback!

simulate/simulate.go
olivier-m-boivin: Thanks for pointing me to this function, was not aware of its existence.

simulate/simulate.go
```diff
@@ -236,9 +237,16 @@ func MutateGene(inputSeq []dna.Base, branchLength float64, geneFile string, dele
...
+	delFound := seqContainsDels(mutatedSequence)
 	if delFound == false && deletions == true {
-		finalSequence = deleteBase(mutatedSequence)
+		delEvent := rand.Float64()
+		switch {
+		case delEvent <= branchLength :
+			finalSequence = deleteBase(mutatedSequence)
+		default:
+			finalSequence = mutatedSequence
+		}
```
olivier-m-boivin, line 248: @craiglowe here is a potential solution to your suggestion of making deletion probabilities a function of the branch length. I made the assumption that branch lengths are typically between 0 and 1 so any branches longer than this should always result in the addition of a deletion.

fileio/simpleio.go
```diff
@@ -115,7 +115,12 @@ func CatchErrThrowEOF(err error) {
 
 // BytesToBuffer will parse []byte and return a pointer to the same underlying bytes.Buffer
 func BytesToBuffer(reader *ByteReader) *bytes.Buffer {
-	_, err := reader.Buffer.Write(reader.line[:len(reader.line)-1])
+	var err error
```
cshoben, line 118: I see that there is no filelio/simpleio.go in gonomics, and this is producing the conflict with the merge. It looks closest to byteio.go
Would the solution be for Olivier to go back, rename the file to correct name, and then do another pull request? I'm confused by git didn't just try to make this into a new file since it's missing versus calling a conflict?

simulate/simulate.go
```diff
@@ -229,7 +236,20 @@ func MutateGene(inputSeq []dna.Base, branchLength float64, geneFile string) []dn
...
+
+	} else {
+		finalSequence = mutatedSequence
+	}
 	return finalSequence
```
cshoben, line 253: What happens if the delFound == true && deletions == true? Like if the simulateEvol.go is trying to add a deletion to a gene that already has one. That should return an error to the user, right? Or maybe tell the user, than they can't make a deletion because one already exists? I'm reading line 155 "deletions bool" as a user set feature. 
Or I'm not seeing it and it is already in the code?

cshoben (commented):
Just a few questions on the way github handles new files/file name changes and how MutateGene() will handle a scenario.

simulate/simulate.go
olivier-m-boivin: The intention here is that we only want a given sequence to have at most one deletion. This will give Christi the opportunity to see how single dna.Gap characters behave with her code. If both these variables are true, MutateGene was originally asked to simulate deletions, but the current sequence already has one so we don't want to add more. Eventually, we will remove this check altogether so that deletion mutations will only be controlled by the deletions bool argument.

craiglowe (commented):
Great to see this new functionality in simulate!  Let's get the current version of main merged into this pull request so that we have a clean pull request everyone can approve.

cfauci (approved):
Looks good!

DanielSnellings (approved):

craiglowe (approved):
Great to have this new functionality in simulate!

PR #380, opened 2021-02-18
Graph Proposal

DanielSnellings:
This is intended to be a pull request to organize and brainstorm about potential changes to SimpleGraph to enable more complex manipulations required for effective comparative genomics.

Please feel free to checkout this branch and make changes. At time of writing there are two significant changes from SimpleGraph detailed below.

**1. Sequence data is moved to a new struct with a reference in the Node struct.**
The primary benefit of this is that it allows multiple graphs to use the same underlying data. In this way we can readily create sub graphs that use the same underlying data, and can easily check for identity between subgraphs by checking the reference of the pointed to data. This functionality is particularly useful for identifying common nodes from a graph alignment block.

**2. Edge probability has been moved to the Node struct.**
This change is required to enable effective creation of subgraphs. Imagine a situation where Humans and Chimps both have Node X as a next node, but the probability for Human X != probability of Chimp X. With the old format this could be achieved by having multiple edges along the same path with some identifier to discriminate between the Human and Chimp edges, however this can be more effectively handled by delegating probability to the Node struct such that subgraphs may have their own set of edge probabilities. A core graph that covers all existing nodes could ignore these fields if desired. 

These changes should enable effective creation of subgraphs and fix some of the problems with graph alignment blocks. Ideally the new Node struct will be very lightweight so data can be effectively shared between subgraphs.


craiglowe:
I am thinking about #1 above.  I think the idea is that the memory location of Data will act as an indirect identifier of a sequence and alignment column combination.  Such that if two Data memory addresses are equal if, and only if, they have the same sequence and are in the same alignment column.
I think this can already be done (more efficiently) with the current structure that has a pointer to a struct containing the sequence.  I think the only difference is whether it is a pointer to a ThreeBit struct or a Data struct.  I worry that creating the Data struct may actually make things harder to reuse since it ties fields like Annotation to the sequence, which may be different between a human graph and a chimp graph (different linear references).
I also worry a little bit about memory location being used as an indirect identifier.  Does it make more sense to store an explicit alignment column id in each node?  This would allow us to quickly analyze aligned graphs as well as separate out SNPs from indels in a single graph

DanielSnellings:
About your hesitancy to use the memory address as an identifier: Do you think identical nodes between different species should have the same underlying sequence in the node, or should the sequence be copied for each species?

PR #385, opened 2021-02-24
removing int64 from bed and other related structs

craiglowe:
I think people have heard this story a few times by now, but a couple years ago I thought it was good to make integer fields in structs int64.  Others and I no longer think this is the case.  In this pull request I am removing int64 from beds, mafs, and cigars.  I also needed to add a few lines of code that calculate the min/max integer values (has to figure out if machine is 32bit or 64bit), but this is only 3 lines of code in its own file in the numbers package.  I also stumbled on a file that was all comments, so I am proposing to remove it.


RavenYL (approved):

cfauci (approved):

rimangan (approved):
A huge step forward. Where are the consts for max values being used? I don't see those constants used yet, or is that still on the way?

PR #387, opened 2021-02-24
Update to multiFaToVcf

rimangan:
Some improvements to multiFaToVcf, which takes in a pairwise multiFa alignment and calls variants to a VCF file. Now uses channels to monitor output as the program is running and has options to retain substitutions only in the output and default behavior now does not write positions with Ns to the output.

### DIFF
--- a/cmd/multiFaToVcf/multiFaToVcf.go
+++ b/cmd/multiFaToVcf/multiFaToVcf.go
@@ -10,14 +10,13 @@ import (
 	"log"
 )
 
-func multiFaToVcf(inFile string, chr string, outFile string) {
+func multiFaToVcf(inFile string, chr string, outFile string, substitutionsOnly bool, retainN bool) {
 	f := fasta.Read(inFile)
-	v := convert.PairwiseFaToVcf(f, chr)
-	vcfWriter := fileio.EasyCreate(outFile)
-	header := vcf.NewHeader("nameFastaSample")
-	vcf.NewWriteHeader(vcfWriter, header)
-	vcf.WriteVcfToFileHandle(vcfWriter, v)
-	vcfWriter.Close()
+	out := fileio.EasyCreate(outFile)
+	header := vcf.NewHeader("")
+	vcf.NewWriteHeader(out, header)
+	convert.PairwiseFaToVcf(f, chr, *out, substitutionsOnly, retainN)
+	out.Close()
 }
 
 func usage() {
@@ -30,6 +29,8 @@ func usage() {
 }
 
 func main() {
+	var substitutionsOnly *bool = flag.Bool("substitutionsOnly", false, "Retain only substitutions in the output VCF.")
+	var retainN *bool = flag.Bool("retainN", false, "By default, multiFaToVcf does not retain positions with N in the alt or ref in the output VCF. This option overrides this behavior.")
 	var expectedNumArgs int = 3
 	flag.Usage = usage
 	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
@@ -44,5 +45,5 @@ func main() {
 	chromName := flag.Arg(1)
 	outFile := flag.Arg(2)
 
-	multiFaToVcf(inFile, chromName, outFile)
+	multiFaToVcf(inFile, chromName, outFile, *substitutionsOnly, *retainN)
 }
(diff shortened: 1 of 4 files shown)


rimangan:
Also because we're using channels and not holding VCF structs in memory this program also runs about 8 times faster than before

rimangan:
Important note: I ran this on genome size data and the headers wound up at the end of the file, not sure why. Does anyone have any fileio insight?

cmd/multiFaToVcf/multiFaToVcf.go
```diff
@@ -10,14 +10,13 @@ import (
...
-	v := convert.PairwiseFaToVcf(f, chr)
-	vcfWriter := fileio.EasyCreate(outFile)
-	header := vcf.NewHeader("nameFastaSample")
-	vcf.NewWriteHeader(vcfWriter, header)
-	vcf.WriteVcfToFileHandle(vcfWriter, v)
-	vcfWriter.Close()
+	out := fileio.EasyCreate(outFile)
+	header := vcf.NewHeader("")
+	vcf.NewWriteHeader(out, header)
+	convert.PairwiseFaToVcf(f, chr, *out, substitutionsOnly, retainN)
```
craiglowe, line 18: I think you will want out instead of *out here because I believe *fileio.EasyWriter implements the io.Writer interface, but fileio.EasyWriter does not (I am not sure about this and did not test it).  This also mirrors os.Create in the standard lib that returns a *File, which implements io.Writer.

convert/convert.go
```diff
@@ -226,15 +226,14 @@ func getWigChromIndex(s string, wigSlice []*wig.Wig) int {
 	return -1
 }
 
-//PairwiseFaToVcf takes in a pairwise multiFa alignment and returns Vcf entries for segregating sites with the first entry as the reference and the second fasta entry as the alt allele.
+//PairwiseFaToVcf takes in a pairwise multiFa alignment and writes Vcf entries for segregating sites with the first entry as the reference and the second fasta entry as the alt allele.
 //This will have to be done by chromosome, as a pairwise multiFa will only have two entries, thus containing one chromosome per file.
-func PairwiseFaToVcf(f []*fasta.Fasta, chr string) []*vcf.Vcf {
+func PairwiseFaToVcf(f []*fasta.Fasta, chr string, out fileio.EasyWriter, substitutionsOnly bool, retainN bool) {
```
craiglowe, line 231: Based on my previous comment, this would need a change from fileio.EasyWriter to *fileio.EasyWriter

convert/convert.go
```diff
@@ -247,7 +246,9 @@ func PairwiseFaToVcf(f []*fasta.Fasta, chr string) []*vcf.Vcf {
 		} else if f[0].Seq[i] != f[1].Seq[i] {
 			pastStart = true
 			if insertion { //catches the case where an insertion, now complete, is followed directly by a snp.
-				answer = append(answer, &vcf.Vcf{Chr: chr, Pos: fasta.AlnPosToRefPos(f[0], insertionAlnPos) + 1, Id: ".", Ref: dna.BaseToString(f[0].Seq[insertionAlnPos]), Alt: []string{dna.BasesToString(f[1].Seq[insertionAlnPos:i])}, Qual: 100.0, Filter: "PASS", Info: "."})
+				if !substitutionsOnly {
+					vcf.WriteVcf(out.File, &vcf.Vcf{Chr: chr, Pos: fasta.AlnPosToRefPos(f[0], insertionAlnPos) + 1, Id: ".", Ref: dna.BaseToString(f[0].Seq[insertionAlnPos]), Alt: []string{dna.BasesToString(f[1].Seq[insertionAlnPos:i])}, Qual: 100.0, Filter: "PASS", Info: ".", Format: []string{"."}})
```
craiglowe, line 250: This should be out instead of out.File.  By writing directly to the file, you "cut in line" to pass the buffer and the possible gzip compression.  This is also why the header appeared at the end.  It was waiting in the buffer for more to be written, while all the following data was written directly to the file, skipping the buffer.

craiglowe (commented):
Nice work!  I added a quick comment about the header at the end.

vcf/compare.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strings"
 )
 
+const debugMode = 0 //set debugMode to 1 to enable prints
```
cfauci, line 11: I really like this method, good job

cfauci (approved):

DanielSnellings (approved):

craiglowe (approved):
Great work!

PR #388, opened 2021-02-25
uncomment so remove header samToWig

cshoben:
Uncommented a line so we that samToWig will remove the headers in the input sam file. Necessary for downstream functionality.


DanielSnellings (approved):

rimangan (approved):

craiglowe (approved):
Good work tracking this down!  I am guessing it was commented out because someone put in a sam file with no header and ReadHeader fails if there is no header?  Maybe either ReadHeader can be updated to accept no header in the file or some other solution can allow both sam with and without a header

PR #389, opened 2021-02-27
Raven sam to wig fix

RavenYL:
Riley and I found a bug in samToWig.go with the placement of brackets. It has been fixed!

### DIFF
--- a/cmd/samToWig/samToWig.go
+++ b/cmd/samToWig/samToWig.go
@@ -45,7 +45,7 @@ func samToWig(samFileName string, reference string, outfile string, paired bool,
 	} else {
 		for aln, done = sam.NextAlignment(samFile); done != true; aln, done = sam.NextAlignment(samFile) {
 			currentBed = convert.SamToBed(aln)
-		}
+		//} this bracket shouldn't be here! checking currentBed != nil should happen within for loop. Otherwise check will happen AFTER the entire sam file was read, and only the last read will be appended to outBed, leading to mostly 0 in the bed and wig files
 		if currentBed != nil {
 			outBed = append(outBed, currentBed)
 		}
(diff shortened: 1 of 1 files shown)


cmd/samToWig/samToWig.go
```diff
@@ -45,15 +45,16 @@ func samToWig(samFileName string, reference string, outfile string, paired bool,
 	} else {
 		for aln, done = sam.NextAlignment(samFile); done != true; aln, done = sam.NextAlignment(samFile) {
 			currentBed = convert.SamToBed(aln)
+			//} this bracket shouldn't be here! checking currentBed != nil should happen within for loop. Otherwise check will happen AFTER the entire sam file was read, and only the last read will be appended to outBed, leading to mostly 0 in the bed and wig files
```
craiglowe, line 48: Good catch!  I think this same bug is above on line 41 inside the if statement where the user gave a fragment length

cmd/samToWig/samToWig.go
```diff
@@ -45,15 +45,16 @@ func samToWig(samFileName string, reference string, outfile string, paired bool,
...
-		}
-	}
 
-	outWig = convert.BedReadsToWig(outBed, ref)
-	log.Printf("Length of outWig: %d", len(outWig))
-	wig.Write(outfile, outWig)
+		outWig = convert.BedReadsToWig(outBed, ref)
+		log.Printf("Length of outWig: %d", len(outWig))
+		wig.Write(outfile, outWig)
+	}
```
craiglowe, line 57: I think this bracket should be moved 3 lines up so that the 3 lines above this line is not in the "else".  This way those three lines will run for both when the fragment length is given, or when that option is not used.

craiglowe (commented):
Good catch!  I think I notice a similar bug a few lines above.

craiglowe (approved):
Great catch!  Excellent to see these fixes.

rimangan (approved):
Great fix!

DanielSnellings (approved):

PR #390, opened 2021-02-28
Missing close in sequelOverlap

rimangan:
The cmd sequelOverlap creates an output file but does not call the close method on it. As a result, some files I was using sequelOverlap had the last line truncated. Quick fix, passes tests.


DanielSnellings (approved):
Nice catch

craiglowe (approved):
Good fix!

cfauci (approved):

PR #391, opened 2021-03-01
Removing Fasta Pointers

DanielSnellings:
Removes pointers to fastas throughout much of the codebase, with the exception of expandedTree which holds a *Fasta in the struct and the functions now switch between pointers and values where appropriate. Luckily most of the fasta-modifying functions already had a fasta return so this change was not too bad. Everything compiles and all tests pass. I did not see any cases where functions were modifying fasta length and did not already have a fasta return, so functionality should be unaffected. That said, I would appreciate if people could look through the changed functions and flag any where they think a change to a value instead of apointer may be disruptive.


craiglowe (approved):
Great work!  I scanned through looking for potential issues, but didn't see any.  Do we have any benchmarks where we can see potential speed changes?

DanielSnellings:
I have not run any before/after benchmarks. Though, I am open to suggestions about what functions would make a good test case. TBH fasta is probably the data type that will benefit from this the least, since we typically read-in an entire fasta file instead of streaming it. It should still get a small performance bump, but the real benefit will be when we remove pointers from vcf, fastq, sam, and bed.

cfauci (approved):

rimangan (approved):

PR #392, opened 2021-03-02
added go.sum file

craiglowe:
This file should also be in the repository now that we are adding compatibility with modules.


DanielSnellings:
Could you also add the go.mod and go.sum files to the .gitignore file?

craiglowe:
I am a little uncertain of all the details with implementing modules and the details of the go.mod and go.sum files.  I am including them in the repo based on this recommendation:
https://github.com/golang/go/wiki/Modules#releasing-modules-all-versions
I am thinking they should not be in the .gitignore since we should check them in?
I am more patching this together from advice here and there and don't have a unified game plan.  I was just looking to get the ball rolling since they are not required with go1.16

cfauci (approved):

rimangan (approved):

DanielSnellings (approved):

PR #394, opened 2021-03-04
Expanded tree test

cfauci:
Added test for ReadTree in expanded tree, updated error handling


expandedTree/expandedTree_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+)
+
+func TestReadTree(t *testing.T) {
+	var actualName = "root"
+	var aFound = false
+	tree, err := ReadTree("testdata/newick.txt", "testdata/fasta.fasta")
+	treeNodes := GetTree(tree)
+	if err != nil {
+		log.Fatal("Problem in ReadTree")
```
DanielSnellings, line 14: Test functions in gonomics have a special syntax for Fatal. Consider switching to using `t.Error()` or `t.Errorf()` instead of log.Fatal for test functions. The behavior is very similar but I believe t.Error() has some special functionality for advanced unit testing.

DanielSnellings (commented):
So for starters I love to see errors getting incorporated into functions, but there is a little nuance here that I think we should consider when evaluating whether or not to return an error. What seems to be happening is that there is really just 1 error being passed here: the error return from ReadNewick. I think Read Newick should definitely have an error return but what we don't want to happen is for our core functions to become littered with 'if err != nil' statements. I think this comes down to the difference between passing errors and handling errors. In this case, the ReadTree function is just passing the error from ReadNewick. But if you look at ReadNewick, the only possible error return is the result of a malformed Newick file, which (i believe) is unrecoverable in almost all cases. If it is true that there is no chance that the program can recover from this error, I think it would be appropriate to log.Fatal in the ReadTree function and just return a *ETree.

expandedTree/expandedTree_test.go
cfauci: This is exactly the kind of feedback I was hoping for. I haven't been exploring this part of go at all, thanks Dan!

cmd/reconstructSeq/reconstructSeq.go
```diff
@@ -10,7 +10,10 @@ import (
 )
 
 func ReconstructSeq(newickInput string, fastaInput string, outputFilename string) {
-	tree := expandedTree.ReadTree(newickInput, fastaInput)
+	tree, err := expandedTree.ReadTree(newickInput, fastaInput)
+	if err != nil {
```
craiglowe, line 14: You can use the one-line check of exception.FatalOnErr(err).  I am a little surprised that "%e" compiles.  It seems like this would convert your error into a floating point number in scientific notation.  Maybe "%s" is what you want?

cmd/simRecon/simRecon.go
```diff
@@ -13,7 +13,10 @@ import (
 
 //SimulateEvolve takes in a root fasta file, a newick tree, and gene structure genePred file for the fasta and returns a full simulated tree and a tree with sequence only at the leaves for reconstruction
 func SimulateEvolve(rootFastaFile string, treeFile string, gp string, simOutFile string, leafOutFile string) {
-	tree := expandedTree.ReadTree(treeFile, rootFastaFile)
+	tree, err := expandedTree.ReadTree(treeFile, rootFastaFile)
+	if err != nil {
```
craiglowe, line 17: same error check comment as above

cmd/simRecon/simRecon.go
```diff
@@ -31,7 +34,10 @@ func SimulateEvolve(rootFastaFile string, treeFile string, gp string, simOutFile
 
 //ReconstructSeq takes in a newick tree and leaf sequences and returns a reconstructed tree
 func ReconstructSeq(newickInput string, fastaInput string, outputFilename string) {
-	tree := expandedTree.ReadTree(newickInput, fastaInput)
+	tree, err := expandedTree.ReadTree(newickInput, fastaInput)
+	if err != nil {
```
craiglowe, line 38: same error check comment as above.  I will not leave any more comments, but I see this at least one more place in the code.

craiglowe (commented):
I had one small comment that appears a few times

expandedTree/expandedTree.go
```diff
@@ -26,12 +26,13 @@ type ETree struct {
...
-	if err == nil {
+	if err != nil {
+		return nil, err
 	}
 	AssignFastas(tr, fastasFilename)
-	return tr
+	return tr, nil
 }
 
 //read in tree from filename
```
cshoben, line 38: I've been trying to wrap my head around some of the code in expandedTree.go but it is old code so I can't necessarily point to it. I will leave my questions here and try to be explicit about what I'm talking about. 
1. For the ReadNewick() ---    Can someone explain in words what is happening at newline 50?
2. "        "  ---- for newline 53, will this only run if the two above "if" statements fail? (newlines 44 and 49) I'm expecting an "else" at this line, but without it my previous understanding if that the "if" statements would be checked, and once completed this "return" statement would always run. 
2. I'm trying to work through all the possible error statements that can come through the edited ReadTree() in expandedTree.go. 
      a. I see an error from the ReadNewick()
             i. This error could come from the os.Open()
            ii. This error could come from the "return nil, errors.New(..." defined on newline 53. 
            iii. Could this also have an error fed in from the "if !strings.HasPrefix()....." which would essentially come from the parseNewick()?
                 1. If yes, then an error can be from the parseNewick() statement if the line (tree) fails to start with "(" and end with ";"
                 2. Also, an error could come from the parseNewickHelper()?
                       i. For errors in this function..... I was reading about Newick tree format and code at newline 165 would fail for the tree I'm reading about? https://evolution.genetics.washington.edu/phylip/newicktree.html 
                           - does this code at 165 work for all trees? or just a very specific example tree?
                       ii. I see all the possible errors in the first batch of if/else statement. 
                      iii. we can get errors from the splitNameAndLength() which I believe has thwo possible error scenarios within itself.  
                      iv. But I do not understand newlines 185-192?

expandedTree/testdata/newick.txt
```diff
@@ -0,0 +1 @@
+(B:0.10,A:0.20)root:0.30;
```
cshoben, line 1: Will these input files ever be more than one line?

cmd/reconstructSeq/reconstructSeq.go
```diff
@@ -3,14 +3,18 @@ package main
...
 	"github.com/vertgenlab/gonomics/expandedTree"
 	"github.com/vertgenlab/gonomics/fasta"
 	"github.com/vertgenlab/gonomics/reconstruct"
 	"log"
 )
 
 func ReconstructSeq(newickInput string, fastaInput string, outputFilename string) {
-	tree := expandedTree.ReadTree(newickInput, fastaInput)
+	tree, err := expandedTree.ReadTree(newickInput, fastaInput)
+	if err != nil {
```
craiglowe, line 15: The exception.FatalOnErr(err) will do the "if err != nil {" for you, so you can remove this line and just have the single call to FatalOnErr.

craiglowe (commented):
Quick comment on error checking

cshoben (approved):

craiglowe (approved):
Great!

rimangan (approved):

PR #395, opened 2021-03-05
Moving Fasta to StablePrep

DanielSnellings:
This moves some of the fasta package to the StablePrep branch (see above that this is not merging into main) and updates a couple existing packages. The vision for StablePrep is to collect core packages/functions for an eventual stable release of gonomics. If anyone has some time to kill, I strongly recommend people pick a package to update, push the updates to main, then push them to the StablePrep branch.


craiglowe (commented):
I see this is pulling from a branch called fasta, but I was expecting it to be pulling from main.  Now that I think about it... I think it is hard to pull select files from one branch to another.  I think I might have lost track of the high-level flow of check-ins.

DanielSnellings:
I played with a few different ideas but I couldn't find a good way to do this. I can pull from main to StablePrep because they have different commit histories. I also tried creating a branch from main, deleting everything, then pulling from main to the new branch and that also did not work (says branch is already up to date with main). The way I made this PR was making a branch off from StablePrep and running `git checkout * *` to update all existing files in StablePrep with any changes to their main counterpart and running `touch 'filename'` to create any new files that I wanted to add to StablePrep, then running `git checkout * *` to update all of the newly created files as well.

cshoben (commented):
Looks good! I'm just unsure if you all decided that the pull request methods all check out?

cfauci (approved):

cshoben (approved):

rimangan (approved):

PR #397, opened 2021-03-07
A new plot emerges

rimangan:
Plotting tools for dissecting the remaining numerical stability bugs in selectionMCMC. Debug only, not for merging.

### DIFF
--- a/cmd/plotFunctions/plotFunctions.go
+++ b/cmd/plotFunctions/plotFunctions.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 )
 
-func plotFunctions(function string, functionArgs string, left float64, right float64, bins int, outFile string) {
+func plotContinuousFunctions(function string, functionArgs string, left float64, right float64, bins int, outFile string) {
 	if function == "AfsStationarity" {
 		words := strings.Split(functionArgs, ",")
 		if len(words) != 1 {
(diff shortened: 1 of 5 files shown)


cmd/plotFunctions/plotFunctions.go
```diff
@@ -58,31 +58,78 @@ func usage() {
...
 			"For Gamma: defined by two parameters (called alpha and beta). ex usage: plotFunctions Gamma 0.5,0.5 0.001 0.999 1000 gammaPlot.txt\n" +
 			"For Normal: defined by two parameters (called mu and sigma). ex usage: plotFunctions Normal 0,0.5 -1 1 1000 normalPlot.txt\n" +
+			"For ChooseN: defined by 1 parameter (called n). Prints the binomial coefficient in logSpace. ex usage: plotFunctions ChooseN 40 chooseNPlot.txt\n" +
 			"options:\n")
 	flag.PrintDefaults()
 }
 
 func main() {
-	var expectedNumArgs int = 6
+	var expectedNumArgs int = 1
```
craiglowe, line 72: Maybe just leave this uninitialized since we are not really using it, but setting it each time.  What happens when someone runs this with no arguments?  What does flag.Arg(0) on new line 77 evaluate to (give an error, nil)?  Maybe we should have the first "if" be a check for len(flag.Args()) being zero and then printing the usage?

popgen/mcmcSelection.go
```diff
@@ -165,6 +157,20 @@ func MetropolisHastings(data AFS, muZero float64, sigmaZero float64, iterations
 	}
 }
 
+func BuildBinomCache(allN []int) [][]float64 {
```
craiglowe, line 160: Didn't we already have this function somewhere?

popgen/plot.go
```diff
@@ -0,0 +1,36 @@
...
+	//DEBUG:fmt.Printf("Len BinomCache:%v. Entry at index n is of length: %v.\n", len(binomCache), len(binomCache[n]))
+
+	//a short header is written to the file
+	fmt.Fprintf(out, "Frequency\tProbability\n")
+	//for each possible allele frequency for a segregating site, we calculate the probability to the file
+	for i := 1; i < n; i++ {
+		fmt.Fprintf(out, "%v\t%e\n", i, AlleleFrequencyProbability(i, n, alpha, binomCache))
+	}
+
+	out.Close()
```
craiglowe, line 35: would be good to check the error here (it returns an error).

popgen/plot.go
```diff
@@ -0,0 +1,36 @@
...
+	out := fileio.EasyCreate(outFile)
+	allN := []int{n}
+	binomCache := BuildBinomCache(allN)
+
+	fmt.Fprintf(out, "Frequency\tF\n")
+
+	for i := 1; i < n; i++ {
+		fmt.Fprintf(out, "%v\t%e\n", i, AFSSampleDensity(n, i, alpha, binomCache))
+	}
+	out.Close()
```
craiglowe, line 18: would be good to check the error here

craiglowe (commented):
This will be really helpful!  Couple small comments

popgen/mcmcSelection.go
rimangan: This code previously existed, but wasn't wrapped up in a helper function. I needed this same code in another context, so I'm just repackaging here

DanielSnellings (approved):

craiglowe (approved):
This functionality will be really useful!

cfauci (approved):

PR #398, opened 2021-03-08
Gonomics Cmd Proof of Concept

DanielSnellings:
Idea on what a gonomics cmd might look like. Lacking the documentation aspect, but the core function is working.

### DIFF
--- a/cmd/gonomics/gonomics.go
+++ b/cmd/gonomics/gonomics.go
@@ -1,20 +1,22 @@
 package main
 
 import (
-	"bufio"
 	"flag"
 	"fmt"
 	"github.com/vertgenlab/gonomics/exception"
-	"io"
 	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
-	"sync"
+	"sort"
+	"strings"
 )
 
 func usage() {
-	flag.PrintDefaults()
+	fmt.Print(
+		"gonomics - A collection of builtin tools that use the gonomics core library.\n\n" +
+			"Usage: gonomics <command> [options]\n\n")
+	printCmdList()
 }
 
 func getBin() (path string, binExists map[string]bool) {
@@ -38,14 +40,14 @@ func getBin() (path string, binExists map[string]bool) {
 	}
 
 	if !binExists["gonomics"] { // check if gonomics is installed by checking for this binary
-		log.Fatalf("ERROR: gonomics binary were not found in %s\n " +
+		log.Fatalf("ERROR: gonomics binary were not found in %s\n "+
 			"Run 'go install ./...' in gonomics directory to compile binary", path)
 	}
 
 	return path, binExists
 }
 
-func getGonomicsFuncs() map[string]bool {
+func getGonomicsCmds() map[string]bool {
 	expectedPath := os.Getenv("GOPATH") + "/src/github.com/vertgenlab/gonomics/cmd"
 	cmds, err := ioutil.ReadDir(expectedPath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/gonomics/gonomics.go
```diff
@@ -0,0 +1,144 @@
...
+	var endFirstLineIdx int
+	var rawOutput []byte
+
+	fmt.Println("Commands:")
+	for _, cmdName := range cmds {
+		cmd := exec.Command(binPath + "/" + cmdName)
+		rawOutput, _ = cmd.Output()
+		endFirstLineIdx = strings.Index(string(rawOutput), "\n")
+
+		switch {
```
craiglowe, line 90: It is good that you handle all these cases, but I think the fact you need this means that our usage statements should be more consistent.

cmd/gonomics/gonomics.go
```diff
@@ -0,0 +1,144 @@
...
+		cmds = append(cmds, key)
+	}
+	sort.Slice(cmds, func(i, j int) bool { return cmds[i] < cmds[j] })
+
+	var endFirstLineIdx int
+	var rawOutput []byte
+
+	fmt.Println("Commands:")
+	for _, cmdName := range cmds {
+		cmd := exec.Command(binPath + "/" + cmdName)
```
craiglowe, line 86: Is this slow?  Probably not, but another option would be to parse the source (might also be slow).  Another option for the future is that we could have a github action write this program... then it would not be live, so maybe it is better to have this main gonomics program reflect the current state of all the programs and not what their state was at the last check-in/merge.

cmd/gonomics/gonomics.go
```diff
@@ -0,0 +1,144 @@
...
+	switch { // Error checks. verbose for clarity
+	case isGonomicsFunc[cmdCalled] && binExists[cmdCalled]:
+		// proceed in main function
+
+	case isGonomicsFunc[cmdCalled] && !binExists[cmdCalled]:
+		log.Fatalf("ERROR: binary for %s was not found. "+
+			"Please run 'go install ./...' in the gonomics directory.", cmdCalled)
+
+	case !isGonomicsFunc[cmdCalled] && binExists[cmdCalled]:
+		log.Fatalf("ERROR: %s is not a gonomics function", cmdCalled)
```
craiglowe, line 130: Maybe instead of "function" we should use command / executable / program?

craiglowe (approved):
Looks great!  Cool to see this!  I approved, but had some minor comments to think about.

cmd/gonomics/gonomics.go
DanielSnellings: Yes, this definitely highlights some inconsistencies in our usage statements. I would be good to standardize eventually. Maybe during the gonomics cleanup day

cmd/gonomics/gonomics.go
DanielSnellings: Yes, switched to cmd.

cmd/gonomics/gonomics.go
```diff
@@ -0,0 +1,154 @@
...
+
+	case isGonomicsCmd[cmdCalled] && !binExists[cmdCalled]:
+		log.Fatalf("ERROR: binary for %s was not found. "+
+			"Please run 'go install ./...' in the gonomics directory.", cmdCalled)
+
+	case !isGonomicsCmd[cmdCalled] && binExists[cmdCalled]:
+		log.Fatalf("ERROR: %s is not a gonomics function", cmdCalled)
+
+	default:
+		log.Fatalf("ERROR: %s does not exist, and is not a gonomics function", cmdCalled)
```
craiglowe, line 143: I see you changed "function" to cmd/command in the code variables, but I think it is still in the text printed for users, like on this line.

cmd/gonomics/gonomics.go
```diff
@@ -0,0 +1,154 @@
...
+
+	switch { // Find binary location. Preference is GOBIN > GOPATH > Default go install location
+	case os.Getenv("GOBIN") != "":
+		path = os.Getenv("GOBIN")
+
+	case os.Getenv("GOPATH") != "":
+		path = os.Getenv("GOPATH") + "/bin" // default to $GOPATH/bin
+
+	default:
+		path = "/usr/local/go/bin" // default install directory
```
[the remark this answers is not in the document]
DanielSnellings: Nice catch, changed.

cmd/gonomics/gonomics.go
DanielSnellings: nice catch

craiglowe (approved):
Cool to have this!

rimangan (approved):

cfauci (approved):

PR #399, opened 2021-03-12
Deletion Effect Prediction

DanielSnellings:
Effect prediction for deletions in the gene package. I also split the Deletion function into smaller functions so things will be a bit easier to read.

### DIFF
--- a/gene/delete.go
+++ b/gene/delete.go
@@ -393,6 +393,7 @@ func deleteEffectPrediction(g *Gene, deletedCodingBases int, answer EffectPredic
 		return answer, err
 
 	} else { // Deletion affects CDS
+		answer.Consequence = InFrameDeletion // until otherwise noted
 		answer.AaPos = answer.CdnaPos / 3
 		startFrame := answer.CdnaPos % 3                          // Where the start of the deletion is relative to codon boundaries
 		numRemovedAA := (deletedCodingBases + startFrame + 2) / 3 // the +2 is to account for potential partial codon at del end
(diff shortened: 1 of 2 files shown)


gene/delete.go
```diff
@@ -339,3 +206,255 @@ func deleteStable(seq *[]dna.Base, delStart int, delEnd int) error {
...
+
+	// Update startPos if necessary
+	if genomeIndexStartPos == 0 {
+		g.startPos += genomeIndexEndPos
+	}
+}
+
+// deleteUpdateCds updates the coding sequence upon deletion
+func deleteUpdateCds(g *Gene, genomeIndexStartPos int, genomeIndexEndPos int) (deletedCodingBases int, err error) {
+	var cdsIndexesToDelete []int
```
craiglowe, line 224: Indexes -> Indices?

gene/delete.go
```diff
@@ -339,3 +206,255 @@ func deleteStable(seq *[]dna.Base, delStart int, delEnd int) error {
...
+	var codingDelStart, codingDelEnd int = -1, -1
+	deletionLen := genomeIndexEndPos - genomeIndexStartPos
+	for i := 0; i < len(g.cdsStarts); i++ {
+
+		switch {
+		// if cds is before deletion
+		case genomeIndexStartPos > g.cdsEnds[i]:
```
craiglowe, line 231: should > be >=?  I am having trouble remembering the numbering system for these variables and I don't see it in the struct declaration.  Is this interval closed-closed or closed-open?  I think we should try to used closed-open if we have control over the struct/file format so that we match golang slices and most numbering in genomics.  I am thinking that the genomeIndex variables are closed-open from the length calc above.

gene/delete.go
```diff
@@ -339,3 +206,255 @@ func deleteStable(seq *[]dna.Base, delStart int, delEnd int) error {
...
+			codingDelStart = int(g.featureArray[g.cdsEnds[i]]) + 1
+
+		// If cds exon is fully deleted
+		case genomeIndexStartPos <= g.cdsStarts[i] && genomeIndexEndPos > g.cdsEnds[i]:
+			// mark cds for deletion
+			cdsIndexesToDelete = append(cdsIndexesToDelete, i)
```
craiglowe, line 237: This is the only place I see cdsIndexesToDelete updated, but it seems like we would also need to remove some indices if we have a deletion that starts in one cds and then ends in another.

craiglowe (commented):
Really like the longer functions being broken up into smaller functions.  Looks great!
I worked my way through the first half and had some questions.

cfauci (approved):

craiglowe (approved):
Great work!

cshoben (approved):

rimangan (approved):

PR #404, opened 2021-03-19
Piecewise Integration of the Stationarity F Function

rimangan:
The function AFSSampleDensity, located in popgen/stationarity.go, is the product of the stationarity and binomial distributions integrated over p, the allele frequency. As the binomial distribution for very small values approaches a very narrow peak, integration across this range often misses the density of the function producing aberrant results. However, by calculating this in two parts (0 to i/n and from i/n to 1, where i/n is the maximum of the binomial distribution), the resulting plots of the Allele Frequency Probability Mass Function are much closer to theoretical expectations. Still waiting to see how this affects MCMC traces as the hardac is super busy tonight so this remains a draft for now until I can assess the downstream consequences of this change.


cfauci (approved):

craiglowe (approved):
Great to see this merged in to main!
Might be interesting to do some of your plots with that 1e-8 changing to be a smaller or larger number.  Smaller numbers (e.g. 1e-10) should be more accurate, but larger numbers (e.g. 1e-6) should run faster.  I think 1e-8 was just an initial guess if I remember correctly.

DanielSnellings (approved):

cshoben (approved):

PR #444, opened 2021-03-22
update to axt package

craiglowe:
This is an update to the axt package.  It does complete all the work that is needed there, but takes a step in the right direction.  Reduces some use of pointers, adds comments, removes old and unused code.
This also presents (what I hope is) a simpler version of the ReadToChan and GoReadToChan functions.  I would be interested to hear everyone's thoughts.  It reduces some of the complexity by removing a third Go routine that waits to close the file.  At one point we had the idea to make these functions more versatile (e.g. could be used to read multiple files at once), but I don't think we are using them in that way and probably will not be in the near future.  I think this presents a streamlined version of functions that cover our use cases.

### DIFF
--- a/axt/axt.go
+++ b/axt/axt.go
@@ -1,14 +1,15 @@
-//Package axt provides the struct and functions that operate on axt alignment formats.
+// Package axt provides the struct and functions that operate on alignments in Axt format.
 package axt
 
 import (
 	"fmt"
 	"github.com/vertgenlab/gonomics/common"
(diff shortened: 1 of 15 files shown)


DanielSnellings (approved):
These changes look good to me. I like the simplified ReadToChan functions. It does preclude them from being used when reading multiple files into the same channel, however I agree that this is an uncommon usage and should probably be implemented as custom functions.

rimangan (approved):

cfauci (approved):

PR #447, opened 2021-03-25
Dna equal

cfauci:
I needed a simple bool return for base comparisons, this just checks two bases against each other to see if they have equal value.


DanielSnellings (commented):
This function could just be `return a == b`. When you use the operator `==` the whole expression becomes a bool. You could also do this in other packages without a dna.Equal function (e.g. `if seq[i] == dna.A { *do some stuff*}`)

PR #448, opened 2021-03-25
genePred clean up

cfauci:
removed pointers and added GoDocs

### DIFF
--- a/genePred/compare.go
+++ b/genePred/compare.go
@@ -5,7 +5,7 @@ import (
 	"strings"
 )
 
-func AllAreEqual(a []*GenePred, b []*GenePred) bool {
+func AllAreEqual(a []GenePred, b []GenePred) bool {
 	if len(a) != len(b) {
 		return false
 	}
@@ -17,7 +17,7 @@ func AllAreEqual(a []*GenePred, b []*GenePred) bool {
 	return true
 }
 
-func Equal(a *GenePred, b *GenePred) bool {
+func Equal(a GenePred, b GenePred) bool {
 	if strings.Compare(a.Id, b.Id) != 0 {
 		log.Print("1")
 		return false
--- a/genePred/compare.go
+++ b/genePred/compare.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 )
 
+// AllAreEqual examines if two genePred files are the same entry by entry.
 func AllAreEqual(a []GenePred, b []GenePred) (bool, []int) {
 	var fields []int
 	var equal bool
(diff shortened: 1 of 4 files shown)


rimangan (dismissed):

genePred/compare.go
```diff
@@ -17,7 +17,7 @@ func AllAreEqual(a []*GenePred, b []*GenePred) bool {
 	return true
 }
 
-func Equal(a *GenePred, b *GenePred) bool {
+func Equal(a GenePred, b GenePred) bool {
 	if strings.Compare(a.Id, b.Id) != 0 {
 		log.Print("1")
```
craiglowe, line 22: These look like they are leftover from debugging?  Or maybe they are meant as diagnostics for when they don't match?  In that second case, maybe this should be in the _test.go file since it is only meant to be used in testing?

genePred/genePred.go
```diff
@@ -119,8 +125,9 @@ func StringToIntSlice(text string) []int {
 	return answer
 }
 
-//TODO: May not work with - strand transcripts
-func CalcExonFrame(gene *GenePred) []int {
+// CaclExonFame calculates the frame of each exon in a genePred based on the information in the fields for exonStarts, exonStops and cdsStarts
+//May not work with - strand transcripts
```
craiglowe, line 129: I would leave the TODO here, or make an issue for this.

craiglowe (approved):
Great!  I had a few minor comments.

genePred/compare.go
cfauci: my thought was that this could be used for comparison purposes in general, so if you wanted to compare entries directly. It's not necessarily useful right at this moment, but i think it could be later on. I think I'm going to return a []int with the bool so that this can be used downstream and replace the prints. Maybe a 1 in every position where the fields dont match and a 0 in the positions they do?

genePred/genePred.go
cfauci: I'll make a TODO and maybe an issue as well

genePred/compare.go
```diff
@@ -1,61 +1,65 @@
...
+		return equal, fields
 	}
-	return true
+	return equal, fields
 }
 
-func Equal(a *GenePred, b *GenePred) bool {
+func Equal(a GenePred, b GenePred) (bool, []int) {
+	var isEqual = true
+	var fields = make([]int, 9)
```
craiglowe, line 23: I see that down below you access fields[9].  Because the slice starts at zero, you need it to have a length of 10 to be able to access fields[9].

genePred/compare.go
```diff
@@ -1,61 +1,65 @@
 package genePred
 
 import (
-	"log"
 	"strings"
 )
 
-func AllAreEqual(a []*GenePred, b []*GenePred) bool {
+func AllAreEqual(a []GenePred, b []GenePred) (bool, []int) {
```
craiglowe, line 7: Returning the []int to give the programmer more detailed information on what didn't match is an interesting idea, but because it is non-standard, I think this function and Equal need to have comments that explain to the user what they are getting back.  It is pretty clear what they get from the code, but the goal is for the documentation to be able to stand on its own without having the person need to look at the actual code.
Is this more for debugging so that the human can know what didn't match (but was anticipated to match)?  Or is the idea that you have a program that needs to know what didn't match and then will do something different if the name didn't match or if the start position didn't match?  If the first case, maybe it makes sense to put this in a _test.go file with error messages that describe which comparison failed?

genePred/compare.go
craiglowe: Is the vision that this function will be called many times and the records will usually match?  If so, we are going to have a constant allocation and non-use of this slice.  Another option would be to declare an empty / nil slice here: "var fields []int" and then use append() to add all the non-matching fields to it (when needed).  For example, instead of "fields[2] = 1", you would have fields = append(fields, 2)".  You would then return something like "(2,4,5)" instead of "(0, 0, 1, 0, 1, 1)".  Another thing to think about is if []bool would be a better fit than []int if you like the current style since it seems like the int is really holding a 0/1 false/true

craiglowe (commented):
Great to see this package getting polished!  I think I found one bug in the slice length allocation and had a few small comments.

genePred/compare.go
cfauci: my idea is the second, where in a real-world scenario you'd want to know if two entries are different.

genePred/compare.go
cfauci: i went with the idea that the field that had the error is stored as an int in the slice. Thank you for pointing out these options!

cfauci (commented):
made some changes and responded to comments

genePred/compare.go
```diff
@@ -1,61 +1,68 @@
...
 	if len(a) != len(b) {
-		return false
+		return false, fields
 	}
 	for i := 0; i < len(a); i++ {
-		if !Equal(a[i], b[i]) {
-			return false
-		}
+		equal, fields = Equal(a[i], b[i])
+		return equal, fields
```
craiglowe, line 17: I think you want to keep this return statement wrapped in an "if" check, otherwise you are only going to be returns the answer from the first time through the loop (i.e. a[0] vs b[0]).

genePred/genePred_test.go
```diff
@@ -27,8 +27,9 @@ var ReadTests = []struct {
 func TestRead(t *testing.T) {
 	for _, test := range ReadTests {
 		actual := Read(test.name)
-		if !AllAreEqual(test.data, actual) {
-			t.Errorf("The %s file was not read correctly.", test.name)
+		equal, fields := AllAreEqual(test.data, actual)
```
craiglowe, line 30: Do we have a check where we compare two lists of GenePreds where one of them are not equal and "not equal" is the correct answer?

craiglowe (commented):
I think there is a bug where the AllEqual returns the state of the first comparison.

rimangan (approved):

cshoben (approved):

DanielSnellings (approved):

craiglowe (approved):
Great!

rimangan (approved):

PR #449, opened 2021-03-26
Discussion of limitation for new ReadToChan functions

DanielSnellings:
I like the simplified version of ReadToChan that Craig proposed for AXT, but it does present a problem with how we handle header returns. This is not a problem with AXT, but with files like sam, we would ideally be able to get header info when we read the file to a channel. As the new functions move file opening/closure to the ReadToChan function, it becomes challenging to do this with a return value. I would like to propose we modify to the ReadToChan standard to enables header info. One option as implemented in this PR is to revert to the old function where GoReadToChan handles the file and we can read a header, then return the header and a channel of data. The other option (which I actually like more) would be to add in a channel input to ReadToChan from which we can send the header info the receiving function. This way the standard that Craig made will not change, but filetypes that have a header will have a second channel return used for retrieving header info. Making the header come from a channel may be somewhat nonintuitive as it is a single element rather than many elements, but I think a header channel could be a nice fix for this problem.

This PR was mostly made for discussion. It can be approved, modified, or closed depending on how we decide to handle this problem.

### DIFF
--- a/axt/axt.go
+++ b/axt/axt.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log"
 	"strings"
+	"sync"
 )
 
 // Axt struct: Naming convention is hard here because UCSC website does not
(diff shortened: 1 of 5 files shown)


axt/axt.go
```diff
@@ -44,29 +45,39 @@ func Read(filename string) []Axt {
...
+	var err error
+	var wg sync.WaitGroup
+	file = fileio.EasyOpen(filename)
+
 	data := make(chan Axt, 1000)
-	go ReadToChan(filename, data)
+
+	wg.Add(1)
+	go func() { // spawn a goroutine that closes the file and channel when ReadToChan is finished
+		wg.Wait()
```
craiglowe, line 74: I don't think this added complexity of having a go routine wait around for a signal to then close the file and close the channel buys us any advantages.  I think that when GoReadToChan finishes reading the file, it should just close the file and close the channel, instead of signalling to a waiting go routine to close the file and close the channel (we already know we are at the end of the file).

cmd/axtSam/axtSam.go
```diff
@@ -52,15 +52,21 @@ func main() {
...
 	//setting up channels and wait groups
 	data, results := make(chan axt.Axt, 824), make(chan *sam.SamAln, 824)
 	var working, writingJob sync.WaitGroup
 
-	go axt.ReadToChan(axtfile, data)
+	var wg sync.WaitGroup
+	go axt.ReadToChan(reader, data, &wg)
+
+	go func() {
+		wg.Wait()
```
craiglowe, line 67: This may actually be an example of the added complexity being difficult where we have now removed the file close from ReadToChan, but it is also not included in the waiting go routine.

cmd/gsw/axtHelper.go
```diff
@@ -33,8 +34,16 @@ func convertAxt(axtFile, format, targetFa, output string) {
...
 func goChannelAxtVcf(axtFile string) <-chan *vcf.Vcf {
+	file := fileio.EasyOpen(axtFile)
+	var wg sync.WaitGroup
 	axtChannel := make(chan axt.Axt, 2408)
-	go axt.ReadToChan(axtFile, axtChannel)
+	wg.Add(1)
+	go axt.ReadToChan(file, axtChannel, &wg)
+
+	go func() {
+		wg.Wait()
```
craiglowe, line 44: I think this might be a second example of the waiting go routine not closing the file.

craiglowe (commented):
I think you bring up a good point that many file formats need to (quickly) give back a header as well as a channel of data.  For this reason, you make a good point that ReadToChan should revert to taking an EasyReader, instead of a filename.  I am uncertain about bringing back the added complexity of the additional go routine that waits to close the channel (and now file as well).  I don't think this added complexity buys us much functionality, so I would continue to have GoReadToChan close the file when it is done and then close the channel.
What do you think about this as the template:
GoReadToChan:
1) Makes a channel for the data type
2) Open the file so that it has an EasyReader
3) Calls a function that takes the EasyReader and gives back a header (optional step)
4) Calls "go ReadToChan" with the EasyReader and the data channel
5) Returns the header (can do this without a channel) and the data channel

axt/axt.go
DanielSnellings: This is a fair point. I considered letting ReadToChan close the file, but it seemed a bit messy to have ReadToChan silently close an already opened file. It seemed nicer to have open/close within the same function, but maybe that is not worth the complexity of another goroutine. As a last proposal I pushed another couple files that lay out how a header channel could work and we could have the best of both worlds. If that doesn't hold up to scrutiny, I will change it so ReadToChan closes the file and make it clear in the godoc that this happens.

craiglowe (commented):
I like the template shown in readToChanOption2.go

PR #454, opened 2021-03-29
Recon acc for bases

cfauci:
Created a func that calculates the percentage of correct base calls in reconstruction for each position in a codon.


reconstruct/reconstruct.go
```diff
@@ -101,9 +105,95 @@ func ReconAccuracy(simFilename string, reconFilename string, leavesOnlyFile stri
...
+
+// ReconAccuracyByBase will run if the option given to ReconAccuracy = 1. This function calculates the percentage of first, second and third bases in codons that were correct.
+func ReconAccuracyByBase(simFilename string, reconFilename string, gpFilename string) map[string][]float64 {
+	sim := fasta.Read(simFilename)
+	recon := fasta.Read(reconFilename)
+	genes := genePred.Read(gpFilename)
+	answer := make(map[string][]float64)
+
+	for s := 0; s < len(sim); s++ {
+		for r := 0; r < len(recon); r++ {
```
craiglowe, line 119: I believe there is a "fasta map" in the fasta package which will let you lookup a sequence by its name.  This will also let you realize when that sequence is not found.  A possible update would be to read "recon" into a map so that you can just lookup sim[s].Name in that map.  This will let you removed the "r" loop and also give an error if it is not found.

cmd/simRecon/simRecon.go
```diff
@@ -52,32 +52,52 @@ func ReconstructSeq(newickInput string, fastaInput string, outputFilename string
...
+		"SimRecon takes in a root fasta file, a genePred, and a tree. It simulates evolution along the tree, performs reconstruction based on the leaf nodes of the tree and then calculates accuracy at each node.\n" +
+			"simRecon [-option] <rootFile.fasta> <treeFile.txt> <genePred.gp> <simOutFile.fasta> <leafOutFile.fasta> <reconOutFile.fasta> <accuracyOutFile.txt>\n" +
 			"options:\n")
 	flag.PrintDefaults()
 }
 
 func main() {
 	var expectedNumArgs = 7
+	var option = flag.Int("option", 0, "Specify to 1 to return an accuracy breakdown by base position in codons. Default value is 0, which returns normal output.")
+	var baseAccFile = flag.String("baseAccFile", "", "Specify a filename for the output of accuracy by position of base in codon (if option = 1).")
```
craiglowe, line 100: Do you need both baseAccFile and option?  It seems like you can't have one without the other and a test of baseAccFile against the empty string could act like checking if option equals 0/1.

cmd/simRecon/simRecon.go
```diff
@@ -52,32 +52,52 @@ func ReconstructSeq(newickInput string, fastaInput string, outputFilename string
...
 
-	answer := reconstruct.ReconAccuracy(simOutFile, reconOutFile, leafOutFile, gp)
+	answer, byBaseAnswer := reconstruct.ReconAccuracy(simOutFile, reconOutFile, leafOutFile, gp, option)
 	out := fileio.EasyCreate(accuracyOutFile)
 	defer out.Close()
 
 	for name, accuracy := range answer {
 		fmt.Fprintf(out, "%s\t%f\n", name, accuracy)
 	}
+	if option == 1 && baseAccFile != "" {
```
craiglowe, line 69: Maybe option can be removed as a variable and you only need to test if baseAccFile != ""?

craiglowe (commented):
Great to see this functionality.  I had a few comments

cmd/simRecon/simRecon.go
DanielSnellings: If you want to keep option, you may also consider making option a flag.Bool since the only acceptable values are 0 and 1.

reconstruct/reconstruct.go
```diff
@@ -101,9 +105,95 @@ func ReconAccuracy(simFilename string, reconFilename string, leavesOnlyFile stri
...
+			var percentage1 float64
+			var percentage2 float64
+			var percentage3 float64
```
DanielSnellings, line 122: Consider merging these variable declarations to the same line.
e.g. `var percentage1, percentage2, percentage3 float64`

reconstruct/reconstruct.go
```diff
@@ -101,9 +105,95 @@ func ReconAccuracy(simFilename string, reconFilename string, leavesOnlyFile stri
...
+			var total1 float64
+			var total2 float64
+			var total3 float64
+			if sim[s].Name == recon[r].Name {
+				for i := 0; i < len(sim[s].Seq); i++ {
+					for g := 0; g < len(genes); g++ {
+						inExon, exon := simulate.CheckExon(genes[g], i)
+						if inExon {
+							loc := findLocationInCodon(genes[g], exon, i)
+							if loc == 1 {
```
DanielSnellings, line 135: Feel free to disagree, but this is the exact situation switch cases were designed for. Consider changing to 
```
switch loc {
case 1: // some work
case 2: // some work
case 3: // some work
}
```

reconstruct/reconstruct.go
```diff
@@ -101,9 +105,95 @@ func ReconAccuracy(simFilename string, reconFilename string, leavesOnlyFile stri
...
+				answer[sim[s].Name] = append(answer[sim[s].Name], 100-percentage1)
+				answer[sim[s].Name] = append(answer[sim[s].Name], 100-percentage2)
+				answer[sim[s].Name] = append(answer[sim[s].Name], 100-percentage3)
+			}
+		}
+	}
 	return answer
 }
 
+// findLocationInCodon returns what position in a codon a given base inhabits (first, second or third).
```
DanielSnellings, line 166: Consider making this zero-based. This is the same as frame in gtf format which uses zero-based.

reconstruct/reconstruct.go
```diff
@@ -101,9 +105,95 @@ func ReconAccuracy(simFilename string, reconFilename string, leavesOnlyFile stri
...
+		location := position - gene.ExonStarts[exon] + 2
+		if location%3 != 0 {
+			positionInCodon = location % 3
+		} else {
+			positionInCodon = 3
+		}
+	} else {
+		log.Panic("Exon Frame is not 0, 1 or 2")
+	}
+	return positionInCodon
```
DanielSnellings, line 194: I am getting confused here. Is location not zero-based?

cmd/simRecon/simRecon.go
```diff
@@ -52,32 +52,52 @@ func ReconstructSeq(newickInput string, fastaInput string, outputFilename string
 }
 
 //SimRecon simulates evolution, performs reconstruction, and then evaluates the accuracy of the reconstruction
-func SimRecon(rootFastaFile string, treeFile string, gp string, simOutFile string, leafOutFile string, reconOutFile string, accuracyOutFile string) {
+func SimRecon(rootFastaFile string, treeFile string, gp string, simOutFile string, leafOutFile string, reconOutFile string, accuracyOutFile string, option int, baseAccFile string) {
```
cshoben, line 55: To make sure I'm reading this correctly, there is no "output" other than what's printed to the screen?

reconstruct/reconstruct_test.go
```diff
@@ -36,13 +36,30 @@ func Test_reconstruct(t *testing.T) {
...
-	fileio.EasyRemove("simOut.fasta")
 	fileio.EasyRemove("RandGeneOutput.fasta")
+}
+
+func TestReconAccuracyByBase(t *testing.T) {
+	_, baseAccuracy := ReconAccuracy("simOut.fasta", "reconOut.fasta", "leavesOnly.fasta", "testdata/genePred.gp", 1)
+
+	for name, data := range baseAccuracy {
+		for d := range data {
+			if d == 0 {
```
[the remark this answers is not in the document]
DanielSnellings: Agreed, this has confused me as I have gone through. I think its OK to use 1-base (especially if that is the file standard), but I think it should be explicitly stated in the godoc which inputs are 1-based

cmd/simRecon/simRecon.go
cshoben: What function would output this "baseAccFile"? Is this the reconstruct.ReconAccuracy output "accBases"? Or maybe you can tell me more about how this file is made?

cmd/simRecon/simRecon.go
cshoben: So maybe baseAccFile is not a file, but simply a name we want a future file to be? And in SimRecon, at the moment of input it is still a name and not actually a file? But then we never output a file. So is baseAccFile ever really a file? I'm likely missing something so I figured I'd write out my thoughts and maybe you can help me understand what I'm failing to comprehend. Thanks!

cshoben (commented):
Looks good. I have some comments in reference to me interpreting the code. I was trying to follow Craig but I think I was struggling to grasp some upstream understanding.

cfauci:
I still need to update the test file for reconstruct.

cmd/simRecon/simRecon.go
cshoben: How do you know the location that the new file will be written into? Is it based on the pwd where you execute the function?

reconstruct/reconstruct.go
```diff
@@ -101,9 +100,72 @@ func ReconAccuracy(simFilename string, reconFilename string, leavesOnlyFile stri
...
+	recon := fasta.ToMap(rec)
+	genes := genePred.Read(gpFilename)
+	answer := make(map[string][]float64)
+
+	for s := 0; s < len(sim); s++ {
+		var percentage0, percentage1, percentage2 float64
+		var mistakes0, mistakes1, mistakes2 float64
+		var total0, total1, total2 float64
+		rSeq, ok := recon[sim[s].Name]
+		if ok {
```
craiglowe, line 119: Could we add an "else" to this "if" that would give an error?  Seems like we should always be able to find the seq in recon, right?

reconstruct/reconstruct.go
```diff
@@ -101,9 +100,72 @@ func ReconAccuracy(simFilename string, reconFilename string, leavesOnlyFile stri
...
+	}
 	return answer
 }
 
+// findLocationInCodon returns what position in a codon a given base inhabits (0, 1, 2)
+func calcLocationInCodon(gene genePred.GenePred, exon int, position int) int {
+	var positionInCodon int
+
+	location := position - gene.ExonStarts[exon] + gene.ExonFrames[exon]
+	if location%3 != 0 {
```
craiglowe, line 161: Do we need this "if/else"?  It seems like going through this "if" is the same as just having the line:
positionInCodon = location % 3

reconstruct/reconstruct_test.go
```diff
@@ -31,18 +33,59 @@ func Test_reconstruct(t *testing.T) {
...
-		accuracyData := ReconAccuracy("simOut.fasta", "reconOut.fasta", "leavesOnly.fasta", "testdata/genePred.gp")
+		accuracyData, _ = ReconAccuracy("simOut.fasta", "reconOut.fasta", "leavesOnly.fasta", "testdata/genePred.gp", false)
 		for name, accuracy := range accuracyData {
 			log.Printf("%s %f \n", name, accuracy)
 		}
 	}
-	fileio.EasyRemove("reconOut.fasta")
-	fileio.EasyRemove("simOut.fasta")
+
+	if accuracyData[leaves[0].Name] != 100 {
```
craiglowe, line 47: This looks like an "if" with nothing inside it (so it could just be removed).  Is there supposed to be something in it?

reconstruct/reconstruct_test.go
```diff
@@ -31,18 +33,59 @@ func Test_reconstruct(t *testing.T) {
...
+				log.Printf("%s Third Base %f \n", name, data[d])
+			}
+		}
+	}
+
+	data, ok := baseAccuracy["A"]
+	if !ok {
+		t.Error("node A not found in baseAccuracy data, check tree input.")
+	} else {
+		if data[0] < 97.0 {
```
craiglowe, line 75: Can we make this a tighter check to >97.9 and < 98?

craiglowe (commented):
Added a few minor comments

cmd/simRecon/simRecon.go
```diff
@@ -51,33 +51,59 @@ func ReconstructSeq(newickInput string, fastaInput string, outputFilename string
...
 	//ReconAccuracy calculates the total number of incorrectly reconstructed base pairs in a tree and returns a percentage of correct base calls
 	SimulateEvolve(rootFastaFile, treeFile, gp, simOutFile, leafOutFile)
 	ReconstructSeq(treeFile, leafOutFile, reconOutFile)
+	var calcBaseAcc = false
+	if baseAccFile != "" {
+		calcBaseAcc = true
+	}
 
-	answer := reconstruct.ReconAccuracy(simOutFile, reconOutFile, leafOutFile, gp)
+	answer, byBaseAnswer := reconstruct.ReconAccuracy(simOutFile, reconOutFile, leafOutFile, gp, calcBaseAcc)
```
cshoben, line 69: Is this byBaseAnswer the accBases output from ReconAccuracy?

cshoben (commented):
Another question as I comprehend the code :)

cmd/simRecon/simRecon.go
```diff
@@ -51,33 +51,59 @@ func ReconstructSeq(newickInput string, fastaInput string, outputFilename string
...
 	out := fileio.EasyCreate(accuracyOutFile)
 	defer out.Close()
 
 	for name, accuracy := range answer {
 		fmt.Fprintf(out, "%s\t%f\n", name, accuracy)
 	}
+	if baseAccFile != "" {
+		baseAccOut := fileio.EasyCreate(baseAccFile)
+		defer baseAccOut.Close()
```
DanielSnellings, line 78: Should probably do an explicit close at the end of the function and check for errors. EasyWriters have some pretty serious errors.

cmd/simRecon/simRecon.go
```diff
@@ -51,33 +51,59 @@ func ReconstructSeq(newickInput string, fastaInput string, outputFilename string
...
+		for species, baseAcc := range byBaseAnswer {
+			firstBase := species + " First Base"
+			secondBase := species + " Second Base"
+			thirdBase := species + " Third Base"
+			for b := range baseAcc {
+				if b == 0 {
+					fmt.Fprintf(baseAccOut, "%s\t%f\n", firstBase, baseAcc[b])
```
DanielSnellings, line 85: Should be handling errors here and the following calls to Fprintf

craiglowe (dismissed):
Great to see this functionality!  Had one comment in the testing file.

DanielSnellings (approved):

rimangan (dismissed):

DanielSnellings (approved):

rimangan (approved):

cshoben (approved):

PR #455, opened 2021-03-29
Sam updates

DanielSnellings:
This is a rather big one. Many changes, some significant changes include changing types in the Sam struct to match file specifications, removed pointers where appropriate, parse some of the header text to make the header more useful, rename sam.Aln to sam.Sam and got rid of the old Sam struct which was just wrapping the header and records. Other than these additions there are a lot more tests. Everything but methods.go now had >80% test coverage. Methods are a bit tricky since they need to interface between types. This will be revisited in the future, maybe when I update the interval package.


sam/io.go
```diff
@@ -0,0 +1,253 @@
...
+	WriteHeaderToFileHandle(file, header)
+	for i := range data {
+		WriteToFileHandle(file, data[i])
+	}
+	err := file.Close()
+	exception.PanicOnErr(err)
+}
+
+// WriteToFileHandle writes a single Sam struct to the input file.
+func WriteToFileHandle(file *fileio.EasyWriter, aln Sam) {
```
craiglowe, line 241: Can this be the more general io.Writer instead of *fileio.EasyWriter?

sam/metadata.go
```diff
@@ -0,0 +1,188 @@
+package sam
```
craiglowe, line 1: Quick comment to re-start my review here

craiglowe (commented):
Wonderful work!  I only made it about 1/3 of the way through, so I am posting my comments so far and will hopefully come back to finish my review soon.

alleles/countSam.go
```diff
@@ -77,13 +77,13 @@ func sendPassedPositionsSam(answer chan<- *Allele, aln *sam.SamAln, samFilename
...
 // countSamRead adds the bases in a single sam read to the currAlleles map
-func countSamRead(aln *sam.SamAln, currAlleles map[Coordinate]*AlleleCount, runningCount []*Coordinate, ref map[string][]dna.Base, minMapQ int64, progress int) (map[Coordinate]*AlleleCount, []*Coordinate) {
+func countSamRead(aln sam.Sam, currAlleles map[Coordinate]*AlleleCount, runningCount []*Coordinate, ref map[string][]dna.Base, minMapQ int64, progress int) (map[Coordinate]*AlleleCount, []*Coordinate) {
 
 	if aln.Cigar[0].Op == '*' {
 		return currAlleles, runningCount
 	}
 
-	if aln.MapQ < minMapQ {
+	if int64(aln.MapQ) < minMapQ {
```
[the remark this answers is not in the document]
DanielSnellings: You are right of course. This is a patch job, but I don't think it is worth changing this value and dealing with the cascading effects down to the cmd calling this function (at least in this PR). Especially since the alleles package is likely to be entirely rewritten in the future.

cmd/samCoverage/samCoverage.go
```diff
@@ -4,16 +4,34 @@ import (
...
 )
 
+func totalAlignedBases(filename string) int {
+	samFile := fileio.EasyOpen(filename)
+	defer samFile.Close()
+	var done bool = false
+	var aln sam.Sam
+	var alignedBases int
+
+	sam.ReadHeader(samFile)
```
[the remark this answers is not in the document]
DanielSnellings: Agreed, I'll leave this to the author of this function.

sam/methods.go
```diff
@@ -56,19 +56,21 @@ func (s SamSlice) Write(filename string) { // Does not write header
...
-	var next *SamAln
+	var next *Sam
+	var curr Sam
 	for nextBytes, err := file.Peek(1); next == nil && !done; nextBytes, err = file.Peek(1) {
 		if err == nil && nextBytes[0] == '@' {
 			fileio.EasyNextLine(file)
 			continue
 		}
-		next, done = NextAlignment(file)
+		curr, done = ReadNext(file)
```
craiglowe, line 72: I am thinking that ReadNext already has the logic to skip the header and just give us the next alignment no matter where the file handler is in the file.  Can't we just call ReadNext(file) here and avoid the whole nextBytes for loop?

craiglowe (approved):
Really wonderful body of work here!

sam/methods.go
DanielSnellings: Yes. This change has been made.

sam/methods.go
```diff
@@ -56,29 +56,26 @@ func (s SamSlice) Write(filename string) { // Does not write header
...
 	var done bool
-	var next *SamAln
-	for nextBytes, err := file.Peek(1); next == nil && !done; nextBytes, err = file.Peek(1) {
-		if err == nil && nextBytes[0] == '@' {
-			fileio.EasyNextLine(file)
-			continue
-		}
-		next, done = NextAlignment(file)
-	}
+	var next *Sam
```
craiglowe, line 65: Not a bug, but it seems like next could be eliminated by changing line 73 to *s=curr

craiglowe (approved):
Renewing approval

craiglowe (approved):
Renew approval

cmd/samCoverage/samCoverage.go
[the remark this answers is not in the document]
rimangan: I'm not sure what you mean, it seems like we do report that number in the print below. The catch here is that by feeding in a bed instead of the sam header, we are able to report coverage relative to the "ungapped" regions instead of coverage relative to the assembly. Just a bit of flexibility, but the default behavior could conceivably be to use the header for genomeSize.

fasta/info.go
```diff
@@ -16,3 +17,14 @@ func IsFasta(filename string) bool {
 		return false
 	}
 }
+
+// ToChromInfo converts a []Fasta into a []ChromInfo. Useful for applications
+// that do not require the entire fasta sequence to be kept in memory, but just
+// the name, size, and order of fasta records.
+func ToChromInfo(records []Fasta) []chromInfo.ChromInfo {
```
rimangan, line 24: another kentUtil reimplemented

rimangan (approved):

cfauci (approved):

PR #458, opened 2021-03-30
Nuanced error handling in fileio

DanielSnellings:
Riley pointed out that permission errors should probably fatal instead of panic during file open/close. Additions to fileio make permission errors and exists/notExist errors fatal, where all other non-nil errors panic.

### DIFF
--- a/fileio/easyio.go
+++ b/fileio/easyio.go
@@ -80,8 +80,7 @@ func EasyPeekReal(file *EasyReader, n int) ([]byte, error) {
 
 // EasyRemove deletes the input file.
 func EasyRemove(filename string) {
-	err := os.Remove(filename)
-	exception.PanicOnErr(err)
+	MustRemove(filename)
 }
 
 // Close the receiving EasyReader.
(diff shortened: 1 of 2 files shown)


fileio/easyio.go
```diff
@@ -80,8 +80,7 @@ func EasyPeekReal(file *EasyReader, n int) ([]byte, error) {
 
 // EasyRemove deletes the input file.
 func EasyRemove(filename string) {
-	err := os.Remove(filename)
-	exception.PanicOnErr(err)
+	MustRemove(filename)
```
craiglowe, line 83: Should EasyRemove disappear?  Seems like places that call EasyRemove could be changed to call MustRemove

craiglowe (approved):
Looks great!  One comment

fileio/easyio.go
DanielSnellings: It could, though the same would be true of EasyPeekReal, EasyNextRealLine, and EasyNextLine which just wrap the non-easy version of the function.

rimangan (approved):

cfauci (approved):

cfauci:
Should I merge this or do we want to have a discussion on EasyRemove vs MustRemove?

craiglowe:
I think we are good to merge.  I can click the button now.

PR #462, opened 2021-04-08
Variants

DanielSnellings:
This is part of a continuing effort to make working with genetic variation more tractable in gonomics. In the past myself and others have typically used vcf records to build functions that evaluate genetic variation. This is obviously the natural choice as it is the standard file format for documenting genetic variation and can fit numerous different types of variants into a (somewhat) tidy format. However, the flexibility of Vcf also makes it very difficult to work with. There are many corner cases that you must consider and an unpleasant amount of parsing needs to be done. I know we would all love to make ref and alt in Vcf a []dna.Base, but because the file format does not have that constraint, we cannot do this. The goal of this PR is to work towards a simpler way to represent genetic variation in gonomics. The Gene package gets at this a little bit, but it is very limited in scope. Here I lay out a couple ideas for simple variant structs: 'Substitutions' 'Insertions' 'Deletions' and 'Delins' are all now separate structs that can be handled independently or through the use of interfaces. I do not really implement these new structs in this PR for the sake of keeping it small, but I would like to move the Gene package towards using Variant structs. This is definitely a WIP so comments/ideas would be very much appreciated. 

P.S. don't be too intimidated by the newline count. The docs and tests are quite verbose.

### DIFF
--- a/variant/effect.go
+++ b/variant/effect.go
@@ -165,7 +165,7 @@ func (d Deletion) Effect(codingSeq []dna.Base, offsetStart int, offsetEnd int) (
 		return answer, ErrInvalidPosition
 	}
 
-	delLen := offsetEndPos - offsetEndPos
+	delLen := offsetEndPos - offsetStartPos
 	startFrame := offsetStartPos % 3
 	endFrame := (offsetEndPos - 1) % 3
 	codonStart := offsetStartPos - startFrame       // start of first codon affected (closed)
@@ -212,13 +212,12 @@ func (di Delins) Effect(codingSeq []dna.Base, offsetStart int, offsetEnd int) (C
 		return answer, ErrInvalidPosition
 	}
 
-	delLen := offsetEndPos - offsetEndPos
+	delLen := offsetEndPos - offsetStartPos
 	lenDiff := len(di.InsSeq) - delLen
 	startFrame := offsetStartPos % 3
 	endFrame := (offsetEndPos - 1) % 3
 	codonStart := offsetStartPos - startFrame       // start of first codon affected (closed)
 	codonEnd := ((offsetEndPos - 1) - endFrame) + 3 // end of last codon affected (open)
-
 	switch {
 	case lenDiff%3 != 0: // frameshift
 		answer.Type = Frameshift
@@ -243,10 +242,16 @@ func (di Delins) Effect(codingSeq []dna.Base, offsetStart int, offsetEnd int) (C
 	default: // in frame & does not disrupt a codon
 		if lenDiff > 0 {
 			answer.Type = InFrameInsertion
-		} else {
+		} else if lenDiff < 0 {
 			answer.Type = InFrameDeletion
+		} else {
+			answer.Type = Missense
 		}
 		answer.RemovedAa, answer.AddedAa, protOffset = aaChange(codingSeq[codonStart:codonEnd], di.InsSeq)
+
+		if len(answer.AddedAa) == 0 && len(answer.RemovedAa) == 0 {
+			answer.Type = Silent
+		}
 	}
 
 	answer.ProteinPos += protOffset
--- a/variant/effect.go
+++ b/variant/effect.go
@@ -126,16 +126,16 @@ func (i Insertion) Effect(codingSeq []dna.Base, offsetStart int, offsetEnd int) 
 	case len(i.Seq)%3 != 0: // frameshift
 		answer.Type = Frameshift
 		shiftedSeq := make([]dna.Base, len(i.Seq)+len(codingSeq[codonStart:]))
-		copy(shiftedSeq[:frame], codingSeq[codonStart:offsetPos]) // add bases before insertion
-		copy(shiftedSeq[frame:frame+len(i.Seq)], i.Seq)                           // add inserted bases
-		copy(shiftedSeq[frame+len(i.Seq):], codingSeq[offsetPos:])           // add bases after insertion
+		copy(shiftedSeq[:frame], codingSeq[codonStart:offsetPos])  // add bases before insertion
+		copy(shiftedSeq[frame:frame+len(i.Seq)], i.Seq)            // add inserted bases
+		copy(shiftedSeq[frame+len(i.Seq):], codingSeq[offsetPos:]) // add bases after insertion
 		answer.RemovedAa, answer.AddedAa, protOffset = aaChange(codingSeq[codonStart:], shiftedSeq)
 
 	case frame != 0: // disrupts a codon
 		answer.Type = InFrameInsertion
 		insSeq := make([]dna.Base, len(i.Seq)+3)
-		copy(insSeq[:frame], codingSeq[codonStart:offsetPos])   // add bases before insertion
-		copy(insSeq[frame:frame+len(i.Seq)], i.Seq)                             // add inserted sequence
+		copy(insSeq[:frame], codingSeq[codonStart:offsetPos])              // add bases before insertion
+		copy(insSeq[frame:frame+len(i.Seq)], i.Seq)                        // add inserted sequence
 		copy(insSeq[frame+len(i.Seq):], codingSeq[offsetPos:codonStart+3]) // add bases after insertion
 		answer.RemovedAa, answer.AddedAa, protOffset = aaChange(codingSeq[codonStart:codonStart+3], insSeq)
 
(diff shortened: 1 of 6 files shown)


variant/variant.go
```diff
@@ -0,0 +1,55 @@
...
+	Chr   string
+	Start int
+	End   int
+}
+
+// Delins describes a combined insertion and a deletion of
+// a sequence. Useful for describing complex variants.
+// Start and End are the 0-base left-closed right-open interval
+// being deleted (e.g. ATG del [1,3) ins C -> AC).
+type Delins struct {
```
craiglowe, line 39: I am not against the Delins, but why not leave it as a deletion followed by an insertion, or an insertion followed by a deletion?  Off the top of my head, I can't think of a special case why we would want to pair these two events and treat them as one, rather than keeping them as separate events where one happens after the other.  I am thinking about a slice of variants where the order of the variants in a slice would describe the order that those variants should be applied to the starting sequence.  We would again run into the question of if variants after the initial insertion/deletion/delins would use the "original" or "current" coordinate system.  Maybe the "current" system so that we could mutate a base that had been inserted and was not in the original sequence?

variant/effect.go
```diff
@@ -0,0 +1,276 @@
...
+
+	switch {
+	case altAa == refAa:
+		answer.Type = Silent
+
+	case altAa == dna.Stop:
+		answer.Type = Nonsense
+
+	default:
+		answer.Type = Missense
```
craiglowe, line 100: Should silent be an option?  I am thinking of mutating a DNA sequence at a site(s) and wanting to know the effect of those mutations

variant/effect.go
```diff
@@ -0,0 +1,276 @@
...
+
+	if offsetPos > len(codingSeq) {
+		return answer, ErrInvalidPosition
+	}
+
+	frame := offsetPos % 3
+	codonStart := offsetPos - frame
+
+	switch {
+	case len(i.Seq)%3 != 0: // frameshift
```
craiglowe, line 126: I realize I am hitting corner cases, but do we know this is a frameshift at this point?  I am imagining the case where I insert a series of bases that is equal to the bases that come after the insertion (tandem duplication; protein seq remains identical).  It seems that to have a frameshift, we have to know that the insertion is not bringing its own stop codon.  Maybe we know that already, or there is a catch for it later on?

variant/effect.go
```diff
@@ -0,0 +1,276 @@
...
+	codonStart := offsetStartPos - startFrame       // start of first codon affected (closed)
+	codonEnd := ((offsetEndPos - 1) - endFrame) + 3 // end of last codon affected (open)
+
+	switch {
+	case delLen%3 != 0: // frameshift
+		answer.Type = Frameshift
+		shiftedSeq := make([]dna.Base, len(codingSeq[codonStart:])-delLen)
+		copy(shiftedSeq[:startFrame], codingSeq[codonStart:offsetStartPos]) // add bases before deletion
+		copy(shiftedSeq[startFrame:], codingSeq[offsetEndPos:])             // add bases after deletion
+		answer.RemovedAa, answer.AddedAa, protOffset = aaChange(codingSeq[codonStart:], shiftedSeq)
```
craiglowe, line 180: I think aaChange handles the case of the stop codon being deleted, but I didn't see code handling the start codon being deleted.  From my quick look, it seemed like we would still start translating, even though there is no start codon

variant/effect.go
```diff
@@ -0,0 +1,276 @@
...
+	Frameshift       EffectType = 5 // shifts reading frame
+	Nonsense         EffectType = 4 // premature termination
+	InFrameInsertion EffectType = 3 // insertion where insertedBases % 3 == 0
+	InFrameDeletion  EffectType = 2 // deletion where deletedBases % 3 == 0
+	Missense         EffectType = 1 // mutation changes coded amino acid
+)
+
+// Effector is satisfied by types that implement the Effect method, which returns
+// a CodingChange describing how changes prescribed by the receiver type alter the
+// resulting amino acid sequence corresponding to the input []dna.Base. The input
```
craiglowe, line 43: Why not have the input []dna.Base be the whole chromosome?  While it might seem wasteful, that way we only need to store the array once and have different slices use the same array.  I don't quite have the big picture of if I have a bunch of genome-wide mutations and a file listing my genes, how I can then start using these functions.  Would I first truncate my deletions to only cover the protein coding seq?

craiglowe (commented):
Wonderful work here!  I bounced around a little bit to read the files in different order because I found it easiest to start with variant.go.  I had a few comments in places that I was not able to fully understand.

variant/effect.go
```diff
@@ -0,0 +1,276 @@
...
+import (
+	"github.com/vertgenlab/gonomics/dna"
+)
+
+// CodingChange describes a change of a protein sequence caused by
+// a mutation affecting the coding sequence of the corresponding gene.
+// CodingPos and ProteinPos refer to the zero-based start position of
+// the change. The CodingPos always refers to the position changed in
+// origin DNA sequence, however ProteinPos is not always equal to
+// CodingPos / 3. If a change would result in a RemovedAa[0] == AddedAa[0]
```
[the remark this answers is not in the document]
DanielSnellings: If you have the following deletion

```
Prot Pos:  0  1  2  3-n
Sequence: ATGATGGGG~~~~~
Deletion:      ***
New Seq:  ATGATG~~~~~~~
```

Since the deletion did not change the first codon, the protein pos is reported as the first codon that was deleted. In this case, it would be the 'GGG' even though the DNA deletion actually removed part of the 'ATG'.

variant/effect.go
DanielSnellings: You could, but it complicates things greatly. You would need to input CDS starts and ends and the program would become far more complicated. The vision here is to create a fairly simple system that we can build off of. The creation of the CDS sequence would be handled by higher level programs that use the variant infrastructure build here (e.g. the gene package). Having written the complicated version in other packages, I think the tradeoff of efficiency for simplicity is worth it here.

DanielSnellings: To answer your second question, I plan to build things so you can take an input VCF, GTF, FASTA. The GTF/FASTA are converted to Gene structs, and the VCF is converted to variants/Effectors. Then you could make a 
```
func AnnotateVariant(g Gene, e Effector) {
      return e.Effect(g.CDS)
}
```

variant/effect.go
```diff
@@ -0,0 +1,276 @@
...
+// resulting amino acid sequence corresponding to the input []dna.Base. The input
+// sequence should be the CDS of a gene such that input[0:3] is the start codon.
+// The last 3 dna.Base of the input is not required to be a stop codon. Trailing
+// sequence after the stop codon (i.e. the 3' UTR) may be optionally added to the
+// input sequence, however the deletion position must be within the coding sequence.
+// This trailing sequence is used when a variant causes a frameshift and a new stop
+// codon in not found prior to the original stop codon. In this case the translation
+// will continue into the 3' UTR attempting to find a new stop codon.
+//
+// The offset integer input for the Effect method is added to the position in
```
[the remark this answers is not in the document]
DanielSnellings: The two offsets are present for flexibility. I expect the most common case is when you have variants that are built on genomic coordinates. In this case you would need to put in an offset value such that the genomic index is shifted into the proper CDS index (i.e. the offset might be gene.StartPos). In cases of deletions spanning splice junctions the offsetEnd may be different from the offsetStart. The task of finding the proper offsets to use for each variant is delegated to a higher level function.

variant/effect.go
DanielSnellings: The first case checks if the altAa is the same as the refAa. If it is, the effect is reported as Silent.

variant/effect.go
DanielSnellings: I think I see what you mean. So if we have a 301bp insertion that duplicates the last 301bp of the gene then, while technically it creates a frameshift, the protein sequence is actually unchanged so we would want to report it differently. I have added a new check for this after the switch case to determine if the effect is silent.

variant/effect.go
DanielSnellings: Yes, I was not quite sure how to handle this. There are cases of non-AUG start codons so I felt it was safer to let this be handled downstream by checking if ProteinPos == 0. Open to alternative ideas however.

variant/variant.go
DanielSnellings: When i initially wrote this I did in fact have Delins as just a Deletion followed by an Insertion. This worked fine for Mutate, but it complicates Effect quite a bit. As a result I decided to split it into its own type. This not only makes Effect prediction a bit easier, it also allows for several small efficiencies that are not possible when doing a deletion and insertion in series. It was also partly motivated by HGVS guidelines (https://varnomen.hgvs.org/recommendations/DNA/variant/delins/). They were not necessarily designed with programing in mind, but they recognize it as a distinct class of mutations.

rimangan (approved):

cfauci (approved):

cshoben (approved):

PR #463, opened 2021-04-12
Vcf to variants

ddsnellings:
Function in VCF package to split a channel of vcf structs to their respective variant representations

### DIFF
--- a/vcf/toVariant.go
+++ b/vcf/toVariant.go
@@ -71,32 +71,54 @@ func GoChanToVariants(c <-chan Vcf) SmallVariantChans {
 // Substitution, Insertion, Deletion, or Delins; parses the data to the appropriate
 // variant type, and sends on the respective channel.
 func sendVariant(sendChans SmallVariantChans, chr string, pos int, refSeq []dna.Base, altSeq []dna.Base) {
+	var matchingOffset int
+	refSeq, altSeq, matchingOffset = trimMatchingBases(refSeq, altSeq)
+	pos += matchingOffset
+	pos -= 1 // from 1-base to 0-base
+
 	switch {
 	case len(refSeq) == 1 && len(altSeq) == 1: // Substitution
-		sendChans.Substitutions <- variant.Substitution{Chr: chr, Pos: pos - 1, Ref: refSeq[0], Alt: altSeq[0]}
+		sendChans.Substitutions <- variant.Substitution{Chr: chr, Pos: pos, Ref: refSeq[0], Alt: altSeq[0]}
 
-	case len(refSeq) < len(altSeq) && len(refSeq) == 1: // Insertion
+	case len(refSeq) < len(altSeq) && len(refSeq) == 0: // Insertion
 		// e.g. ATG -> ATCG
 		// VCF: pos = 2; ref = T; alt = TC
 		// Desired: pos = 2; insSeq = C
-		sendChans.Insertions <- variant.Insertion{Chr: chr, Pos: pos, Seq: altSeq[1:]} // No -1 on pos since first base trimmed
+		sendChans.Insertions <- variant.Insertion{Chr: chr, Pos: pos, Seq: altSeq}
 
-	case len(refSeq) > len(altSeq) && len(altSeq) == 1: // Deletion
+	case len(refSeq) > len(altSeq) && len(altSeq) == 0: // Deletion
 		// e.g. ATG -> AG
 		// VCF: pos = 1; ref = AT; alt = A
 		// Desired: start = 1; end = 2
-		sendChans.Deletions <- variant.Deletion{Chr: chr, Start: pos, End: pos + len(refSeq[1:])}
+		sendChans.Deletions <- variant.Deletion{Chr: chr, Start: pos, End: pos + len(refSeq)}
 
-	case len(refSeq) > 1 && len(altSeq) > 1: // Delins
+	case len(refSeq) > 0 && len(altSeq) > 0: // Delins
 		// e.g. ATG -> AC
 		// VCF: pos = 1; ref = ATG; alt = AC
 		// Desired: start = 1; end = 3; insSeq = C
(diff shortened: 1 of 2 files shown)


vcf/toVariant.go
```diff
@@ -0,0 +1,136 @@
...
+
+// ChanToVariants splits an incoming channel of VCFs to a channel of variant types.
+// Must be run as a goroutine else the thread will deadlock.
+// See documentation for SmallVariantChans for more information on variant channels.
+func ChanToVariants(c <-chan Vcf, sendChans SmallVariantChans) {
+	for v := range c {
+		if !canParseToBases(v.Ref) {
+			continue // cannot parse vcf record so skip
+		}
+		var refSeq, altSeq []dna.Base
```
craiglowe, line 48: Would it be better to put these up outside the for loop so that we don't keep re-declaring them?

vcf/toVariant.go
```diff
@@ -0,0 +1,136 @@
...
+	answer.Records = make(chan Vcf, buffSize)
+	return answer
+}
+
+// ChanToVariants splits an incoming channel of VCFs to a channel of variant types.
+// Must be run as a goroutine else the thread will deadlock.
+// See documentation for SmallVariantChans for more information on variant channels.
+func ChanToVariants(c <-chan Vcf, sendChans SmallVariantChans) {
+	for v := range c {
+		if !canParseToBases(v.Ref) {
```
craiglowe, line 45: I think we should include in the comment that vcf records will be silently skipped if the ref (or alt) is not one, or more, bases

vcf/toVariant.go
```diff
@@ -0,0 +1,136 @@
...
+	var matchingOffset int
+	refSeq, altSeq, matchingOffset = trimMatchingBases(refSeq, altSeq)
+	pos += matchingOffset
+	pos -= 1 // from 1-base to 0-base
+
+	switch {
+	case len(refSeq) == 1 && len(altSeq) == 1: // Substitution
+		sendChans.Substitutions <- variant.Substitution{Chr: chr, Pos: pos, Ref: refSeq[0], Alt: altSeq[0]}
+
+	case len(refSeq) < len(altSeq) && len(refSeq) == 0: // Insertion
```
craiglowe, line 83: I thought an insertion in a vcf record would be something like:
ref = A
alt = ACGT
So wouldn't len(refSeq) be equal to 1 instead of zero?  Maybe something is trimming the ref and alt alleles to a new format?  From the comment lines below it looks like this should be a check for len(refSeq) == 1

vcf/toVariant.go
```diff
@@ -0,0 +1,136 @@
...
+		// e.g. ATG -> ATCG
+		// VCF: pos = 2; ref = T; alt = TC
+		// Desired: pos = 2; insSeq = C
+		sendChans.Insertions <- variant.Insertion{Chr: chr, Pos: pos, Seq: altSeq}
+
+	case len(refSeq) > len(altSeq) && len(altSeq) == 0: // Deletion
```
craiglowe, line 89: Some comment as for insertion.  I am guessing there is something trimming the vcf strings before sending them to this func, but maybe the comments should change to reflect this?

vcf/toVariant.go
```diff
@@ -0,0 +1,136 @@
...
+		sendChans.Delins <- variant.Delins{Chr: chr, Start: pos, End: pos + len(refSeq), InsSeq: altSeq}
+
+	default:
+		log.Panicf("could not parse identity of following variant\nChr: %s\nPos: %d\nRef: %s\nAlt: %s",
+			chr, pos, dna.BasesToString(refSeq), dna.BasesToString(altSeq))
+	}
+}
+
+func trimMatchingBases(a, b []dna.Base) (aNew []dna.Base, bNew []dna.Base, offset int) {
+	aNew, bNew = a, b
```
craiglowe, line 108: Do you need these variables?  Can you just keep re-slicing a and b and then return a and b?

vcf/toVariant.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if len(aNew) == 0 && len(bNew) == 0 {
+		log.Panicf("error, all bases match trimmed in\n%s\nand\n%s", dna.BasesToString(a), dna.BasesToString(b))
+	}
+	return
+}
+
+// canParseToBases determines if a given string from a vcf file can be parsed to a []dna.Base
+func canParseToBases(s string) bool {
+	return !strings.ContainsAny(s, ":>[.")
```
craiglowe, line 126: Should this instead be a check for only "acgtACGT"

craiglowe (approved):
All the code looks good.  There are two places I thought some additional comments would be helpful.

vcf/toVariant.go
ddsnellings: Yes, good catch

vcf/toVariant.go
ddsnellings: Yes, done

vcf/toVariant.go
ddsnellings: That is correct, though the call to `trimMatchingBases` on line 78 removes the 'A' from both ref and alt, simplifying the logic

vcf/toVariant.go
ddsnellings: These bases get trimmed inside this function. This is important as multi-allelic sites can give some ref/alt sets where there are multiple matching bases which significantly complicate the logic if not removed
e.g.
ref = ACG
alt = A, ACGCG

vcf/toVariant.go
ddsnellings: You can. I wanted to have named returns to be explicit, but this is probably not worth constantly re-declaring slices. changed.

vcf/toVariant.go
ddsnellings: Yes

rimangan (approved):

craiglowe (approved):
Great work Dan!

cshoben (approved):

Issue #430, opened 2021-03-19
MultiFaBrowser Testing Function


rimangan:
Currently no testing in the Browser package.

PR #464, opened 2021-04-13
Dunn fix

rimangan:
Changing from slice fastas to map fastas led to the addition of an unneeded panic that breaks this program. Fixed

### DIFF
--- a/popgen/dunn.go
+++ b/popgen/dunn.go
@@ -5,8 +5,6 @@ import (
 	"github.com/vertgenlab/gonomics/dna"
 	"github.com/vertgenlab/gonomics/fasta"
 	"github.com/vertgenlab/gonomics/numbers"
-	"log"
-
 	//"log"
 	"math"
 )
@@ -41,19 +39,15 @@ func Dunn(b *bed.Bed, aln []fasta.Fasta, g []*Group) (float64, string) {
 func FindMaxIntra(subFa []fasta.Fasta, g *Group) int {
 	var answer int = 0
 	var faI, faJ []dna.Base
-	var faFound bool
+	var faFoundI, faFoundJ bool
 	faMap := fasta.ToMap(subFa)
 	for i := 0; i < len(g.Members); i++ {
 		for j := i + 1; j < len(g.Members); j++ {
-			faI, faFound = faMap[g.Members[i]]
-			if !faFound {
-				log.Panicf("%s not found in fasta map", g.Members[i])
+			faI, faFoundI = faMap[g.Members[i]]
+			faJ, faFoundJ = faMap[g.Members[j]]
+			if faFoundI && faFoundJ {
+				answer = numbers.Max(answer, dna.Dist(faI, faJ))
 			}
-			faJ, faFound = faMap[g.Members[j]]
-			if !faFound {
-				log.Panicf("%s not found in fasta map", g.Members[i])
-			}
-			answer = numbers.Max(answer, dna.Dist(faI, faJ))
 		}
 	}
 	return answer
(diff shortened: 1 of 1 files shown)


ddsnellings (dismissed):

popgen/dunn.go
```diff
@@ -41,19 +42,15 @@ func Dunn(b *bed.Bed, aln []fasta.Fasta, g []*Group) (float64, string) {
...
+	var faFoundI, faFoundJ bool
 	faMap := fasta.ToMap(subFa)
 	for i := 0; i < len(g.Members); i++ {
 		for j := i + 1; j < len(g.Members); j++ {
-			faI, faFound = faMap[g.Members[i]]
-			if !faFound {
-				log.Panicf("%s not found in fasta map", g.Members[i])
+			faI, faFoundI = faMap[g.Members[i]]
+			faJ, faFoundJ = faMap[g.Members[j]]
+			if faFoundI && faFoundJ {
```
craiglowe, line 51: Seems like it should be an error if one of the seq is not found?  This would mean someone asked the function to calculate Dunns for a group of sequence names, but then not all of those names had sequences.

popgen/dunn.go
```diff
@@ -64,18 +61,14 @@ func FindMinInter(g []*Group, subFa []fasta.Fasta) int {
...
-			faI, faFound = faMap[g[0].Members[i]]
-			if !faFound {
-				log.Panicf("%s not found in fasta map", g[0].Members[i])
-			}
-			faJ, faFound = faMap[g[1].Members[j]]
-			if !faFound {
-				log.Panicf("%s not found in fasta map", g[1].Members[j])
+			faI, faFoundI = faMap[g[0].Members[i]]
+			faJ, faFoundJ = faMap[g[1].Members[j]]
+			if faFoundI && faFoundJ {
```
craiglowe, line 69: Similar comment to above.  Seems like it should be an error if these seq are not found.  I worry about asking for the Dunn's for a group of 10 against a group of 5, and it silently calcs a group fo 8 against a group of 4.

craiglowe (commented):
Great to see these improvements!  I had two related questions.

browser/multiFa.go
```diff
@@ -22,7 +22,7 @@ func MultiFaVisualizer(infile string, outfile string, start int, end int, noMask
 	if noMask {
 		fasta.AllToUpper(records)
 	}
-	for i := range records {
+	for i := 1; i < len(records); i++ {
```
rimangan, line 25: here a range statement was added incorrectly by me, as this loop should be from 1 to n, not 0 to n. Fixed here.

popgen/dunn.go
rimangan: So we could go either way on this. Right now if some samples are missing, this function still attempts to calculate Dunn's with the remaining samples. The function that calls this one reports which samples were missing. However, for now we have been discarding any regions with any missing samples, so perhaps this is not so useful

popgen/dunn.go
```diff
@@ -5,15 +5,13 @@ import (
 	"github.com/vertgenlab/gonomics/dna"
 	"github.com/vertgenlab/gonomics/fasta"
 	"github.com/vertgenlab/gonomics/numbers"
-	"log"
-
 	//"log"
 	"math"
 )
 
 //Dunn returns the dunn index as a float64 and missing group members as a string.
```
ddsnellings, line 12: could you add a comment describing the new int return?

ddsnellings (approved):

rimangan:
Resolves #430

craiglowe (approved):
Great to see these fixes and improvements!

cshoben (approved):

PR #466, opened 2021-04-21
Remove vcf pointers

ddsnellings:
All tests pass. This may break some functions that do not have tests. The most common error to look out for is calls to functions that modify the vcf but did not previously have a vcf return (e.g. InvertVcf).


rimangan (approved):

cmd/vcfFilter/vcfFilter_test.go
```diff
@@ -32,8 +32,8 @@ func TestVcfFilter(t *testing.T) {
 	var err error
 	for _, v := range VcfFilterTests {
 		vcfFilter(v.inputFile, "tmp.vcf", v.groupFile, v.chrom, v.minPos, v.maxPos, v.ref, v.alt, v.minQual, v.biAllelicOnly, v.substitutionsOnly, v.segregatingSitesOnly, v.removeNoAncestor, v.onlyPolarizableAncestors)
-		records := vcf.Read("tmp.vcf")
-		expected := vcf.Read(v.expectedOutputFile)
+		records, _ := vcf.Read("tmp.vcf")
```
craiglowe, line 35: This is obviously minor, but is it natural for the header to come before the data?  Meaning that we would get back: header, data rather than data, header.

interval/lift.go
```diff
@@ -43,7 +43,8 @@ func ReadToLiftChan(inputFile string, send chan<- Lift) {
 	case ".vcf":
 		receive, _ := vcf.GoReadToChan(inputFile)
 		for val := range receive {
-			send <- val
+			curr := val
```
craiglowe, line 46: I don't see a bug here, but I am curious why we don't send val directly and instead make a copy of it into curr and then send curr?

vcf/io.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/vertgenlab/gonomics/common"
+	"github.com/vertgenlab/gonomics/fileio"
+	"io"
+	"log"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+// Read parses a slice of VCF structs from an input filename. Does not store or return the header.
```
craiglowe, line 14: This func seems to return the header.

vcf/io.go
```diff
@@ -0,0 +1,213 @@
...
+	return answer, header
+}
+
+// ReadToChan is a helper function of GoReadToChan.
+func ReadToChan(file *fileio.EasyReader, data chan<- Vcf, wg *sync.WaitGroup) {
+	for curr, done := NextVcf(file); !done; curr, done = NextVcf(file) {
+		data <- curr
+	}
+	file.Close()
+	wg.Done()
```
craiglowe, line 34: I think we can save a go routine by closing data here.

vcf/io.go
```diff
@@ -0,0 +1,213 @@
...
+// GoReadToChan parses VCF structs from an input filename and returns a chan of VCF structs along with the header of the VCF file.
+func GoReadToChan(filename string) (<-chan Vcf, Header) {
+	file := fileio.EasyOpen(filename)
+	header := ReadHeader(file)
+	var wg sync.WaitGroup
+	data := make(chan Vcf)
+	wg.Add(1)
+	go ReadToChan(file, data, &wg)
+
+	go func() {
```
craiglowe, line 46: I believe this go routine  can be eliminated by closing data at the comment above

vcf/io.go
```diff
@@ -0,0 +1,213 @@
...
+	wg.Done()
+}
+
+// GoReadToChan parses VCF structs from an input filename and returns a chan of VCF structs along with the header of the VCF file.
+func GoReadToChan(filename string) (<-chan Vcf, Header) {
+	file := fileio.EasyOpen(filename)
+	header := ReadHeader(file)
+	var wg sync.WaitGroup
+	data := make(chan Vcf)
+	wg.Add(1)
```
craiglowe, line 43: I don't think this waitgroup is needed (in relation to nearby comments).

vcf/io.go
```diff
@@ -0,0 +1,213 @@
...
+	} else {
+		_, err = fmt.Fprintf(file, "%s\t%v\t%s\t%s\t%s\t%v\t%s\t%s\t%s\t%s\n", input.Chr, input.Pos, input.Id, input.Ref, strings.Join(input.Alt, ","), input.Qual, input.Filter, input.Info, FormatToString(input.Format), SamplesToString(input.Samples))
+	}
+	common.ExitIfError(err)
+}
+
+// Write writes a []Vcf to an output filename.
+func Write(filename string, data []Vcf) {
+	file := fileio.EasyCreate(filename)
+	defer file.Close()
```
craiglowe, line 189: Should we check the error on this?

vcf/methods.go
```diff
@@ -70,23 +70,32 @@ func (v *VcfSlice) Pop() interface{} {
...
-	WriteVcf(file, v)
+	WriteVcf(file, *v)
 }
 
 func (v *Vcf) NextRealRecord(file *fileio.EasyReader) bool {
 	var done bool
-	var next *Vcf
-	for next == nil && !done {
+	var next Vcf
+	for next.Chr == "" && !done {
```
craiglowe, line 92: I am not sure I understand this loop.  It is not clear to me why we would call NextVcf more than once for a single call to NextRealRecord

craiglowe (commented):
Great work!  Few minor comments

julianaccarvalho (approved):

craiglowe (approved):
Great to see these improvements!

PR #468, opened 2021-04-22
vcfAncestorAnnotation Speedup

rimangan:
A small change in the refPosToAlnPos structure allows vcfAncestorAnnotation to move from O(n^2) to O(n). Now we save the refPos/alnPos pair instead of starting over each time.


fasta/multiFa.go
```diff
@@ -22,6 +23,19 @@ func RefPosToAlnPos(record Fasta, RefPos int) int {
 	return AlnPos
 }
 
+//RefPosToAlnPosCounter is like RefPosToAlnPos, but the starting AlnPos and refCounter can be set.
```
craiglowe, line 26: I think a programmer might need more explanation for what to pass in as refStart and alnStart

fasta/multiFa.go
```diff
@@ -22,6 +23,19 @@ func RefPosToAlnPos(record Fasta, RefPos int) int {
...
+func RefPosToAlnPosCounter(record Fasta, RefPos int, refStart int, alnStart int) int {
+	for t := alnStart; refStart < RefPos; t++ {
```
craiglowe, line 28: Do we need to check the input to make sure that refStart is less than or equal to RefPos?  If not, we could set both refStart and alnStart to zero.

vcf/vcfAncestor_test.go
```diff
@@ -68,10 +68,11 @@ func TestVcfAnnotateAncestorFromFa(t *testing.T) {
 	reader, _ := GoReadToChan("testdata/TestVcfAnnotateAncestorFromFa.vcf")
 	records := fasta.Read("testdata/testAncestorSequence.fa")
 	var i int = 0
+	var currRef, currAln int = 0, 0
```
craiglowe, line 71: Do we need something inside the loop that sets these back to zero when we go to a new chromosome?

craiglowe (commented):
Looks like an excellent speed-up!  I had a few comments.

ddsnellings:
Do you think it may be more efficient to just spend the time upfront to initialize a []int of length len(fasta.Seq) that stores the gap value for each position?

ddsnellings (approved):

fasta/multiFa.go
```diff
@@ -7,19 +7,25 @@ import (
...
-	var refCounter int = 0
+	return RefPosToAlnPosCounter(record, RefPos, 0, 0)
+}
 
-	for t := 0; refCounter < RefPos; t++ {
-		AlnPos++
+//RefPosToAlnPosCounter is like RefPosToAlnPos, but can begin midway through a chromosome at a refPosition/alnPosition pair, defined by the input variables refStart and alnStart.
+func RefPosToAlnPosCounter(record Fasta, RefPos int, refStart int, alnStart int) int {
+	if refStart > RefPos {
+		refStart, alnStart = 0, 0 //in case the refStart was improperly set (greater than the desired position, we reset these counters to 0.
```
cshoben, line 18: Could you explain why we wouldnt want/it doesn't work for refStart to be greater than RefPos? And why we would want to reset the alnStart if the refStart is greater than RefPos? I'm thinking mid-way through a chromosome these start positions would always be greater (if we were on the positive sense).

cshoben (approved):
Looks great! I had a few conceptual questions but nothing to hold up an approval.

julianaccarvalho (approved):

craiglowe (approved):
Great work!

PR #473, opened 2021-05-07
randSeq

rimangan:
RandSeq produces randomly generated DNA sequences. Options include specified GC content, sequence number, sequence length, and pseudorandom number control. Makes use of the simulate package code for sequence generation. Passes tests.


craiglowe (approved):
Great!

dasnellings (approved):

julianaccarvalho (approved):

PR #476, opened 2021-05-11
Major MCMC Updates

rimangan:
Changes to selectionMCMC:

1. Changing the proposal distribution for the sigma hyperparameter to a normal distribution instead of a gamma distribution. Sigma is now bounded between 0 and 0.5.
2. Set verbose as a user-specified option. Might be a good template for other complex programs.
3. New divergence-based ascertainment bias correction occurs base-by-base.


cmd/selectionMCMC/selectionMCMC.go
```diff
@@ -57,6 +58,8 @@ func main() {
 		UnPolarized:   *unPolarized,
 		Derived:       *derived,
 		Ancestral:     *ancestral,
+		FixedSigma:    *fixedSigma,
+		D:             1, //D is hardcoded as 1 for now. This represents the size of the ascertainment subset.
```
craiglowe, line 62: I am not totally sure what this value captures.  For us is it the number of sites where hg38 does not match the HCA?

rimangan: This refers to ascertainment correction, where the condition for ascertainment was divergence or nondivergence between 1 human and 1 ancestor sequence, thus D=1.

cmd/afsPmf/afsPmf.go
```diff
@@ -9,8 +9,8 @@ import (
 	"log"
 )
 
-func afsPmf(vcfFile string, outFile string, unPolarized bool) {
-	g, err := popgen.VcfToAfs(vcfFile, !unPolarized) //VcfToAfs is written in terms of polarized, so this is inverted here.
+func afsPmf(vcfFile string, outFile string, s popgen.McmcSettings) {
+	g, err := popgen.VcfToAfs(vcfFile, s) //VcfToAfs is written in terms of polarized, so this is inverted here.
```
craiglowe, line 13: Do we need to pass the whole settings struct?  I am guessing that we are only using a couple of those settings, but for a programmer it might be hard to know what ones need to be set.  Do we only use unPolarized?  Maybe we can just pass in s.UnPolarized?

cmd/vcfAfs/vcfAfs.go
```diff
@@ -9,8 +9,8 @@ import (
 	"log"
 )
 
-func vcfAfs(vcfFile string, outFile string, unPolarized bool) {
-	g, err := popgen.VcfToAfs(vcfFile, !unPolarized) //VcfToAFS is written in terms of polarized, so this is inverted here.
+func vcfAfs(vcfFile string, outFile string, s popgen.McmcSettings) {
+	g, err := popgen.VcfToAfs(vcfFile, s)
```
craiglowe, line 13: Can we just pass the value for unpolarized?

popgen/mcmcSelection.go
```diff
@@ -114,67 +110,71 @@ func InitializeTheta(m float64, sig float64, data Afs, binomCache [][]float64, s
...
 		log.Printf("Initial Theta: mu: %f. sigma: %f. LogLikelihood: %e.", t.mu, t.sigma, t.likelihood)
 	}
-	if verbose == 1 {
+	if s.Verbose == 1 {
 		log.Printf("OldMu\tNewMu\tOldLikelihood\tNewLikelihood\tpAccept\tlogRand\tDecision\n")
 	}
-	fmt.Fprintf(out, "Iteration\tMu\tSigma\tAccept\n")
-
+	_, err = fmt.Fprintf(out, "Iteration\tMu\tSigma\tAccept\n")
+	if err != nil {
```
craiglowe, line 159: I don't think you need this "if", I think PanicOnErr does it for you

popgen/mcmcSelection.go
```diff
@@ -114,67 +110,71 @@ func InitializeTheta(m float64, sig float64, data Afs, binomCache [][]float64, s
...
-		if MetropolisAccept(t, tCandidate, s.SigmaStep) {
+		if MetropolisAccept(t, tCandidate, s) {
 			t = tCandidate
 			currAccept = true
 		} else {
 			currAccept = false
 		}
-		fmt.Fprintf(out, "%v\t%e\t%e\t%t\n", i, t.mu, t.sigma, currAccept)
+		_, err = fmt.Fprintf(out, "%v\t%e\t%e\t%t\n", i, t.mu, t.sigma, currAccept)
+		if err != nil {
```
craiglowe, line 171: I don't think you need this "if"

craiglowe (approved):
Great work!  I am excited to see the results you get!
Passing around the settings struct simplifies some things, but I am wondering if you think it will make things harder to maintain in the future.  It seems like often only a few of the values are used in the function it get passed into, but it seems hard to know which of those values in the struct would need to be set for someone reusing these functions.

popgen/stationarity.go
```diff
@@ -20,15 +20,24 @@ Equations from this thesis that are replicated here are marked.
 
 const IntegralBound float64 = 1e-12
 
+type LikelihoodFunction byte
+
+const (
+	Uncorrected LikelihoodFunction = 0
```
dasnellings, line 26: This is totally fine (and even better in some ways) but just to share, I recently learned about the `iota` keyword for defining consts. If you wanted to, you could write this as 
```
const (
     Uncorrected LikelihoodFunction = iota
     Ancestral
     Derived
)
```

This is useful when you want discrete values for a variable, but don't care about the particular number it is assigned to. This is not any more efficient as far as I know (except typing less), but something to know exists.

dasnellings (dismissed):

dasnellings (approved):

craiglowe (dismissed):
This is great.  I am approving, but all of Dan's comments seem like good things to address.

popgen/stationarity.go
rimangan: This is cool! Implemented in the next commit

dasnellings (approved):

RavenYL (approved):
The changes in the code all look reasonable to me, although I don't know much about model selection by MCMC computation

craiglowe (approved):
Great to have this functionality!

PR #477, opened 2021-05-11
test file and supporting test data for bedMaxWig

cshoben:
This test file was written to test bedMaxWig in it's current state. I plan to make updates to bedMaxWig in the near future but wanted to get at least a basic test file in the system. 

Reminder: bedMaxWig reads in a bed file, wig file, and chrom.sizes file to produce a new bed file where a number is provided in a new column representing the highest wig value present in the corresponding chrom/coordinates (ie first three columns)

One thing you may notice is the output of the bed file has a zero and a dash between the coordinates and the bedRangeMax value (6th column). 
If you look at the bed.Bed struct you'll see why. https://github.com/vertgenlab/gonomics/blob/main/bed/bed.go
The bedRangeMax value is being written into the Annotation field. I considered changing this but I've also read that you can not omit intermediate fields/columns 
https://genome.ucsc.edu/FAQ/FAQformat.html#format1

However, our bed struct does not perfectly follow the additional optional BED fields anyways. I decided in the end to leave it as it was. When visualizing in the genome browser, this bedRangeMax value simply prints over the coordinate block as if it was a name. I thought about putting this bedRangeMax value into the "Name" field/coulmn in the BED but then if we ever do want to bedMaxWig a bed file with those fields, we would be overwriting them and that seemed bad. It's easy to remove any unused column downstream by the user, but also wanted to bring it up in case people have other thoughts.


craiglowe (approved):
Great work!
I think you bring up good points about how we can make this a "legal" bed file (which needs integers), but we will have float values since we are getting the values from a wig.  I think the only legal place to sneak it in is the "name" field.  We could add a delimiter like underscore or colon to put the name and the float score in the name field?  Another option would be to add a multiplier as an option and then report in the score field int(score * multiplier)?

dasnellings (approved):

PR #482, opened 2021-06-01
added HOME/go to check for source files

dasnellings:
also fixed a bug in how the gonomics binary was being checked.

### DIFF
--- a/cmd/gonomics/gonomics.go
+++ b/cmd/gonomics/gonomics.go
@@ -43,7 +43,13 @@ func getBin() (path string, binExists map[string]bool) {
 
 // getGonomicsCmds parses the gonomics source code to return a set of cmd names
 func getGonomicsCmds() map[string]bool {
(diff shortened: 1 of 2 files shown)


cshoben (approved):
These changes are super cool! I second Raven's questions from her original pull request approval. Also, I'd enjoy talking through again the GOPATH versus HOME thing like Craig mentioned in the original pull request at our next code review. Maybe we can all even look at our set up in real time and work together on making sure our set up is correct to use this new feature you've built.

RavenYL (approved):

craiglowe (approved):
I was reading your comment on the difficulty with finding the source code.  I have two ideas, which I think could even be used together:
1) We add the cache file to the repo such that the cache file will reflect what the repo looks like on github main.  This will not help with changes the user has made, but will cover cases where people just want to use the code.  If the program can't find the source, it can just use it's existing cache file.
2) We could add an option to the gonomics command where someone could give the path to either gonomics or the cmd directory (or some other spot that would be a reference point).

PR #483, opened 2021-06-01
Wig update mem usage

cshoben:
take away pointers, including pointers used with wig.Read
no longer handle variableStep wigs, just fixedStep but this can be added back if we end up needing it. 
values { position, value} has been simplified to a [ ] float 64. 
change in WriteToFileHandle so that a value of zero is documented as a “0” instead of a “0.000000” in order to safe mem. 
bedMaxWig, wigPeak.go, convert.go, and samToWig.go updated to handle changes in wig package.


cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -14,15 +14,15 @@ import (
...
+	var wigData []wig.Wig = wig.Read(database)
 	var sizes []chromInfo.ChromInfo = chromInfo.ReadToSlice(chromsizeFile)
 	var outlist []*bed.Bed
 	var currentBed *bed.Bed = records[0]
 	var chromSlice []float64
 	var i int
 
 	for i = 0; i < len(sizes); i++ {
-		chromSlice = WigChromToSlice(data, sizes[i].Size, sizes[i].Name)
+		chromSlice = WigChromToSlice(wigData, sizes[i].Size, sizes[i].Name)
```
rimangan, line 25: Can we get rid of this variable now? Do we use chromSlice since we updated Values to be a []float64?

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -34,16 +34,13 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
 	bed.Write(outfile, outlist, 7)
 }
 
-func WigChromToSlice(w []*wig.Wig, size int, chrom string) []float64 {
+func WigChromToSlice(w []wig.Wig, size int, chrom string) []float64 {
```
rimangan, line 37: Similarly, we might be able to remove this function now

rimangan (approved):
Great work! I had a few small suggestions but it looks great!

RavenYL (approved):
Thanks for updating wigPeaks and preserving my comments too!

convert/convert.go
```diff
@@ -120,19 +121,19 @@ func BedScoreToWig(infile string, reference map[string]chromInfo.ChromInfo) []*w
...
 		}
 
-		wigSlice[chromIndex].Values[midpoint].Value = float64(current.Score)
+		wigSlice[chromIndex].Values[midpoint] = float64(current.Score)
 
 	}
 	return wigSlice
 }
 
 //BedScoreToWigRange uses bed entries from an input file to construct a Wig data structure where the Wig value at each position is equal to the score of an overlapping bed entry, and zero if no bed regions overlap.
```
craiglowe, line 134: This function has a really similar name and description to the above function, BedScoreToWig.  Is there a difference?  I realize this is not part of the current pull request, but I am noticing it now.

convert/convert.go
```diff
@@ -169,37 +171,39 @@ func BedScoreToWigRange(infile string, reference map[string]*chromInfo.ChromInfo
...
 		}
 		chromIndex = getWigChromIndex(current.Chrom, wigSlice)
-		if wigSlice[chromIndex].Values[midpoint].Value != 0 {
+		if wigSlice[chromIndex].Values[midpoint] != 0 {
 			log.Fatalf("Multiple scores for one position.")
 		}
 		for k := current.ChromStart; k < current.ChromEnd; k++ {
 			//DEBUG: fmt.Printf("b[j].Chrom: %s, b[j].ChromStart: %d, b[j].ChromEnd: %d, k: %d, len(wigSlice[chromIndex].Values) %d\n", b[j].Chrom, b[j].ChromStart, b[j].ChromEnd, k, len(wigSlice[chromIndex].Values))
-			wigSlice[chromIndex].Values[k+1].Value = float64(current.Score)
+			wigSlice[chromIndex].Values[k+1] = float64(current.Score)
```
craiglowe, line 179: Here we need to have current.Score be set, but on line 168/170 above we don't complain if they did not provide a score.  I think score probably needs to be mandatory in the input bed file.  I know this is not something introduced here, but something I am now noticing.

convert/convert.go
```diff
@@ -169,37 +171,39 @@ func BedScoreToWigRange(infile string, reference map[string]*chromInfo.ChromInfo
...
 	var i, x int = 0, 0
+	var currentWig wig.Wig
+
 	//generate Wig skeleton from reference
 	for _, v := range reference {
-		currentWig := wig.Wig{StepType: "fixedStep", Chrom: v.Name, Start: 1, Step: 1}
-		currentWig.Values = make([]*wig.WigValue, v.Size)
+		currentWig = wig.Wig{StepType: "fixedStep", Chrom: v.Name, Start: 1, Step: 1}
+		currentWig.Values = make([]float64, v.Size)
 		for x = 0; x < v.Size; x++ {
```
craiglowe, line 196: for loop no longer needed

wig/wig.go
```diff
@@ -1,115 +1,107 @@
...
+			return currentWig, doneReading
 		}
 	}
-	return answer
+	return currentWig, doneReading
 }
 
 //Prints the first record in a Wig struct. Mainly used for debugging.
-func PrintFirst(rec []*Wig) {
+func PrintFirst(rec []Wig) {
```
craiglowe, line 89: Maybe this can disappear?  Or get moved to a debug.go file?  Not for now, but just leaving my thoughts.

wig/wig.go
```diff
@@ -119,27 +111,29 @@ func Write(filename string, rec []*Wig) {
...
 
-	for j := range rec.Values {
+	for i := range rec.Values {
 
 		if rec.StepType == "fixedStep" {
-			_, err = fmt.Fprintf(file, "%f\n", rec.Values[j].Value)
-			common.ExitIfError(err)
-		} else if rec.StepType == "variableStep" {
-			_, err = fmt.Fprintf(file, "%d\t%f\n", rec.Values[j].Position, rec.Values[j].Value)
+			if rec.Values[i] == 0 {
```
craiglowe, line 130: Maybe a more general solution (for all integers instead of only zero) would be:
if rec.Values[i] == float64(int(rec.Values[i])) {
     _, err = fmt.Fprintf(file, "%d\n", int(rec.Values[i]))

craiglowe (commented):
Wow!  Great improvements!  A few small comments

cmd/bedMaxWig/bedMaxWig.go
cshoben: I decided to leave this pull request focused on just getting these wig changes in (I have some other working that cant move forward until then), but I do plan to do a future pull request with bedMaxWig updates and these comment are super helpful on that endeavor!

1) I have already tried to remove this variable but ran into the problem that the rest of the code is written so that the wig and bed are matched up based on the order of the chrom.sizes file. And to adjust for this felt best to be done in another pull request. I dont think that changing records[k].Chrom == wigData.Values will work unless we always expect only one wig struct per chromosome. I am looking at a wig file I am currently working with and I do have only one wig struct per chrom (this wig was made from samToWig) but I'm not sure what is typical of wigs produced elsewhere. 
2) Craig, I dont have a good feel for wigs outside of the ones I'm using which were all produced by samToWig and are all just one struct per chrom (or chrom fragment). It's sounding like maybe the last idea you expressed about sorting would help to ensure both speed and low-mem in either circumstance?

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -34,16 +34,13 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
...
 	output := make([]float64, size)
 	for _, v := range w {
 		if v.Chrom == chrom {
-			for i := 0; i < len(output); i++ {
-				output[i] = v.Values[i].Value
-			}
+			output = v.Values
```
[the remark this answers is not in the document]
cshoben: Yes, current set up is to return the moment a wig chrom name matches the input chrom name in the third argument. It does not continue to search for more once it finds a match. I think this hasnt been a problem because use thus far has been with wigs with only one struct per chromosome.

convert/convert.go
```diff
@@ -89,15 +89,16 @@ func BedScoreToWig(infile string, reference map[string]chromInfo.ChromInfo) []*w
...
 	var current *bed.Bed
+	var currentWig wig.Wig
 
 	//generate Wig skeleton from reference
 	for _, v := range reference {
-		currentWig := wig.Wig{StepType: "fixedStep", Chrom: v.Name, Start: 1, Step: 1}
-		currentWig.Values = make([]*wig.WigValue, v.Size)
+		currentWig = wig.Wig{StepType: "fixedStep", Chrom: v.Name, Start: 1, Step: 1}
+		currentWig.Values = make([]float64, v.Size)
 		for x = 0; x < v.Size; x++ {
```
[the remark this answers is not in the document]
cshoben: Should I remove this in this pull request or do a new one for it?

convert/convert.go
cshoben: They also look identical to me. @rimangan do these look identical to you?

wig/wig.go
cshoben: Interested in talking with you more about a "debug.go" file at a future meeting.

cmd/bedMaxWig/bedMaxWig.go
craiglowe: A lot of our code is only going to have one wig struct per chrom, but many will have more.  chromSlice is going to allow us to more easily deal with multiple wigs per chrom by being a place to hold all the data for a single chrom in one place.  It is not needed, but I think is worth the cost to make the code easier to write, review, debug.  While we don't often create more than one wig struct per chrom, many other programs will, and we should do our best to handle this, but we can save that for a later pull request.
Sorting both files and reading them simultaneously would be the fanciest way to do this, but would require a rewrite and I don't think it is worth the time investment right now.  The improvements you are making in this pull request are wonderful and will be a big help.  If needed, a smaller update would be to process only one chrom at a time from the wig file, but I think we should only do that if it is needed.

convert/convert.go
cshoben: Hmmm, I don't connect how the length of words being greater than or equal to 5 means that a score was not provided?

cmd/bedMaxWig/bedMaxWig.go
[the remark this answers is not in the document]
craiglowe: Don't need to fix this now, but you can you make a github issue about this so that we don't forget?

wig/wig.go
cshoben: Would this be converting all values to an integer? 
- If yes, does this effect my ability to do math on these numbers later?  Like if I do math and there is a long decimal, I would want to be able to keep that. But I think all math would be outside of this program (downstream) so maybe it does not matter. 
Or maybe this "typing" is only specific/important to when we are creating the file? And in that case, why would we use [ ]float64 over [ ]integers?

convert/convert.go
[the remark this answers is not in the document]
craiglowe: I would remove it right here if you agree.  Just delete the 3 lines of this loop.

craiglowe (commented):
Added a few responses to comments

cmd/bedMaxWig/bedMaxWig.go
cshoben: Awesome! This sounds like a great plan. I'll add a github issue about the "not handling multiple wig structs for a single chrom" and remove the unnecessary for loop we discussed in convert.go . I will wait to hear from Riley before I delete that seemingly duplicated function.

edotau (approved):
LGTM

convert/convert.go
craiglowe: The length of "words" is the number of columns in the bed file.  The score is the fifth field, so if the length is not at least 5, then the user did not supply a score for each bed region.  This should probably be an error because it doesn't make sense to convert the scores in the bed file to wiggle format if the bed file doesn't have a score in it.

wig/wig.go
craiglowe: This would check to see if the value stored as a float can be stored equally well as an integer.  If so, it would visualize/print the number as an integer, which would not influence any downstream math.  It would just make the file size smaller in many cases by not printing a bunch of trailing zeros.

craiglowe (approved):
I saw your slack message about pushing updates to this branch / pull request, but I am not seeing those updates.  Example: I still see those for loops that set all the values to zero when they are already zero.

convert/convert.go
cshoben: I'll do a separate pull request with all these convert.go changes, including this one.

wig/wig.go
cshoben: Adding this to the list of convert.go changes.

PR #485, opened 2021-06-03
Remove defer File.Close() when running a file remove command. 

cshoben:
Instead of defer file close, explicitly close file before the file is removed. Otherwise, close file will run after remove file and recreate the file if there is anything left in the buffer (case by case basis)


RavenYL (approved):

craiglowe (approved):
Great work!

dasnellings (approved):

PR #488, opened 2021-06-07
Vcf Filter

dasnellings:
This PR adds some functionality to the vcfFilter command. I needed a way to test unknown tags so I have changed vcf filter to accept a `[]func(vcf.Vcf) bool` which contains all the testing functions. If all pass, then the record is included in the output. With this functionality, I was able to add 2 new flags `format` and `info` which work by passing them expressions as strings which are then parsed to create test functions. E.g. the string `"DP > 100"` is parsed to 
```
func(v vcf.Vcf) bool {
   if v.QueryInt("DP") > 100 {
      return true
   }
   return false
```

You can also do multiple tests (e.g. `"DP > 100 ; GQ >= 20 ; MQ == 10"`) which are all interpreted as AND. I tried to add support for more complex logic but it turned out to be quite difficult, so that is a TODO for now. 

One last note. I have been using this to process some of my exome data and it is FAST. Orders of magnitude faster than the GATK version (though I am sure GATK is doing a bit more under the hood than we are). I ran vcfFilter with `"DP > 100"` and it filtered a 750MB vcf in 12 seconds.


cmd/vcfFilter/expression.go
```diff
@@ -0,0 +1,255 @@
...
+		tag = strings.Trim(tagValuePair[0], " ")
+		if len(tagValuePair) == 2 {
+			value = strings.Trim(tagValuePair[1], " ")
+		}
+		answer = append(answer, getRelationshipTest(tag, value, op, header, isFormatElseInfo, includeMissingInfo))
+	}
+	return answer
+}
+
+func searchOp(exp string) operator {
```
rimangan, line 47: Wondering if these functions should be housed in another package. Logical operations might be broadly applicable for other cmds

rimangan: Particularly with the TestInteger-like functions

cmd/vcfFilter/vcfFilter.go
```diff
@@ -102,5 +285,7 @@ func main() {
 	infile := flag.Arg(0)
 	outfile := flag.Arg(1)
 
-	vcfFilter(infile, outfile, *groupFile, *chrom, *minPos, *maxPos, *ref, *alt, *minQual, *biAllelicOnly, *substitutionsOnly, *segregatingSitesOnly, *removeNoAncestor, *onlyPolarizableAncestors)
+	total, removed := vcfFilter(infile, outfile, c, *groupFile, parseFormat, parseInfo)
+	log.Printf("Processed  %d variants\n", total)
```
rimangan, line 289: Are these debug prints? Should be removed?

rimangan (dismissed):
An imaginative new way of representing multiple logical tests for filtering, very exciting! I think this gives us a lot of flexibility for expanding vcfFilter to many types of filtering operations. My one major question is whether this style can be expanded to other types of filtering cmds and whether some of the core infrastructure for parsing logical operators is general enough to belong in its own package. Just possible suggestions, I'm approving now in its current form

vcf/queryInfo_test.go
```diff
@@ -200,10 +200,13 @@ func TestQueryInt(t *testing.T) {
 	v := ParseFormat(data[0], header)
 	v = ParseInfo(v, header)
 
-	if !equalInt(QueryInt(v, header.Info["InfoA"].Key), expectedInfo["InfoA"]) {
+	val, found := QueryInt(v, header.Info["InfoA"].Key)
+	if !found || !equalInt(val, expectedInfo["InfoA"]) {
 		t.Errorf("troble querying read info")
```
cshoben, line 205: Are all these error messages supposed to say "trouble querying..."?

cshoben (approved):
Looks great!

cmd/vcfFilter/expression.go
dasnellings: At first this was going to be in a package, but I ended up dropping a lot of planned features mostly relating to complex logic (AND/OR). I decided to put the dumbed-down version in the cmd until I had the time to make a full version. But, I an open to moving this to a package as-is if people feel strongly.

cmd/vcfFilter/vcfFilter.go
dasnellings: I found myself constantly calculating this with `wc -l` after each run so I figured it would be nice to have this summary after it finished running. If that's not the style we are going for in gonomics then I'll go ahead and remove them, but I think it could be useful for a number of commands. Also, it prints to  stderr so it shouldn't cause problems with piping in unix.

vcf/queryInfo_test.go
dasnellings: nice catch. The dangers of copypasta

cshoben (approved):

cmd/vcfFilter/vcfFilter.go
rimangan: ok the piping was my only major concern. Going to approve

rimangan (approved):

cmd/vcfFilter/expression.go
```diff
@@ -0,0 +1,255 @@
...
+		if err != nil {
+			log.Fatalf("Error: value '%s' is not an integer as expected for tag '%s'.", value, tag)
+		}
+		return func(v vcf.Vcf) bool {
+			var answer bool = true
+			queryResult, found := vcf.QueryInt(v, tagKey)
+			if !found {
+				return includeMissingInfo
+			}
+			for _, recordVal := range queryResult[0] {
```
craiglowe, line 100: I think that here we are looping over all values of the tag and since each needs to pass, we are doing an "and" over those tests.  I think this is probably the most useful test, but with only a few (small?) changes I think we could pass a flag that would do an "or" over them.

cmd/vcfFilter/vcfFilter.go
craiglowe: Maybe a cool option would be to have a "verbose" mode/flag that would print these stats?

vcf/queryInfo.go
```diff
@@ -109,28 +109,30 @@ func parseValue(v Vcf, s []string, k Key) interface{} {
...
 
-	case typeString:
+	case String:
 		data := make([][]string, len(s))
 		for i := range s {
-			stringValues = strings.Split(s[i], ",")
-			data[i] = stringValues
+			//stringValues = strings.Split(s[i], ",")
+			//data[i] = stringValues
+			data[i] = []string{s[i]} // many fields ignore specifications and use literal commas
```
craiglowe, line 123: I don't totally understand this change and comment.

vcf/queryInfo.go
```diff
@@ -109,28 +109,30 @@ func parseValue(v Vcf, s []string, k Key) interface{} {
...
 		}
 
-		for i := range data {
-			if !checkNumber(v, k, len(data[i])) {
-				log.Panicf("unexpected number of values")
-			}
-		}
+		// Illegal string fields are common, so we will suppress this check.
```
craiglowe, line 126: Is there a common thing that happens we could allow, but still have some checks?  What kind of stuff commonly happens in files?  Does this make it hard to apply the filter criteria that I user may specify?

vcf/queryInfo.go
```diff
@@ -195,18 +197,23 @@ func checkNumber(v Vcf, k Key, length int) bool {
 // The return is a slice of slices where the first slice corresponds to the sample
 // (this is always len == 1 when querying the Info field) and the second slice corresponds
 // to multiple values that may be present for the given tag (e.g. ref/alt read depth may be "9,1").
-func QueryInt(v Vcf, k Key) [][]int {
-	if k.dataType != typeInteger {
-		log.Panicf("requested QueryInt but key records data type as '%s'", k.dataType)
+//
+// The second return is false if the requested value is not present in the input record.
```
craiglowe, line 201: Is the second return value encoding a single common error case in a way?

craiglowe (approved):
This is really wonderful!  I am excited to hear more about this at code review.  I would especially enjoy a short overview of the capabilities and to hear a little about the speed testing.

PR #491, opened 2021-06-10
Bed Strand Fix

rimangan:
In the BED file format documentation on UCSC, the 'strand' field can be either '+', '-', or '.' to signify a region of the positive strand, negative strand, or with no strand information, respectively. Our code base has represented this quantity with a bool, with true representing + and false representing -. However, I have been running into issues where our parser is unable to handle bed data with a '.' in the strand field, even though this is a valid BED format file. Therefore, I've changed Strand to a Byte constant instead of a bool to accommodate this third state.

### DIFF
--- a/bed/bed.go
+++ b/bed/bed.go
@@ -27,7 +27,7 @@ type Bed struct {
 }
 
 //Strand stores strand state, which can be positive, negative, or none.
-type Strand rune
+type Strand byte
 
 const (
 	Positive Strand = '+'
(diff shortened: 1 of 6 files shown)


craiglowe (commented):
This looks good!  I see this is a draft pull request.  Should we discuss this at code review?

rimangan:
> This looks good! I see this is a draft pull request. Should we discuss this at code review?

yes we should discuss. Something still does not compile and I'm not sure where the error is.

cshoben (approved):

RavenYL (approved):

craiglowe (approved):
Excellent!

julianaccarvalho (approved):

PR #492, opened 2021-06-15
bgzf reader

dasnellings:
Implements a reader for the bgzf format as defined in https://raw.githubusercontent.com/samtools/hts-specs/master/SAMv1.pdf

Writer coming soon. Eventually a bam reader.


bgzf/bgzf.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	r.buf = bufio.NewReader(r.file) // for ReadByte method
+
+	r.decompressor, err = gzip.NewReader(r.buf)
+	if err != nil {
+		log.Panic(err)
+	}
+
+	r.decompressor.Multistream(false)
```
craiglowe, line 62: Is this needed here?  I see it also at the start of ReadBlock.  I am wondering if because it is always called at the beginning of ReadBlock, maybe it doesn't need to be called here.

craiglowe (approved):
Very cool!  Two small questions/comments.

bgzf/bgzf.go
```diff
@@ -0,0 +1,108 @@
...
+type Reader struct {
+	file         *os.File
+	buf          *bufio.Reader
+	decompressor *gzip.Reader
+}
+
+// Block represents a single BGZF block stores as a bytes.Buffer
+// initialized to have a capacity of 65536 (max uint16).
+type Block struct {
+	bytes.Buffer // cap == 2^16
```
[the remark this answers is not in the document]
dasnellings: I need Block to have access to the methods of bytes.Buffer. I tried doing `type Block bytes.Buffer` but that does not preserve methods. Embedding was the way to inherit methods.

bgzf/bgzf.go
dasnellings: You are correct. I will remove that. The multistream was a pain to figure out and I think I got a bit trigger-happy with it.

craiglowe (approved):
Wonderful!

RavenYL (approved):

cshoben (approved):
Havent messed with bgzf files before but this seems extremely useful as I've read a bit more about them. Thanks for adding this into the gonomics toolkit!

PR #493, opened 2021-06-16
weakToStrong option for vcfFilter

rimangan:
Functionality to filter "weakToStrong" variants from Vcf data (i.e. AT to GC). Passes tests.


cmd/vcfFilter/testdata/expected_weakToStrongOnly.vcf
```diff
@@ -0,0 +1,4 @@
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	First	Second
```
craiglowe, line 1: I see this vcf file and the others always listing the ancestral allele (AA) as T, even when T is not a ref or alt allele.

vcf/filter.go
```diff
@@ -209,6 +209,32 @@ func IsPolarizable(v Vcf) bool {
 	return true
 }
 
+//IsWeakToStrong returns true if an input biallelic substitution variant is a weak to strong variant, false otherwise.
+func IsWeakToStrong(v Vcf) bool {
+	if !IsBiallelic(v) || !IsSubstitution(v) {
+		return false
+	}
+	if v.Ref == "A" || v.Ref == "T" {
```
craiglowe, line 217: Don't we also need to check the ancestral allele in this function?  Seems like we need the ancestral allele to be A/T and the derived allele (could be ref or alt) to be G/C.

craiglowe (commented):
It will be exciting to see the results from this!

cmd/vcfFilter/testdata/expected_weakToStrongOnly.vcf
rimangan: This is a copy paste error I can fix on the next commit.

vcf/filter.go
rimangan: This could go either way. If we are worried that we will be biased for weak to strong mutations in the derived state, we can overestimate positive selection with high frequency WtS mutants. But conversely, if there is a corresponding bias against StW variants, we may also want to remove those or risk overestimating negative selection. I'm not sure what's best

craiglowe: I understand.  While this may be what you want for your research question.  I think it might cause confusion if someone has their G->A mutations removed when they say they want to remove "weak to strong" mutations.  Maybe we want to change the name, or to implement both IsWeakToStrong and IsStrongToWeak and then you will give both those flags for your research.

craiglowe (commented):
Comment added

vcf/filter.go
cshoben: If what I'm inferring about the definition/use of "strong to weak" etc to be true, then I do agree that it would make sense to me to break it up. I'm thinking lines 217 to 221 are referring to weak to strong and lines 222 to 228 are strong to weak... assuming that ref is ancestral since there is directionality in the way this is interpreted? Assuming ref --> alt 
But if we think of there being bidirectionality, then how Riley has it makes sense to me. Lines 222 thought 226 would be flagging to G->A as a weak to strong (by backwards). 
However, I do think I preference it being broken up into two separate functions.

cmd/vcfFilter/testdata/expected_weakToStrongOnly.vcf
```diff
@@ -0,0 +1,4 @@
...
+chr3	30	IsStrongToWeak	G	T	100	PASS	AA=T	GT:DP	0|1:30	1|1:30
```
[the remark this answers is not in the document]
dasnellings: I think its because G-C pairs have 3 hydrogen bonds (strong) whereas A-T pairs only have 2 (weak)

cmd/vcfFilter/vcfFilter.go
```diff
@@ -235,6 +244,8 @@ func main() {
 	var segregatingSitesOnly *bool = flag.Bool("segregatingSitesOnly", false, "Retains only variants that are segregating in at least one sample.")
 	var removeNoAncestor *bool = flag.Bool("removeNoAncestor", false, "Retains only variants with an ancestor allele annotated in the info column.")
 	var onlyPolarizableAncestors *bool = flag.Bool("onlyPolarizableAncestors", false, "Retains only variants that can be used to construct a derived allele frequency spectrum. Must have a subsitution where the ancestral allele matches either alt or ref.")
+	var weakToStrongOnly *bool = flag.Bool("weakToStrongOnly", false, "Retains only variants that are weak to strong mutations.")
+	var noWeakToStrong *bool = flag.Bool("noWeakToStrong", false, "Removes weak to strong variants.")
```
dasnellings, line 248: These flag descriptions are pretty ambiguous. More explanation would be helpful (even just a G/C <-> A/T)

vcf/filter.go
```diff
@@ -209,6 +209,32 @@ func IsPolarizable(v Vcf) bool {
 	return true
 }
 
+//IsWeakToStrong returns true if an input biallelic substitution variant is a weak to strong variant, false otherwise.
+func IsWeakToStrong(v Vcf) bool {
```
dasnellings, line 213: I disagree with the name of this function. WeakToStrong implies Ref = A/T, Alt = C/G. What you are really testing is whether there is a A/T to C/G change regardless of ref/alt status. I think this needs to be either renamed into something more explicit (I think of `ChangesBondStrength` but that is admittedly not great) or make both a `WeakToStrong` and a `StrongToWeak`.

vcf/filter.go
dasnellings: I agree with Craig and Chelsea. Either change the name or make 2 separate functions.

rimangan: Name changed in most recent commit.

vcf/filter.go
rimangan: Fixed.

vcf/filter.go
craiglowe: I don't see this name change reflected here.  Maybe that commit still needs a push to github?

craiglowe (commented):
From the comments it sounds like you have addressed everything, but I don't see the changes on github

vcf/filter.go
rimangan: The name here hasn't changed. This function is actually just "WeakToStrong" and a new function has been added to cover the case "WeakToStrongOrStrongToWeak". The most recent commit returned the "strongtoweak" portion of this function and implemented that elsewhere. Should be in line with the comments in the current commit

cmd/vcfFilter/vcfFilter.go
```diff
@@ -92,21 +92,27 @@ func usage() {
...
+	minQual                        float64
+	ref                            string
+	alt                            []string
+	biAllelicOnly                  bool
+	substitutionsOnly              bool
+	segregatingSitesOnly           bool
+	removeNoAncestor               bool
+	onlyPolarizableAncestors       bool
+	weakToStrongOrStrongToWeakOnly bool
+	noWeakToStrongOrStrongToWeak   bool
```
RavenYL, line 108: Are these redundant to the lines below (refWeakAltStrongOnly, refStrongAltWeakOnly, notRefWeakAltStrong, notRefStrongAltWeak)?

RavenYL (approved):

craiglowe (approved):
Great!

dasnellings (dismissed):

craiglowe (approved):
Great work!

RavenYL (approved):

dasnellings (approved):

PR #494, opened 2021-06-16
bgzf writer

dasnellings:
Dependent on my bgzf reader PR.

Implements writing bgzf files using an underlying gzip writer.


bgzf/writer.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// NewWriter creates a bgzf writer from any input writer.
+func NewWriter(w io.Writer) Writer {
+	var err error
+	var zw Writer
+	zw.w = w
+	zw.zipBlock = NewBlock()
+	zw.compressor = gzip.NewWriter(zw.zipBlock)
+	_, err = zw.compressor.Write([]byte{}) // write header
```
craiglowe, line 30: I don't totally understand writing the gzip header here.

bgzf/writer.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		log.Panic(err)
+	}
+	zw.zipBlock.Reset()
```
craiglowe, line 34: I see the header write and the reset both here when initializing the writer and then again on Write commands.  Do we need them here?

bgzf/writer.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// magicEOF generates the magic EOF byte slice from the hex const.
+func magicEOF() []byte {
+	// empty gzip block per sam specs
+	answer, err := hex.DecodeString(magicHexEOF)
+	if err != nil {
+		log.Panic(err)
+	}
+	return answer
```
craiglowe, line 115: Should we just store this return value as the const at the top of the file?

craiglowe (approved):
Great work!  I included a couple questions/comments.

bgzf/writer.go
dasnellings: I think I added this here before I trashed the header on every write. This is no longer needed now that it is in the Write function and has been removed.

bgzf/writer.go
dasnellings: I think I added this here before I trashed the header on every write. This is no longer needed now that it is in the Write function and has been removed.

bgzf/writer.go
dasnellings: Good idea. I've made it a private var.

craiglowe (approved):
Wonderful!

RavenYL (approved):

bgzf/writer.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// writeHeader for custom bgzf header.
+func (w Writer) writeHeader(compSize int) {
+	var header [18]byte
+	header[0] = 31  // gzip ID 1
+	header[1] = 139 // gzip ID 2
+	header[2] = 8   // gzip compression method
+	header[3] = 4   // gzip flags
+	// header[4:8] = mod time uint32
```
cshoben, line 85: Does this mean that headers 4 through 8 are blank or skipped? I'm trying ti visualize this in my head.

cshoben (approved):
I appreciate these being two separate PR. Just a conceptual question :)

bgzf/writer.go
dasnellings: This means that 4-8 are set to 0 (the default value). I initialize the array to have a specific size where everything is set to zero by default. If I want any values to stay zero, I can just ignore them. I just leave the comment here for clarity.

PR #495, opened 2021-06-17
SelectionMLE first draft

rimangan:
SelectionMLE is a new cmd that performs maximum likelihood estimation (MLE) for natural selection based on the derived allele frequency spectrum from an input set of variants in a VCF file. This cmd incorporates much of the functionality of our Bayesian model, including divergence-based ascertainment bias correction, integralAccuracy control, and flexible user input for search space bounds and error thresholds, but performs MLE orders of magnitude faster than MCMC. Lacks the statistical rigor of MCMC, but great for getting a first look at data.
In this model, all bases have the same selection parameter, which I think ensures that our posterior distribution will be unimodal. Thus, we can use a unimodal maximization search algorithm to identify the peak and return that answer to a file.
I've chosen to use Golden-Section Search as our MLE algorithm, which has much less complexity than gradient-based methods like BFGS. While GSS is restricted to unimodal functions, it is very fast and minimizes the number of function calls. 
I've attached a YouTube link below to a 10 minute video explaining the GSS algorithm.
https://www.youtube.com/watch?v=wpGN2xus75w


cmd/selectionMLE/tmp.txt
```diff
@@ -0,0 +1,2 @@
+#FILENAME	MaximumLikelihood
```
craiglowe, line 1: Should this be in testdata?

popgen/stationarity.go
```diff
@@ -176,3 +176,34 @@ func AfsLikelihood(afs Afs, alpha []float64, binomMap [][]float64, integralError
 	}
 	return answer
 }
+
+//AfsLikelihoodFixedAlpha calculates the likelihood of observing a particular frequency spectrum for a given alpha, a selection parameter.
+//This represents the special case where every segregating site has the same value for selection, which enables faster, more simplified calculation.
+func AfsLikelihoodFixedAlpha(afs Afs, alpha float64, binomMap [][]float64, integralError float64) float64 {
```
craiglowe, line 182: Is the idea with these 3 new functions that since alpha is fixed we can do lots of caching?

craiglowe (approved):
Great!  Excited to see the results!

popgen/stationarity.go
rimangan: Yes! Especially for sets like the UCEs, where half of the variants have a dAF of 1/n, we can save a substantial amount of runtime by caching probabilities.

cmd/selectionMLE/tmp.txt
rimangan: I commented out the os.Remove while debugging, will fix on next commit

RavenYL (approved):

numbers/search.go
```diff
@@ -0,0 +1,58 @@
+package numbers
```
dasnellings, line 1: I don't think this is needed now, but we are quickly reaching critical mass where it may made sense to split into `numbers` and `stats`

dasnellings (approved):
Looks good to me, though most of the stats goes over my head.

PR #496, opened 2021-06-17
bai reader

dasnellings:
More infrastructure working towards a bam reader. Dependent on my last 2 bgzf PRs so you can ignore all the changes in the bgzf files. 

bai is the index format for bam files (the output of `samtools index *.bam`). The concept of the bai file is a bit involved so it might be good to discuss this at code review if people are interested.


sam/bai.go
```diff
@@ -0,0 +1,295 @@
...
+// magicBai is a 4 byte sequence at the start of a bai file
+const magicBai string = "BAI\u0001"
+
+// Bai is an index of the BAM format that allows for efficient
+// random access of reads given a query genomic range. The Bai
+// struct should not be accessed directly, but used through
+// query functions present within the sam package of gonomics.
+type Bai struct {
+	refs []reference
+
```
craiglowe, line 22: blank line?

sam/bai.go
```diff
@@ -0,0 +1,295 @@
+package sam
```
craiglowe, line 1: Everything looks good after quick pass, but I am having trouble doing an in-depth review without knowing the specs for the file format.  Can you include a comment at the top of the file to the version/document of the specs you used?

sam/bai.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	for i := range bai.refs {
+		bai.refs[i] = annotateBinRanges(bai.refs[i])
+		bai.refs[i] = assembleTree(bai.refs[i])
+	}
+	return bai
+}
+
+// alias for ease of use and readability
```
craiglowe, line 100: This might make it easier for people who are reading the entire file top-to-bottom, but maybe it makes it harder for people who get an error and jump to a specific function/line to see what is going on (e.g. parseBin).  Then they need to either scroll around or use an smart IDE feature to figure out what "le" refers to.

craiglowe (commented):
Looks great!  Can you put a link/reference to the file format at the top?  This way we know what version it works for.  In the future, this may help disentangle a bug from simply not having new features/changes implemented.

sam/bai.go
dasnellings: I have added a ref to the document.

sam/bai.go
dasnellings: It makes things much nicer to read imo, but I do agree with you. How do you feel about this compromise? I've moved the alias to the top of the file, and added a note to the first instance of `le` in each function so others would only need to look to the first use of `le` like they would for any variable declaration.

rimangan (approved):
Would love to chat one on one for more info about the BAI format, but happy to approve

cshoben (approved):

craiglowe (approved):
Great work!

PR #497, opened 2021-06-18
bedEnrichments

rimangan:
bedEnrichments is a new cmd in gonomics that takes in two bed files, representing sets of regions in the genome, and a 'noGap' bed file containing all possible ungapped positions in the genome, and calculates whether they are more likely to have overlapping elements than would be expected if the elements in set 2 were distributed randomly throughout the genome. Ported from C code written by Craig.
Passes tests.


rimangan:
Added a "DO NOT MERGE" tag while we figure out some performance issues.

rimangan:
Approximate method is from this thread: https://stats.stackexchange.com/questions/9510/probability-distribution-for-different-probabilities

bed/enrichment.go
```diff
@@ -0,0 +1,125 @@
...
+	subtractFromBedCoord(noGapRegions, 0, length-1, tempNoGap)
+	return float64(OverlapLengthSum(tempElements, tempNoGap)) / float64(TotalSize(tempNoGap))
+}
+
+func ElementOverlapProbabilities(elements1 []*Bed, elements2 []*Bed, noGapRegions []*Bed) []float64 {
+	var tempElements1 []*Bed = make([]*Bed, len(elements1))
+	var tempElements2 []*Bed = make([]*Bed, len(elements2))
+	var tempNoGap []*Bed = make([]*Bed, len(noGapRegions))
+	var answer []float64 = make([]float64, len(elements2))
+	copy(tempElements1, elements1)
```
craiglowe, line 40: I notice that these copies need to be not only on the pointers, but on the data/structs themselves.  The way it is coded now, we are modifying the original data.

craiglowe (commented):
One quick comment on copying

cshoben (approved):

bed/info_test.go
```diff
@@ -0,0 +1,43 @@
...
+	var b []Bed
+	var actual bool
+	for _, v := range SelfOverlappingTests {
+		b = Read(v.filename)
+		actual = IsSelfOverlapping(b)
+		if actual != v.expected {
+			t.Errorf("Error in IsSelfOverlapping. Expected: %t. Actual: %t.", v.expected, actual)
+		}
+	}
+}
```
craiglowe, line 43: made it partway through.  Saving my spot

craiglowe (commented):
Looks like great work.  I only made it partway through, but putting my comments up now

cmd/bedEnrichments/bedEnrichments.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if verbose > 0 {
+		log.Println("Reading bed files.")
+	}
+
+	elementsOne := bed.ReadLite(inFile)
+	elementsTwo := bed.ReadLite(secondFile)
+	noGapRegions := bed.ReadLite(noGapFile)
+
+	if trimToRefGenome {
```
craiglowe, line 28: This will likely be easier to both read, debug, and test if it were a separate function

cmd/bedEnrichments/bedEnrichments.go
```diff
@@ -0,0 +1,160 @@
...
+		var overlap1, overlap2 []interval.Interval
+		var trimmedE1 []bed.Bed = make([]bed.Bed, 0)
+		var trimmedE2 []bed.Bed = make([]bed.Bed, 0)
+
+		var e1Intervals []interval.Interval
```
craiglowe, line 33: These could be function calls here too, to go from a slice of beds to an interval tree

cmd/bedEnrichments/bedEnrichments.go
```diff
@@ -0,0 +1,160 @@
...
+		tree1 := interval.BuildTree(e1Intervals)
+		var e2Intervals []interval.Interval
+		for i := range elementsTwo {
+			e2Intervals = append(e2Intervals, elementsTwo[i])
+		}
+		tree2 := interval.BuildTree(e2Intervals)
+
+		for i := range noGapRegions {
+			overlap1 = interval.Query(tree1, noGapRegions[i], "within")
+			if overlap1 != nil {
```
craiglowe, line 46: Can be removed.  It is ok to range over a nil list.

cmd/bedEnrichments/bedEnrichments.go
```diff
@@ -0,0 +1,160 @@
...
+				for j := range overlap1 {
+					trimmedE1 = append(trimmedE1, overlap1[j].(bed.Bed))
+				}
+			}
+			overlap2 = interval.Query(tree2, noGapRegions[i], "within")
+			if overlap2 != nil {
```
craiglowe, line 52: Can be removed.  ok to range over nil list

cmd/bedEnrichments/bedEnrichments.go
```diff
@@ -0,0 +1,160 @@
...
+	_, err = fmt.Fprintf(out, "%s\t%s\t%s\t%d\t%d\t%d\t%f\t%f\t%f\t%e\n", method, inFile, secondFile, len(elementsOne), len(elementsTwo), overlapCount, summarySlice[0], summarySlice[1], float64(overlapCount)/summarySlice[1], summarySlice[2])
+	exception.PanicOnErr(err)
+
+	err = out.Close()
+	exception.PanicOnErr(err)
+}
+
+func usage() {
+	fmt.Print(
+		"bedEnrichments - Returns the p-value of enrichment for overlaps between the elements in two input bed files.\n" +
```
craiglowe, line 123: I think we probably need a sentence about the asymmetry of sets one and two.

cmd/bedEnrichments/bedEnrichments.go
```diff
@@ -0,0 +1,160 @@
...
+	var summarySlice []float64
+	overlapCount := bed.OverlapCount(elementsOne, elementsTwo)
+
+	if verbose > 0 {
+		log.Println("Calculating enrichment probabilities.")
+	}
+
+	switch method {
+	case "exact":
+		probs := bed.ElementOverlapProbabilities(elementsOne, elementsTwo, noGapRegions)
```
craiglowe, line 94: I think we need to double-check these function calls.  I am worried that the above call to OverlapCount gives us a max of len(elementsOne) as the number of overlaps, while the call to ElementOverlapProbabilities is designed such that it is counting the number of elementsTwo that overlap anything in elementsOne.  Is that correct?  If so, we could just swap the args to OverlapCount

craiglowe (commented):
Great work!  Had a few small comments

PR #498, opened 2021-06-18
bam reader

dasnellings:
The bam reader is (finally) here. One important TODO is a parser for the extra field. Parsing to a string as is in the current sam format would be rather wasteful, so it is stored in the unexported `unparsedExtra` field. The idea is that a future function can parse this if and only if the user needs to access these fields (similar to what we have for vcf). Read-write fidelity will not be possible until this parse is complete so this will be high on my priority list. I would also like to make a multithreaded reader using a bam index.

Just for grins I did some benchmarks against samtools and the biogo bam reader. Faster than biogo, but Heng Li is unstoppable... It's pretty serviceable, but I'd love to hear any ideas on ways to get a speed boost. Also big asterisk that we are not parsing the extra field here, so it wouldn't be unreasonable to add an extra 10-15%(?) to the gonomics runtime. 

1k reads:
```
BenchmarkGonomicsBamRead-8           314           3671723 ns/op         1409264 B/op      21896 allocs/op
BenchmarkBiogoBamRead-8               96          11181032 ns/op         4662195 B/op      73633 allocs/op
BenchmarkSamtoolsBamRead-8           182           6460661 ns/op     
```

10k reads:
```
BenchmarkGonomicsBamRead-8            27          37127673 ns/op        13489844 B/op     221656 allocs/op
BenchmarkBiogoBamRead-8                9         112586681 ns/op        41579128 B/op     783714 allocs/op
BenchmarkSamtoolsBamRead-8            80          14798673 ns/op         
```

1m reads:
```
BenchmarkGonomicsBamRead-8             1        3823531000 ns/op        1356606584 B/op 22761097 allocs/op
BenchmarkBiogoBamRead-8                1        11586829916 ns/op       4104614664 B/op 78393247 allocs/op
BenchmarkSamtoolsBamRead-8             1        1014599708 ns/op   

```


cshoben (approved):

sam/tag.go
```diff
@@ -0,0 +1,307 @@
...
+//
+// The returned value is an interface that may store any of: uint8, int8,
+// uint16, int16, uint32, int32, float32, string, or a slice of interface{}
+// with any of the previous underlying types.
+// It is the callers responsibility to know/determine the type
+// and perform the assertion.
+//
+// The second return is false if the requested tag was not found.
+func QueryTag(s Sam, t string) (value interface{}, found bool, error error) {
+	if s.parsedExtra == nil {
```
edotau, line 26: Just a commenting that the tags should probably be re-used in the giraf package as well

sam/bamRead_test.go
```diff
@@ -0,0 +1,202 @@
...
+			if err == io.EOF {
+				break
+			}
+			fmt.Fprint(ioutil.Discard, s)
+		}
+		r.Close()
+	}
+}
+
+func BenchmarkBiogoBamRead(b *testing.B) {
```
edotau, line 158: Another cool bam reader I found is from grail: [https://github.com/grailbio/bio/tree/master/encoding/bam](https://github.com/grailbio/bio/tree/master/encoding/bam). I'd be interested to see how your benchmarks compare to this implementation

edotau (approved):
Good work dan I enjoyed looking at this. LGTM

sam/bamRead_test.go
dasnellings: Nice! I didn't know Grail had one. I ran a couple benchmarks to check it out (grail func below).


For a printing task they they perform pretty comparably in terms of speed (gonomics looks to be marginally faster), but grail has ~50% fewer allocs/op
```
BenchmarkGonomicsBamRead-8            10         101756525 ns/op        37637304 B/op    1115598 allocs/op
BenchmarkGrailBamRead-8               10         107098446 ns/op        33841868 B/op     616302 allocs/op
```

For a straight reading task (just removed `fmt.Fprintf(ioutil.Discard, read)`), grail wins by a pretty substantial margin on time. Gonomics has slightly fewer allocs/op, but uses significantly less memory (2.6MB vs 15.2MB).

```
BenchmarkGonomicsBamRead-8            58          20298426 ns/op         2623551 B/op      33174 allocs/op
BenchmarkGrailBamRead-8               82          14090170 ns/op        15241928 B/op      36918 allocs/op
```

One potential saving grace is that on my machine, the grail reader uses 8 threads for reading whereas the gonomics reader currently uses only 1 thread. I will need to add a concurrent reader and see if we get any performance bump. 

Not familiar with how to operate this package so I copied a func from one of their benchmarks. Let me know if you think this should be change to make it a better test.
```
func BenchmarkGrailBamRead(b *testing.B) {
	for i := 0; i < b.N; i++ {
		provider := bamprovider.NewProvider(bigBam)
		iters, err := bamprovider.NewPairIterators(provider, true)
		if err != nil {
			b.Error(err)
		}
		var waitGroup sync.WaitGroup
		waitGroup.Add(len(iters))
		for i, iter := range iters {
			go func(id int, iter *bamprovider.PairIterator) {
				for iter.Scan() {
					record := iter.Record()
					fmt.Fprint(ioutil.Discard, record.R1)
					fmt.Fprint(ioutil.Discard, record.R2)
				}
				waitGroup.Done()
			}(i, iter)
		}
		waitGroup.Wait()
		if err := provider.Close(); err != nil {
			b.Error(err)
		}
	}
}
```

edotau: Awesome, I'll take a closer look later today. I forget what method/function it is, but I know there's a way to set a cap to number of threads used during the benchmarks.

cshoben (approved):

rimangan (approved):

RavenYL (approved):

PR #499, opened 2021-06-19
Remove tmp file from testing in selectionMLE

rimangan:
A tmp.txt file created during testing selectionMLE is supposed to be deleted, but the main branch version has this deletion commented out from my debugging. Corrected here.


craiglowe (approved):
Good Catch!

dasnellings (approved):

RavenYL (approved):

PR #500, opened 2021-06-21
Raven vcf filter rsID

RavenYL:
- Added an rsID filter, for a single string this time, to vcfFilter.
- The changes pass cmd/vcfFilter's go test, and go build and run worked.
- I hope to expand this rsID filter to accept []string in the future.


dasnellings (approved):

julianaccarvalho (approved):

vcf/filter.go
```diff
@@ -40,6 +41,10 @@ func Filter(v Vcf, chrom string, minPos int, maxPos int, ref string, alt []strin
 	if onlyPolarizableAncestors && !IsPolarizable(v) {
 		return false
 	}
+	//raven's note: added a check for id
+	if !FilterId(v, id) {
```
craiglowe, line 45: I might consider the style of the above "if" statements above where you first check to see if a filter exists and then apply the filter (example: if id != "" && IsIdMatch(v, id))
Right now "FilterId" has the check (id != "") hidden in it, where it might not be obvious from the documentation that the blank string is a wildcard that will match everything.

craiglowe (approved):
Great!  I approved, but had one minor comment

vcf/filter.go
RavenYL: I just made this change in my new branch. Thank you!

PR #501, opened 2021-06-22
Remove bed pointers

dasnellings:


rimangan (approved):

bed/bed.go
```diff
@@ -62,64 +62,65 @@ func ToString(bunk *Bed, fields int) string {
...
 }
 
-//WriteBed writes an input Bed struct to an os.File with a specified number of Bed fields.
-func WriteBed(file *os.File, input *Bed, fields int) {
+// WriteBed writes an input Bed struct to an io.Writer with a specified number of Bed fields.
+func WriteBed(file io.Writer, input Bed) {
 	var err error
-	_, err = fmt.Fprintf(file, "%s\n", ToString(input, fields))
-	common.ExitIfError(err)
+	_, err = fmt.Fprintf(file, "%s\n", input)
```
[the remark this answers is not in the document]
RavenYL: It seems like we can still control the number of fields with "FieldsInialitized int", looking at bed_test.go?

RavenYL (approved):

edotau (approved):
LGTM!

bed/methods.go
```diff
@@ -46,24 +46,27 @@ func (b *BedSlice) Pop() interface{} {
 }
 
 func (b BedSlice) Write(file string) {
-	Write(file, b, 7)
+	a := make([]Bed, len(b))
```
craiglowe, line 49: It catches my eye that we are making a copy of the slice here, which could be large.  Is it worth maintaining a WritePtr as well as a Write to avoid this copy?  I am thinking that BedSlice is used by the overlap program?

bed/methods.go
```diff
@@ -46,24 +46,27 @@ func (b *BedSlice) Pop() interface{} {
...
-	WriteToFileHandle(file, b, b.FieldsInitialized) //adaptive field writing seems most flexible for the method
+	WriteToFileHandle(file, *b) //adaptive field writing seems most flexible for the method
 }
 
 func (b *Bed) NextRealRecord(file *fileio.EasyReader) bool {
 	var done bool
-	var next *Bed
-	for next == nil && !done {
+	var next Bed
+	for next.Chrom == "" && !done {
```
craiglowe, line 63: Is it more intuitive to check for FieldsInitialized being zero?

go.mod
```diff
@@ -4,5 +4,6 @@ go 1.15
 
 require (
 	github.com/biogo/hts v1.2.2
+	github.com/edotau/goFish v0.0.0-20210617055710-9385c85e5fc5
```
craiglowe, line 7: I didn't see added imports, but maybe I missed them?

go.sum
```diff
@@ -1,31 +1,498 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
craiglowe, line 1: Lots of dependencies being added here that I didn't expect from removing pointers.

craiglowe (commented):
Great work!  I had a few small comments and was wondering why we are adding so many dependencies just by removing a bunch of bed pointers.

go.sum
edotau: good catch maybe the go.mod go.sum files were updated by mistake? running `go mod tidy` and rechecking in the `go.mod go.sum` files should resolve this.

PR #504, opened 2021-06-24
Bed merge upgrade

rimangan:
bedMerge, a cmd that accepts a bed file and merges overlapping entries, previously required presorting. Now the cmd simply sorts the input bed file. Also went ahead and added testing for this cmd.


rimangan:
This branch is more up to date than the bedEnrichments PR, and includes all of that code, so now let's plan on using this PR as a candidate to merge.

bed/compare.go
```diff
@@ -68,6 +116,39 @@ func Compare(a Bed, b Bed) int {
...
+	}
+	if a.ChromEnd > b.ChromEnd {
+		return 1
+	}
+	return 0
+}
+
+//CompareChromEndByChrom compares beds by chromosome and then by chromEnd, but not by chromStart.
+func CompareChromEndByChrom(a Bed, b Bed) int {
+	chromComp := strings.Compare(a.Chrom, b.Chrom)
```
dasnellings, line 145: We should not be using strings.Compare. `a.Chrom == b.Chrom`

bed/compare.go
```diff
@@ -68,6 +116,39 @@ func Compare(a Bed, b Bed) int {
...
+		return -1
+	}
+	if sizeA > sizeB {
+		return 1
+	}
+	return 0
+}
+
+//Compare returns zero for beds with an equal ChromEnd position and otherwise returns the ordering of the two Bed entries by chromEnd.
+func CompareChromEnd(a Bed, b Bed) int {
```
dasnellings, line 133: I think we should remove many of these compare functions. Or at least make them private. I would strongly argue that `if a.ChromEnd < b.ChromEnd {...}` is always clearer then `if CompareChromEnd(a, b) == -1 {...}`

bed/enrichment.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+//subtractFromBedCoord is a helper function of overlapProbability that subtracts the values subStart and subEnd from entries in a slice of Bed b.
+//bClone is a clone of b used to save on memory allocation.
+func subtractFromBedCoord(b []Bed, subStart int, subEnd int, bClone []Bed) {
+	var prevEnd int = 0
+	var prevChrom string = ""
+
+	for i := range b {
+		if strings.Compare(prevChrom, "") == 0 || strings.Compare(prevChrom, bClone[i].Chrom) != 0 {
```
dasnellings, line 17: We should really not be using strings.Compare. This should be 
`preChrom == "" || prevChrom != bClone[i].Chrom`

cshoben (approved):

RavenYL (approved):

bed/enrichment.go
```diff
@@ -0,0 +1,232 @@
...
+	var mu, sigma float64 = 0, 0
+
+	for i := range elementOverlapProbs {
+		mu += elementOverlapProbs[i]
+		sigma = sigma + elementOverlapProbs[i]*(1-elementOverlapProbs[i])
+	}
+
+	sigma = math.Sqrt(sigma) //sigma will represent the standard deviation for our normal approximation. the above sum gives us the variance.
+
+	answer[0] = 1.0
```
craiglowe, line 75: Do we need to return this, or can we just have two returns?

bed/enrichment.go
```diff
@@ -0,0 +1,232 @@
...
+	pValue := numbers.NormalDist(float64(overlapCount), mu, sigma)
+	for s := overlapCount + 1; s <= len(elementOverlapProbs); s++ {
+		pValue += numbers.NormalDist(float64(s), mu, sigma)
+	}
+	answer[2] = pValue
+
+	return answer
+}
+
+//EnrichmentPValueExact performs an exact calculation fo enrichment bgaased on a set of overlap probabilities and the observed overlap count.
```
craiglowe, line 88: Couple spelling typos

bed/enrichment.go
```diff
@@ -0,0 +1,232 @@
...
+//The exact method is non-polynomial, and is thus not recommended for large datasets.
+//Returns a slice of three values. The first is the debug check, the second is the expected number of overlaps, and the third is the pValue of the observed number of overlaps.
+func EnrichmentPValueExact(elementOverlapProbs []float64, overlapCount int) []float64 {
```
craiglowe, line 91: We can update this to have multiple returns, instead of a slice

bed/enrichment.go
```diff
@@ -0,0 +1,232 @@
...
+	pValue, underflow := numbers.BinomialDist(numTrials, overlapCount, prob)
+	if underflow {
+		answer[2] = 0.0
+		if verbose > 0 {
+			log.Println("Underflow detected in the binomial distribution at overlapCount. p value is too small to detect.")
+		}
+	} else {
+		for s := overlapCount + 1; s <= numTrials; s++ {
+			curr, underflow = numbers.BinomialDist(numTrials, s, prob)
+			if underflow {
```
craiglowe, line 198: I think we want the updated logic in UpperBound

bed/enrichment.go
```diff
@@ -0,0 +1,232 @@
...
+//EnrichmentPValueUpperBound, together with EnrichmentPValueLowerBound, provide a range of possible values for the pValue of overlap.
+//Returns a slice of three values. The first is the debug check, the second is the expected number of overlaps, and the third is the pValue of the observed number of overlaps.
+func EnrichmentPValueUpperBound(elements1 []Bed, elements2 []Bed, noGapRegions []Bed, overlapCount int, verbose int) []float64 {
+	var numTrials int = len(elements2)
+	var answer []float64 = make([]float64, 3)
+	var tempElements1 []Bed = make([]Bed, len(elements1))
+	copy(tempElements1, elements1)
+	var tempNoGap []Bed = make([]Bed, len(noGapRegions))
+	copy(tempNoGap, noGapRegions)
+	minElements2 := findLargestBedLength(elements2)
```
craiglowe, line 146: Is this line the only difference between upper and lower bound?  Can we have a single function that takes in a single element length and number of elements?

cmd/bedEnrichments/bedEnrichments.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if verbose > 0 {
+		log.Println("Reading bed files.")
+	}
+
+	elementsOne := bed.Read(inFile)
+	elementsTwo := bed.Read(secondFile)
+	noGapRegions := bed.Read(noGapFile)
+
+	if trimToRefGenome {
```
craiglowe, line 28: I think you should consider this being a separate function(s).  You could have a function call that makes the tree structure for the noGap, and then have two function calls that trim the two bed files to the noGap tree

craiglowe (commented):
Great work!  Had a couple small comments

bed/compare.go
dasnellings: My main problem is the use of `strings.Compare` instead of the builtin operators. I have pasted the source code for Compare below (comments included). We should stick to builtin operators to achieve the desired logic. 
```
package strings

// Compare returns an integer comparing two strings lexicographically.
// The result will be 0 if a==b, -1 if a < b, and +1 if a > b.
//
// Compare is included only for symmetry with package bytes.
// It is usually clearer and always faster to use the built-in
// string comparison operators ==, <, >, and so on.
func Compare(a, b string) int {
	// NOTE(rsc): This function does NOT call the runtime cmpstring function,
	// because we do not want to provide any performance justification for
	// using strings.Compare. Basically no one should use strings.Compare.
	// As the comment above says, it is here only for symmetry with package bytes.
	// If performance is important, the compiler should be changed to recognize
	// the pattern so that all code doing three-way comparisons, not just code
	// using strings.Compare, can benefit.
	if a == b {
		return 0
	}
	if a < b {
		return -1
	}
	return +1
}
```

Not sure I caught all the times this function is called, but it looks like it is only being used to detect cases where a < b. Only 2 of the 3 comparisons done by the Compare functions are ever used here, so taking the time to evaluate all possible relationships is overkill. You would need to double check the logic is the same, but I think this may be able to be replaced with a simpler function like 
```
func ChromEndLess(a, b bed.Bed) bool {
	if a.Chrom != b.Chrom {
		return a.Chrom < b.Chrom
	}
	return a.ChromEnd < b.ChromEnd
}
```

craiglowe: After looking at this and thinking about it, I agree with you.  Would be good to use these basic functions in the future.

bed/enrichment.go
craiglowe: agree

craiglowe (approved):
Excellent work!  Good to have all of this in here.  I am approving because this is wonderful and I don't see any bugs, but there are some style comments I feel are important (e.g. breaking large blocks of code into separate functions).  Please look at the comments to see if you can address them in a small follow-up pull request.  Great work!

PR #506, opened 2021-06-27
Norm bed max wig

cshoben:
This PR includes the addition of a new flag in bedMaxWig called "norm". The norm flag allows a user to normalize their bedMaxWig output by the total number of reads sampled. I calculated the total number of reads sampled but totaling the number of wig values, which is scaled to the wig Step value. 

Future: bedMaxWig still has some limitations. 
1) Only built to handle fixedStep. This is also dependent on our wig package, which can also only handle fixedStep at this moment. 
2) Can only handle a start =1 , again I think the wig package can only handle start 1 as well. 
3) Can only handle step = 1; I wrote the norm flag to handle a step value outside of one, but I think the wig package still has this limitation (would be great if someone can look at the wig.go file really quick and confirm what I think I'm reading).

Sample wig for reference if you are not familiar with the layout:
fixedStep chrom=chr3 start=1 step=1
9
11
22
33
365
389

And you can read more about this data type here: http://genome.ucsc.edu/goldenPath/help/wiggle.html


cmd/bedMaxWig/bedMaxWig_test.go
```diff
@@ -32,3 +32,27 @@ func TestBedMaxWig(t *testing.T) {
 		}
 	}
 }
+
+var bedMaxWigNormFlagTest = []struct {
```
RavenYL, line 36: This is just a stylistic comment but I think bedMaxWigTests and bedMaxWigNormFlagTest can be merged, and they can both use TestBedMaxWig rather than having to use TestBedMaxWigNormFlag, if the bedMaxWigTests struct has 1 more field that is "norm bool", so you can distinguish between the true and false cases, and in the TestBedMaxWig for loop norm status can be specified with v.norm

RavenYL (approved):
You mentioned possibly writing more flexibility into the start position. I want to do that for cigarToBed too, and I feel like that just involves reading the header of a wig or fasta. Gonomics might already be able to do this, but if not maybe it's not so hard to write in

cmd/bedMaxWig/bedMaxWig_test.go
cshoben: That's how I originally wrote this (with them merged) but had caught a snag. I cant remember what it was now though. I know after a conversation with Craig he said to do them separate, and I think that is generally how the test files are written? Where we would have a separate test for each flag? Maybe that makes it easier to troubleshoot problems/test fails?

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -12,14 +12,27 @@ import (
...
 	var wigData []wig.Wig = wig.Read(database)
 	var sizes []chromInfo.ChromInfo = chromInfo.ReadToSlice(chromsizeFile)
 	var outlist []bed.Bed
 	var currentBed bed.Bed = records[0]
 	var chromSlice []float64
 	var i int
+	var wigTotal float64 = 0
+
+	if norm == true { //TODO: This assumes fixedStep. Either should check for fixedStep or be able to handle both cases (fixedStep and variableStep)
+		for i := 0; i < len(wigData); i++ { //Goal here is to cycle through all the "chromosomes" of the wig: wig[0], wig[1], etc.
```
dasnellings, line 26: Just a style comment. This could be `for i := range wigData { do stuff with wigData[i] }`. Ranges are generally easier to read if you are iterating over all elements in a slice. The exception is when `i` does not start at 0 in which case you need a traditional for loop.

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -12,14 +12,27 @@ import (
...
+			var wigCounterByChrom float64 = 0
+			for k := 0; k < len(wigData[i].Values); k++ { //Cycle through each value in the float64[]
```
dasnellings, line 28: could be range too

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -12,14 +12,27 @@ import (
...
+				var chromValueMultiplyByStep float64 = 0
+				chromValueMultiplyByStep = (float64(wigData[i].Step) * wigData[i].Values[k]) // multiply each value by the step for that chrom
+				wigCounterByChrom = wigCounterByChrom + chromValueMultiplyByStep
```
dasnellings, line 31: Not a needed change, but just FYI you can also do this as 
`wigCounterByChrom += chromValueMultiplyByStep`

dasnellings (approved):

cmd/bedMaxWig/bedMaxWig.go
cshoben: If consensus for the lab is to use range, I'm happy to change it and refresh this PR.

cmd/bedMaxWig/bedMaxWig.go
cshoben: Ahhh, totally forgot about this move! Will remember for next time!

cmd/bedMaxWig/bedMaxWig_test.go
RavenYL: Interesting! That's good to know!

cmd/bedMaxWig/bedMaxWig.go
craiglowe: I think Dan makes a good point that it is a little quicker to read and possibly less prone to bugs to use the "range" syntax when doing a "normal" for loop over all elements of a slice.  Not a big deal, but something to keep in mind.

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -12,14 +12,27 @@ import (
...
 	var records []bed.Bed = bed.Read(infile)
 	var wigData []wig.Wig = wig.Read(database)
 	var sizes []chromInfo.ChromInfo = chromInfo.ReadToSlice(chromsizeFile)
 	var outlist []bed.Bed
 	var currentBed bed.Bed = records[0]
 	var chromSlice []float64
 	var i int
+	var wigTotal float64 = 0
+
+	if norm == true { //TODO: This assumes fixedStep. Either should check for fixedStep or be able to handle both cases (fixedStep and variableStep)
```
craiglowe, line 25: The code inside this "if" could have been a function that takes a slice of wiggles and gives back the sum of the values.  This might make the code a little bit easier to read, and also let people more easily re-use parts of your work

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -29,8 +42,16 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
 				if currentBed.FieldsInitialized < 7 {
 					currentBed.FieldsInitialized = 7
 				}
-				currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%f", bedRangeMax(chromSlice, records[k].ChromStart, records[k].ChromEnd)))
-				outlist = append(outlist, currentBed)
+				if norm == true {
+					maxWig := bedRangeMax(chromSlice, records[k].ChromStart, records[k].ChromEnd)
+					normMaxWig := maxWig / wigTotal
```
craiglowe, line 47: I think this is the only statement that needs to be different when norm == true.  I think you could have:
1) calc the max value with bedRangeMax
2) have an "if norm == true" that overwrites the max value with the max value divided by wigTotal
3) add the max value to the bed (might have been normalized in the previous step)
4) add this bed to the outlist.

cmd/bedMaxWig/bedMaxWig_test.go
craiglowe: I think either way could work.  I can't remember exactly, but at the time I think it was helpful to separate them out to confirm that the non-normalized version was still working, but the normalized version was failing.

craiglowe (approved):
Great work!  Exciting to have this improvement!

cmd/bedMaxWig/bedMaxWig.go
cshoben: Ooo I think I understand. I would have maxWig be calculated, then an if statement for if we do math on it, and then lines 48 and 49, which would allow me to remove lines 50-52. I will keep this in mind and will change in a future bedMaxWig.go PR.

PR #508, opened 2021-06-29
raven cigarToBed3

RavenYL:
- cmd/cigarToBed uses globalAlignment (affineGap instead of constGap) to align 2 .fasta files, each with only 1 sequence, then convert cigars to ins and del beds
- I am hoping to get this merged so others in lab can use the cigarToBed function while I continue working on some changes in the align package that would allow the alignment of large sequences
- fresh gonomics, fresh branch/PR, cigarToBed only


cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/vertgenlab/gonomics/bed" //TODO: raven added this line for cigarToBed
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/genomeGraph"
+	"log"
+	"os"      //raven added this line for MSA Fasta output
+	"strings" //raven added this line for CountSeqIdx
+)
+
+//raven did not put this helper function into the globalAlignment function because it is used twice within the globalAlignment function
```
cshoben, line 16: By globalAlignment function, are you referring to the function on line 60 or a function in a different file?

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+
+//raven did not put this helper function into the globalAlignment function because it is used in the globalAlignment function
+//faONe is target, faTwo is query
+func cigarToGraph(target fasta.Fasta, query fasta.Fasta, aln []align.Cigar) *genomeGraph.GenomeGraph {
+	answer := genomeGraph.EmptyGraph()
+	//use targetEnd and queryEnd to track position number for each fasta sequence as we move along.
+	var targetEnd, queryEnd int = 0, 0
+	var curr *genomeGraph.Node
+
+	//Creating the first node. This is done independent of all other number because this node has no 'previous' nodes.  All following code leverages the cigar output (second number printed in the Op position of the struct {}) to determine if the alignment returned a 0:match, 1:insertion, or 2:deletion. All indels are relative to the target sequence.
```
cshoben, line 40: Now that I've seen Dan do it, you could break up this big comment line into multiple lines that fit better on the screen (instead of having to scroll right). But 100% a stylistic thing.

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+//raven moved helper functions from main and non-usage functions into this function
+func GlobalAlignment_CigarToBed(inputFileOne *fileio.EasyReader, inputFileTwo *fileio.EasyReader, outFa string, outIns_bed string, outDel_bed string, FirstPos_InsBed int, FirstPos_DelBed int, Chrom string) {
+	//make sure files meet the usage requirements of globalAlignment.go
+	faOne, faDoneOne := fasta.NextFasta(inputFileOne)
+	faTwo, faDoneTwo := fasta.NextFasta(inputFileTwo)
+	fasta.ToUpper(faOne) //raven added these 2 lines to prevent lowercase letters from indexing out of range in align/linearGap.go
+	fasta.ToUpper(faTwo)
+	if faDoneOne || faDoneTwo {
```
cshoben, line 66: Just to understanding conceptually.... faDoneOne and faDoneTwo are error handlers, and this OR statement is essentially to check if these variables hold a value. If they do not, there is no error. If fasta.NextFast() didnt work properly we expect  faDoneOne and/or faDoneTwo to have some value, indicating that the input file we failed to read?

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+		log.Fatalf("error, unable to read .fa files, check for > symbol before each name and that each fasta entry is only one line. Check to make sure there are only four bases or N, globalAlignment is unable to use anything outside A-T-G-C-N.\n")
+	}
+	numSeqOne := CountSeqIdx(inputFileOne)
+	numSeqTwo := CountSeqIdx(inputFileTwo)
+	if numSeqOne > 1 || numSeqTwo > 1 {
+		log.Fatalf("multiple sequnces detected in .fa files: %v sequences in the first .fa file and %v sequences in the second .fa file. This program is designed for .fa files with only 1 sequence in them\n", numSeqOne, numSeqTwo)
+	}
+	//attempt at Dan's len(faOne) suggestion
+	//if len(string(faOne)) != 1 || len(string(faTwo)) != 1 {
```
cshoben, line 75: So I guess it's fair, that the code at line 71 wouldn't catch if there is no fasta/it is blank. But then again, line 62/63 would fail if there wasn't at least one entry anyways?

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+	//log.Fatalf("multiple sequnces detected in .fa files. This program is designed for .fa files with only 1 sequence in them.\n")
+	//}
+
+	//needleman wunsch (global alignment)
+	//raven's note: scoring matrices are in align/align.go, e.g. HumanChimp matrix, same as what UCSC Genome Browser uses
+	//raven's note: ConstGap is a function in align/linearGap.go, with only constant gapPen (gap penalty)
+	//bestScore, aln := align.ConstGap(faOne.Seq, faTwo.Seq, align.HumanChimpTwoScoreMatrix, -430)
```
cshoben, line 82: I'm guessing line 82/83 are more notes you meant to hold onto?

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+
+//raven edited this block to specify only 1 sequnce is expected in each fasta file and add Usage nad options
+func usage() {
+	fmt.Print(
+		"./cigarToBed\n" +
+			" Uses globalAlignment, affineGap (instead of constGap) to align 2 .fasta files, each with only 1 sequence, then convert cigars to ins and del beds\n" +
+			"Usage:\n" +
+			" cigartobed target.fasta query.fasta\n" +
+			"options:\n" +
+			"	-faOut=fasta MSA output filename\n" +
```
cshoben, line 159: I actually don't think you'll need to write out all these flags (lines 159-164). I believe that the flag.String and flag.Int that you use on lines 173-177 will actually trigger the info on those lines to print out with the usage anyways. I was just looking at this and testing how the usage of other programs print out and that's how I figured this out. I too wrote out lines like you have here and then deleted them when I saw how the flag functions end up all coming together. But someone more versed in this should weigh in and make sure I'm correct.

cshoben (approved):

cmd/cigarToBed/cigarToBed_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+func TestCigarToBed(t *testing.T) {
+	for _, v := range CigarToBedTests {
+		inputFileOne := fileio.EasyOpen(v.inputFileOne_Name)
+		inputFileTwo := fileio.EasyOpen(v.inputFileTwo_Name)
+		GlobalAlignment_CigarToBed(inputFileOne, inputFileTwo, v.outFa, "ins_tmp.bed", "del_tmp.bed", v.FirstPos_InsBed, v.FirstPos_DelBed, v.Chrom)
+
+		records_ins := bed.Read("ins_tmp.bed")
+		expected_ins := bed.Read(v.outIns_bed_expected)
+		if !bed.AllAreEqual(records_ins, expected_ins) {
```
cshoben, line 33: I've recently learned that bed.AllAreEqual only checks the first three columns of the bed for matching. It doesnt compare past there. Just a heads up, as it wont be comparing the name column which I believe you are using in your system here (ie column 4).

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+	var curr *genomeGraph.Node
+
+	//Creating the first node. This is done independent of all other number because this node has no 'previous' nodes.  All following code leverages the cigar output (second number printed in the Op position of the struct {}) to determine if the alignment returned a 0:match, 1:insertion, or 2:deletion. All indels are relative to the target sequence.
+	curr, targetEnd, queryEnd = genomeGraph.FaSeqToNode(target, query, targetEnd, queryEnd, aln[0], 0)
+	curr = genomeGraph.AddNode(answer, curr)
+	//Drawing the remaining nodes and all edges. Method for adding edges is based on previous nodes.
+	for i := 1; i < len(aln); i++ {
+		curr, targetEnd, queryEnd = genomeGraph.FaSeqToNode(target, query, targetEnd, queryEnd, aln[i], i)
+		curr = genomeGraph.AddNode(answer, curr)
+		if aln[i].Op == 0 {
```
dasnellings, line 47: It would be a bit easier to follow the logic here if you used the name for the Op rather than the underlying integer (e.g. `aln[i].Op == align.ColM` and `aln[i].Op == align.ColD`). It doesn't change the function at all, but makes it a bit clearer for people that don't know which integer go with which Op.

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+	//raven's edit: try AffineGap in align/affineGap.go, with diffrent gapOpen and gapExtend penalty, since gapExtend<gapOpen, encourages big gap in order to align better in unbroken chunks
+	bestScore, aln := align.AffineGap(faOne.Seq, faTwo.Seq, align.HumanChimpTwoScoreMatrix, -600, -150)
+	fmt.Printf("Using AffineGap, Alignment score is %d, cigar is %v \n", bestScore, aln)
+
+	//TODO: raven's cigarToBed main body starts here
+	//type definitions: Cigar is defined in align/align.go; aln is a []cigar, according to ConstGap defined in align/linearGap.go
+	//modeled after: mafIndels.go, wigPeaks.go
+	//insertion bed
+	insBed := fileio.EasyCreate(outIns_bed)
+	defer insBed.Close()
```
dasnellings, line 93: Best to avoid defer close when working with the fileio package so you can check the error. Because the fileio writers are buffered and potentially gzipped, there can be some serious errors.

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+	//TODO: now have to write FirstPos, but find a way to grab from fasta header
+	ChromCurrent := FirstPos_InsBed - 1 //initialize a variable to keep track of current position on chromosome
+	ChromStart := FirstPos_InsBed - 1   //initialize variables for ChromStart and ChromEnd, will need to set them equal to one another during updates, but may be more efficient to have fewer variables
+	ChromEnd := FirstPos_InsBed - 1
+	for i := 0; i < len(aln)-1; i++ { //loop through each entry in the []cigar, start from 0, end at len-2 so can still assess aln[i+1]
+		if aln[i].Op == 0 && aln[i+1].Op == 1 { //if there is an M before I, then write to bed
```
dasnellings, line 99: You might consider using the Op names instead of the integer here too. Though because your code is well commented, it is less of an issue.

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+			ChromStart = ChromCurrent + int(aln[i].RunLength) + 1                                                       //RunLengths are int64, but bed fields like ChromStart are int, so need to convert RunLengths to int, get the position at which I starts
+			ChromEnd = ChromStart + int(aln[i+1].RunLength)                                                             //get the last position that is I
+			ins := bed.Bed{Chrom: Chrom, ChromStart: ChromStart, ChromEnd: ChromEnd, Name: "ins", FieldsInitialized: 4} //TODO: now just write "chr1", but find a way to grab from fasta header?
```
dasnellings, line 102: Because you immediately write this bed, you could declare ins before the loop and reuse the struct instead of allocating a new one for each run through the loop.

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+			bed.WriteBed(insBed.File, ins)
+		}
+		if aln[i].Op != 2 { //in insertion bed, only need to update ChromCurrent if the cigar fragment is M or I, not D
+			ChromCurrent += int(aln[i].RunLength)
+		}
+	}
+	//deletion bed
+	delBed := fileio.EasyCreate(outDel_bed) //TODO: add flag for output file name
+	defer delBed.Close()
```
dasnellings, line 111: It would be good to close at the end an check the error.

cmd/cigarToBed/cigarToBed.go
dasnellings: You are right. The call to `flag.PrintDefaults()` on line 165 handles this. If you do not like the defaults for whatever reason you could make your own like you have now, but doing both is unnecessary.

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+	if len(flag.Args()) != expectedNum {
+		flag.Usage()
+		log.Fatalf("error, expecting 2 .fasta files to be able to align, only found %d files...\n", len(flag.Args()))
+	}
+
+	//read in sequences that should be put in as fasta type files.
+	//raven edited this block to save fileio.EasyOpen as file handles, so that the file is only opened 1 time for 2 purposes: faDone and CountSeqIdx
+	inputFileOne := fileio.EasyOpen(flag.Arg(0)) //raven's note: EasyOpen returns the type EasyReader
+	inputFileTwo := fileio.EasyOpen(flag.Arg(1))
+	outFa := *faOut
```
dasnellings, line 190: This works fine, though another option would be to de-reference the pointers when you call the CigarToBed function (e.g. `GlobalAlignment_CigarToBed( ... , *faOut, *insBedOut, *delBedOut, ... )`)

dasnellings (approved):
I really only had stylistic comments so I am approving. As a side note, your comments make the code so much easier to review. Keep it up!

craiglowe (approved):
Great work to have these improvements!  Nice!
As a high-level comment, I would consider an additional pull request in the future were this is no longer a separate program from globalAlignment, but rather updating the globalAlignment program to have these improvements.  For example, the user could select the type of alignment algorithm they would like (constant gap or affine gap) and the type(s) of output they would like (insBed, delBed, graph, fasta)

cmd/cigarToBed/cigarToBed_test.go
cshoben: @RavenYL I recommend actually making this change a priority or your test file might not be functioning. You can run fileio.AreEqual(ins_tmp.bed, v.outIns_bed_expected) I believe to do the testing until the bed.AllAreEqual() is fixed (really the change will be in bed.Equal() which bed.AllAreEqual() depends on).

RavenYL: I just switched to fileio.AreEqual and commented on the bed.AllAreEqual situation in my new PR. Thank you!

cmd/cigarToBed/cigarToBed.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+//raven did not put this helper function into the globalAlignment function because it is used twice within the globalAlignment function
+//raven wrote this block to count sequences based on the Read function in gonomics/fasta/fasta.go
+//raven changed the input variable from filename string to inputFile EasyReader, so that the file is only opened 1 time for 2 purposes: faDone and CountSeqIdx
+func CountSeqIdx(inputFile *fileio.EasyReader) int {
+	var line string
+	var seqIdx int = 1 //I know in Read seqIdx is int64 and starts with -1, but I am using it differently here. EasyReader comes in having read the first fasta, so seqIdx starts with 1
+	var doneReading bool = false
+	defer inputFile.Close()
```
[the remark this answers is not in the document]
RavenYL: Just implemented this change for both cigarToBed and globalAlignment. Thank you!

cmd/cigarToBed/cigarToBed.go
RavenYL: Yes I remember seeing it too! I'll do that moving forward.

go.sum
```diff
@@ -1,12 +1,15 @@
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
+github.com/biogo/boom v0.0.0-20150317015657-28119bc1ffc1 h1:LAHY5JxqhOgJDeDBGKsQ4300qd3sG8C0j5CQS8gD+Kw=
```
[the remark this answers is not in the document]
RavenYL: I think I ran go mod tidy when I first made the branch, and when I just ran it again the below is my output, but my local go.sum didn't seem to have changed according to git:
`raven-luo@Raven-Surface-Pro:/mnt/c/Users/raven/Desktop/GOPATH/src/github.com/vertgenlab/gonomics$ go mod tidy
go: downloading github.com/biogo/boom v0.0.0-20150317015657-28119bc1ffc1
go: downloading gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15
go: downloading github.com/kr/pretty v0.2.0
go: downloading github.com/kr/text v0.1.`
Maybe when I push the commit I will see a different go.sum?
Or maybe it was go mod tidy that added these lines? I read on stack overflow that those duplicate lines may be checksums that are needed to ensure consistency when I am loading or changing dependencies.

cmd/cigarToBed/cigarToBed.go
RavenYL: Looking back to the definition of the fasta.NextFasta function, I agree! Thank you!

cmd/cigarToBed/cigarToBed.go
RavenYL: Just fixed this. Thank you!

cmd/cigarToBed/cigarToBed.go
RavenYL: Just fixed. Thanks!

cmd/cigarToBed/cigarToBed.go
RavenYL: Just changed this. Now there are much fewer lines. Thank you!

cmd/cigarToBed/cigarToBed.go
RavenYL: Good to know! Just fixed this.

cmd/cigarToBed/cigarToBed.go
RavenYL: Just fixed. Thank you! I can see how it'll speed things up

cmd/cigarToBed/cigarToBed.go
RavenYL: I agree. That does make the code easier to comprehend. Just fixed!

PR #509, opened 2021-06-29
Seamless bam reading

dasnellings:
Made some changes to the Sam GoReadToChan function as we discussed in code review. It will now detect if the input file has the '.bam' extension and parse using the bam reader if it does. I also added a new read to chan style function called GoReadToChanRecycle. This implements the whiteboard strategy as we discussed before and works with both Bam and Sam (although the bam recycle is far more efficient than the sam recycle which can be revisited later if needed). This function saves ~50% of allocations per read. It is still not 100% as there are a few allocations that are currently unavoidable when reading a bam file. 

I also fixed a couple bugs in the bam reader that came up as I was testing.

### DIFF
--- a/sam/bamRead.go
+++ b/sam/bamRead.go
@@ -9,7 +9,9 @@ import (
 	"github.com/vertgenlab/gonomics/dna"
 	"io"
 	"log"
+	"reflect"
 	"strings"
+	"unsafe"
 )
 
 // bam is a binary version of sam compressed as a bgzf file
@@ -221,14 +223,12 @@ func DecodeBam(r *BamReader, s *Sam) (binId uint32, err error) {
 
 	// ******
 	// Unsafe String Conversion
-	/*
-		s.Qual = unsafeByteToString(s.Qual, qual) // unsafe version
-	*/
+	s.Qual = unsafeByteToString(s.Qual, qual) // unsafe version
 	// ******
 
 	// ******
 	// Safe String Conversion
-	s.Qual = string(qual) // TODO this is 1 alloc per read, should change to []byte and remove unsafe ref above
+	//s.Qual = string(qual) // TODO this is 1 alloc per read, should change to []byte and remove unsafe ref above
 	// ******
 
 	// The sam.Extra field is not parsed here as it would require parsing tags to their value, then
@@ -240,7 +240,6 @@ func DecodeBam(r *BamReader, s *Sam) (binId uint32, err error) {
 		lenReadName + (4 * numCigarOps) + (((lenSeq) + 1) / 2) + lenSeq)) // to get remaining bytes in alignment
 	if cap(s.unparsedExtra) < len(ex) {
 		s.unparsedExtra = make([]byte, len(ex))
-		copy(s.unparsedExtra[:len(ex)], ex)
 	}
 	s.unparsedExtra = s.unparsedExtra[:len(ex)]
 	copy(s.unparsedExtra, ex)
@@ -289,7 +288,7 @@ func trimNulOrPanic(s string) string {
 	return strings.TrimRight(s, "\u0000")
 }
 
-/* // Uncomment to use unsafe string conversion
+// Uncomment to use unsafe string conversion
 // unsafeByteToString mutates the input string s to the string formed by
 // calling `string(toCopy)`. This makes s unsafe for use between calls to
 // DecodeBam unless s is copied to a new variable.
@@ -310,4 +309,3 @@ func unsafeByteToString(s string, toCopy []byte) string {
 	copy(b, toCopy)                              // mutate s
 	return s
 }
-*/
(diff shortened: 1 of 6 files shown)


craiglowe:
This looks like it also has the unsafe and reflect code that I was thinking would only be in the other pull request

dasnellings:
The unsafe function was still there, but the code that used it is commented out in favor of the safe version. I have also commented out the unsafe function to avoid importing unsafe and reflect if we are not using them.

sam/bamRead.go
```diff
@@ -207,17 +219,31 @@ func DecodeBam(r *BamReader, s *Sam) (binId uint32, err error) {
...
 	// saved in the unparsedExtra field of sam, such that a user may call a function to parse these
 	// bytes if and only if they need access to the tag fields. This should also make bam reading a
 	// fair bit faster.
-	s.unparsedExtra = r.next(blkSize - (staticBamAlnSize +
+	ex := r.next(blkSize - (staticBamAlnSize +
 		lenReadName + (4 * numCigarOps) + (((lenSeq) + 1) / 2) + lenSeq)) // to get remaining bytes in alignment
-
+	if cap(s.unparsedExtra) < len(ex) {
+		s.unparsedExtra = make([]byte, len(ex))
+		copy(s.unparsedExtra[:len(ex)], ex)
```
craiglowe, line 243: I think you can get rid of this line because you have what I think is a copy to do the same thing in a couple lines.

sam/bamRead.go
```diff
@@ -207,17 +219,31 @@ func DecodeBam(r *BamReader, s *Sam) (binId uint32, err error) {
...
+	}
+	s.unparsedExtra = s.unparsedExtra[:len(ex)]
```
craiglowe, line 245: I think you want to wrap this line in an "else" because you don't need to do it if you went into the above "if"

sam/io.go
```diff
@@ -40,17 +40,158 @@ func ReadToChan(filename string, data chan<- Sam, header chan<- Header) {
...
+
+	if strings.HasSuffix(filename, ".bam") {
+		go readBamToChanRecycle(filename, parsedRecordsInit, recycledStructsInit, headerChan)
+	} else {
+		go readSamToChanRecycle(filename, parsedRecordsInit, recycledStructsInit, headerChan)
+	}
+
+	// spawn a temporary goroutine that will allocate bufferSize number of
+	// Sam structs and send them to the readToChan function for reading.
+	go func(chan<- *Sam) {
```
craiglowe, line 179: Is it easier/simpler to have this not be a go routine and just do a for loop that fills up the buffer?

sam/io.go
```diff
@@ -165,6 +324,78 @@ func processAlignmentLine(line string) Sam {
 	return curr
 }
 
+// processAlignmentLineRecycle parses a string representation of a sam file into a single Sam struct.
+func processAlignmentLineRecycle(line string, curr *Sam) {
```
craiglowe, line 328: I am wondering if we can reduce code duplication between this function and the original processAlignmentLine.  Can process alignment line become a wrapper around this function that:
1) creates a new sam struct
2) passes it to this function
3) returns the sam

craiglowe (approved):
Awesome!

sam/io.go
```diff
@@ -40,17 +40,158 @@ func ReadToChan(filename string, data chan<- Sam, header chan<- Header) {
...
+// GoReadToChan will detect if the input file ends in ".bam" and
+// automatically switch to a bam parser.
 func GoReadToChan(filename string) (<-chan Sam, Header) {
 	data := make(chan Sam, 1000)
 	header := make(chan Header)
-	go ReadToChan(filename, data, header)
+
+	if strings.HasSuffix(filename, ".bam") {
+		go readBamToChan(filename, data, header)
+	} else {
```
cshoben, line 137: Does it make sense to have another if for .sam and then else is a fatal or something. That way we catch is someone puts into something that is neither bam or sam? There could be something upstream that already does this, and I'm just missing it.

sam/io_test.go
```diff
@@ -204,3 +204,90 @@ func TestReadAndWriteReal(t *testing.T) {
 		}
 	}
 }
+
+func TestGoReadToChanRecycleSam(t *testing.T) {
+	receiveChan, recycleChan, _ := GoReadToChanRecycle("testdata/small.sam", 1)
+	expected := []Sam{r001, r002, r003, r004, r003Supplemental, r001Supplemental}
+	var i int
+	for val := range receiveChan {
+		if !Equal(*val, expected[i]) {
```
cshoben, line 213: Jw, what is this Equal() coming from? Is it in the sam package and that's why we dont need a prefix like fileio.AreEqual()?

cshoben (approved):
Just a few questions embedded. Maybe we could talk a bit again at code review about how channel work by walking through an example in your code that does the recycling.

craiglowe (approved):
Great work!!!!

rimangan (approved):

RavenYL (approved):

PR #510, opened 2021-06-29
Unsafe bam reading

dasnellings:
Doing naughty things with strings to drastically reduce allocations when reading bam files. 

This PR is not for merging but for discussion. 

Strings in go are immutable (i.e. `string[i] = "A"` is illegal) even though under the hood they are just a `[]byte`. This has some nice properties and makes strings a very 'safe' type. This does however introduce some inefficiencies. Specifically, you cannot directly convert a `[]byte` to a `string`, go automatically copies the `[]byte` under the hood whenever you do `string([]byte)`. We can avoid this process by making strings mutable using the unsafe and reflect packages. I can not imagine a reasonable use case where this would be a problem for bam, and given the potential benefits I think we should consider switching to the unsafe version until it becomes problematic. 

This is maybe something we should discuss at code review as I would not put it past anyone if they are staunchly against introducing the reflect or unsafe packages into gonomics (i am on the fence myself).

### DIFF
--- a/sam/bamRead.go
+++ b/sam/bamRead.go
@@ -9,6 +9,7 @@ import (
 	"github.com/vertgenlab/gonomics/dna"
 	"io"
 	"log"
+	"reflect"
 	"strings"
 	"unsafe"
 )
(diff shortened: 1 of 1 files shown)


craiglowe:
What kind of time/speed improvements do we see with this?
I took a quick look at the code, and it seems like it does not avoid the copy, but avoids the re-allocation of memory of you have memory lying around that can be re-used.  Is that the general idea?  From the code it looks like you can doname/re-use an unneeded string you have lying around, and it will copy the []byte into that memory instead of making its own new memory.

craiglowe (approved):
I had to stare at this for a bit, but this looks like a good idea!  I don't see any bugs and we have the safer version commented out if anything starts to go wrong.

dasnellings:
I think there is still room for improvement here. As you say, it does not avoid the copy, just the allocation. My thought is that generally copying is fast, but allocations to the heap (i.e. strings and slices in Sam) will increase pressure on the garbage collector which can lead to serious slowdowns for long-running programs. When I test this there is no real change is speed, but a reduction in allocations and GC calls. I think a stress test is in order to see if this has time benefits on a larger scale. 

You could avoid the copy altogether by casting the byte slice directly to a string such that the underlying memory is part of the read buffer. In this case the Sam is only valid until the next call to DecodeBam (with any *Sam input) which is perhaps tolerable, but for now I went with the slightly safer version where the Sam is valid until the next time this specific *Sam is input to DecodeBam. I am open to changing to a direct cast to string though.

PR #512, opened 2021-07-03
bmw fields init and test equality corrected

cshoben:
**1) Correct fields initialized so input bed is properly handled and bedMaxWig bed is properly written.** 

Discovered that if input bed was only 3 fields, the program did not create the bed.Annotation field, so all the bed max wig score were not being written into the new file. This update tests for the number of fields and if less than needed if updates it to the necessary 7. 
Bed struct reminder:

` type Bed struct {`
`	Chrom             string`
	`ChromStart        int`
	`ChromEnd          int`
	`Name              string`
	`Score             int`
	`Strand            Strand`
	`FieldsInitialized int      //number of fields that are initialized, used for smart writing.`
	`Annotation        []string //long form for extra fields`
`}`

**2) Correct equality testing in test file, original wias only testing first three columns.**

The test file was using bed.Equal() which looks like:
//Equal returns true if two input Bed entries have the same Chrom, ChromStart, and ChromEnd. False otherwise.
`func Equal(a Bed, b Bed) bool {`
	`if strings.Compare(a.Chrom, b.Chrom) != 0 {`
	`	return false`
	`}`
	`if a.ChromStart != b.ChromStart {`
	`	return false`
	`}`
	`if a.ChromEnd != b.ChromEnd {`
	`	return false`
	`}`
	`return true`
`}`

so it was not testing for anything past the third field, including the bmw values that were being output. Updated to use fileio.AreEqual() which compares for file equality. 

Future PR, update the bed.Equal() function to test all fields/columns of the beds to be compared.


cshoben:
I see it is failing the test, but even when my branch is updated with main, I am passing on my end. Maybe someone can help me troubleshoot this?

cshoben:
> I see it is failing the test, but even when my branch is updated with main, I am passing on my end. Maybe someone can help me troubleshoot this?

Nvm, I see it is passing now that I reran the checks. Does it do that often? Failing and then passing?

craiglowe (approved):
Nice catch and fix!

dasnellings (approved):

RavenYL (approved):

rimangan (approved):

PR #513, opened 2021-07-10
Gaussian Elimination

rimangan:
A new function, Rref, transposes an input matrix [][]float64 into row reduced echelon form using Gaussian elimination. Adapted from pseudocode on the wikipedia page. Passes tests.
I need some matrix operations for some new ideas I'm chasing, more info to come


craiglowe (approved):
This looks great!  Will be always be able to put any matrix in RREF?  I am thinking yes, but want to double check there are not corner cases.

RavenYL (approved):

dasnellings (approved):
LGTM

PR #514, opened 2021-07-10
Raven alignLowMem2

RavenYL:
- In the last PR I made a lowMem version of constGap. In this PR I made a lowMem version of affineGap, which is another alignment algorithm
- I removed the old constGap and affineGap and replaced them with the new lowMem versions. They should function exactly the same as the old versions, but just with lower memory because of the recycling of rows in the scoring matrix. I switched i/j row/column alpha/beta around in constGap but it didn't change downstream applications
- go test passes in both the align package and the gonomics repository as a whole


craiglowe (approved):
Great work!  A couple small comments on things I noticed.

dasnellings (approved):
LGTM

rimangan (approved):

align/affineGap.go
```diff
@@ -7,27 +7,29 @@ import (
 	"log"
 )
 
-// the data structure is a 3d slice where the first index is 0,1,2 and represents
-// the match, gap in x (first seq), and gap in y (second seq).
-func initAffineScoringAndTrace(firstSeqLen int, secondSeqLen int) ([][][]int64, [][][]ColType) {
-	m := make([][][]int64, 3)
+// the trace data structure is a 3d slice where the first index is 0,1,2 and represents the match, gap in x (first seq), and gap in y (second seq).
+// m used to have the same data structure as trace, but has been simplified into a 2d slice, where the second index for mColumn is removed in order to recycle memory by rows
+func initAffineScoringAndTrace(firstSeqLen int, secondSeqLen int) ([][]int64, [][]int64, int, [][][]ColType) {
```
cshoben, line 12: I'm not used to seeing/interpreting stacked slices like this, so just to be sure, is [ ][ ][ ] where we have a slice of slices of slices??? Like [ [ [ABC][ABC] ]  [ [DEF][DEF] ]   ]  ?

RavenYL: Yes! I think of it as a 3D matrix, or a gridded diagram with 3 axes (x,y,z) and there is a number in each box. But in this case I just think of a 2D matrix*3 since the outermost [] has only 3 possible values (0,1,2)

craiglowe (approved):
Wonderful!

PR #515, opened 2021-07-13
makeExponentialOutputNormBMW

cshoben:
- Working with my new normalization flag, I realize that the original output of rounding to 6 decimal places doesnt work well with real data. It causes majority of my signal to zero out. In order to capture small values, I've updated the output to be an exponential value, which will round to the 6th decimal place. 

- I've also integrated advice received on my last PR for bedMaxWig updates. (range, +=, and streamline of appending to the output bed)
 
- Relevant test file was updated.


cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -42,16 +42,14 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
...
-					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%f", float64(normMaxWig))) // %f will round to 6th decimal place
-					outlist = append(outlist, currentBed)
-				} else {
-					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%f", bedRangeMax(chromSlice, records[k].ChromStart, records[k].ChromEnd)))
-					outlist = append(outlist, currentBed)
-
+					maxWig = maxWig / wigTotal
+					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%e", float64(maxWig))) // %e will have 6 decimals in scientific notation (eg; 1.234456e+78)
+				}else {
+					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%f", float64(maxWig))) // %f will round to 6th decimal place
```
cshoben, line 50: I tried to combine lines 47 through 50 but having the output for non-normalized bedMaxWig seemed difficult to read. Didnt test it, but I think that the exponential way takes up more bytes too?

Like seeing:
chr3    1       2       trf     52      .       1.100000e+01
instead of:
chr3    1       2       trf     52      .       11.000000

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -42,16 +42,14 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
...
-					normMaxWig := maxWig / wigTotal
-					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%f", float64(normMaxWig))) // %f will round to 6th decimal place
-					outlist = append(outlist, currentBed)
+					maxWig = maxWig / wigTotal
+					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%e", float64(maxWig))) // %e will have 6 decimals in scientific notation (eg; 1.234456e+78)
 				} else {
-					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%f", bedRangeMax(chromSlice, records[k].ChromStart, records[k].ChromEnd)))
-					outlist = append(outlist, currentBed)
-
+					currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%f", float64(maxWig))) // %f will round to 6th decimal place
```
rimangan, line 50: Should this also be %e? Does that make sense in the non-normalized function as well?

cshoben: Hmm not sure if you were able to see my comment above. But I didnt personally like the output when I used %e here too, so I left it as %f. Since the number of reads is always a whole number, we should only have whole numbers in the non-normalized so using exponents felt unnecessary. But maybe there is a good reason I am not thinking of?

rimangan: Yes, I was on my phone and missed the comment above. Number of reads is always a whole number, and represents our major use case, although in many cases we can have non-integer values in wig files. A program I'm currently writing, for example, might benefit from %e formattin without normalization. Perhaps we can have an option for number format in the output?

cshoben: If there is use for the %e then it sounds like that is the way to go? It's not like we "can't" read whole numbers as exponentials, I was just being lazy. 
If we do %e as all output it also simplifies the code, but do you think this adds a lot of extra bytes to the file when we don't always need them?
Does any one else have a preference between all output as %e versus writing a new option for this?
Also, I guess we could see if the wig input is whole numbers and use an "if" statement for choosing between %f and %e, although if the wig is a mix of whole numbers and fractions I feel like it then is more complicated than it's worth.

rimangan: In this case we are adding extra bytes to bed files, which are generally small as is. Thinking about it, I think %e in both cases might be the way to go. Curious what other people have to say in code review

craiglowe: People might like %g
It has some logic to print different numbers in the most human-readable format and avoid trailing zeros.

cmd/bedMaxWig/bedMaxWig.go
craiglowe: You could consider merging lines 48 and 50 into a single line that uses %g

craiglowe (commented):
Things look really good.  I am wondering if %g may be a compromise among all the comments.

cmd/bedMaxWig/bedMaxWig.go
cshoben: I like this change. Here is how it outputs compared to how it did with the code currently in this pull request. The top "diff" is without the norm flag, the bottom "diff" is with the norm flag. I ran the bedMaxWig_test.go to see how the output changes. 
image
I'm going to plan to change to %g and update the pull request unless @rimangan you think this won't work for your needs?

rimangan: Love it! Wasn't aware of this format option

cshoben:
%g is now in use and updates to test files have been done.

craiglowe (approved):
Great work!

RavenYL (approved):

rimangan (approved):

PR #516, opened 2021-07-13
Bam writer

dasnellings:


sam/bamWrite.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// WriteToBamFileHandle writes a single Sam struct to a bam file.
+func WriteToBamFileHandle(bw *BamWriter, s Sam, bin uint16) {
+	bw.recordBuf.Reset()
+
+	// refID
+	idx, ok := bw.refMap[s.RName]
+	if !ok {
+		idx = -1
```
craiglowe, line 87: Does this mean that the read is mapping to a chromosome/scaffold that is not in the header?  I would have thought that is not allowed / is an error.  I realize some reads may be unmapped, but I am guessing that is something we can explicitly look for.

sam/bamWrite.go
```diff
@@ -0,0 +1,335 @@
...
+	// len seq
+	le.PutUint32(bw.u32[:4], uint32(len(s.Seq)))
+	bw.recordBuf.Write(bw.u32[:4])
+
+	// next ref id
+	if s.RNext != "=" {
+		idx, ok = bw.refMap[s.RNext]
+	}
+	if !ok {
+		idx = -1
```
craiglowe, line 123: Similar comment to above.

sam/bamWrite.go
```diff
@@ -0,0 +1,335 @@
...
+	case 'H':
+		return 5
+	case 'P':
+		return 6
+	case '=':
+		return 7
+	case 'X':
+		return 8
+	default:
+		return 0 // unreachable
```
craiglowe, line 227: Should this give an error if reached?

craiglowe (approved):
Amazing!  Looks great.  My only comments where places where I thought we could consider adding errors

sam/bamWrite.go
dasnellings: I was using this to catch unmapped reads, but you are right that an explicit check would be more appropriate. Changed.

sam/bamWrite.go
dasnellings: Yes, nice catch. I must have forgotten. I put the comment there, but not the error.

sam/bamWrite.go
craiglowe: Do we need this check to be above the "if !ok" a few lines before?  I am guessing that "*" will not be present in bw.refMap and that unmapped reads will error on line 87 before they get to this if statement that catches them.  I was think we might have something like: if unmapped, then set idx to -1.  else if !ok, then give an error.

craiglowe (commented):
One quick question added

sam/bamWrite.go
dasnellings: Yes, nice catch.

craiglowe (approved):
Great work!

RavenYL (approved):

sam/bamWrite.go
```diff
@@ -0,0 +1,349 @@
...
+const (
+	nul byte = '\u0000'
+)
+
+// NewBamWriter creates a new bam writer to the input io.Writer.
+// The magic bam bytes and header are immediately written to the BamWriter.
+func NewBamWriter(w io.Writer, h Header) *BamWriter {
+	var bw BamWriter
+	bw.Writer = bgzf.NewWriter(w)
+	bw.buf.WriteString(magicBam)
```
cshoben, line 34: Where does this magicBam come from? And so I understand this line, will the output from the WriteString() be placed in  the bw variables struct, in the buf position? I'm also not sure where this WriteString is coming from?

rimangan (approved):

PR #518, opened 2021-07-15
Giraf test file, close file and delete it

cshoben:
I noticed a giraf test file kept coming up as a changed file in my git status. Looks like a test file was missing the code to delete it so I added a remove in there. 

I also opted for a  outfile.Close() instead of "defer outfile.Close()" so that we didnt have to worry about it closing after deleting (something that "defer outfile.Close()" can cause which ends up recreating the file we intended to delete with any leftover info in the buffer).


craiglowe (approved):
Good catch!

dasnellings (approved):

RavenYL (approved):

PR #519, opened 2021-07-16
Automatic gzip detection

dasnellings:
This PR is a branch from my Read Stdin/Stdout PR.

I did not like how the user needed to specify stdin.gz to read in a gzip file so here I implement automatic gzip detection by checking for the 'magic' gzip bytes. Magic bytes are a sequence of bytes at the beginning of a file that serve as a type of fingerprint to tell parsers what kind of file they are. For gzip the bytes are `1f 8b`. 

To check for gzip files I have added the `fileio.IsGzip` function that will tell you if the input `io.ReadSeeker` is a gzip file. Because gzip only uses 2 bytes, just checking the magic bytes is not 100% foolproof (though it is very unlikely that any other file would start with these bytes). Because of this the `EasyOpen` function will only parse a file as gzip if it has the magic bytes AND if it has the `.gz` suffix. Error/Warning is thrown if one but not the other of these is present. 

One caveat is that the `fileio.IsGzip` function does not work with stdin because it is not seekable. Therefore there is a new (unexported) reader that is implemented in magic.go that can be used to peek at the first n bytes of the stream.

Because stdin only has the magic byte check and not the `.gz` check, it is theoretically possible that this could error on some files, but because the magic bytes are illegal for many file types (such as UTF-8 files) I don't think this is something we need to worry about.

### DIFF
--- a/fileio/easyio.go
+++ b/fileio/easyio.go
@@ -32,8 +32,13 @@ func EasyOpen(filename string) *EasyReader {
 	if strings.Contains(filename, "http") {
 		return EasyHttp(filename)
 	}
+
 	answer := EasyReader{}
-	answer.File = MustOpen(filename)
+	if filename == "stdin" {
+		answer.File = os.Stdin
+	} else {
+		answer.File = MustOpen(filename)
+	}
 	var err error
 
 	if strings.HasSuffix(filename, ".gz") {
@@ -50,7 +55,16 @@ func EasyOpen(filename string) *EasyReader {
 // EasyCreate creates a file with the input name. Panics if errors are encountered.
 func EasyCreate(filename string) *EasyWriter {
 	answer := EasyWriter{}
-	answer.File = MustCreate(filename)
+
+	switch filename {
+	case "stdout":
+		answer.File = os.Stdout
+	case "stderr":
+		answer.File = os.Stderr
+	default:
+		answer.File = MustCreate(filename)
+	}
+
 	answer.internalBuff = bufio.NewWriter(answer.File)
 
 	if strings.HasSuffix(filename, ".gz") {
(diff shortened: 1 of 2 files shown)


RavenYL (approved):
Glad that the stdin issue was addressed!

fileio/easyio.go
```diff
@@ -32,25 +33,63 @@ func EasyOpen(filename string) *EasyReader {
...
+	var readerInput io.Reader
+
+	if strings.HasPrefix(filename, "stdin") {
+		// when reading stdin we will assume the input is gzipped
+		// if the file begins with the two magic gzip bytes 1f8d.
+		// If it does, append .gz to the filename so it is parsed
+		// as gzip in the following switch case.
+		answer.File = os.Stdin
+		readerInput, hasMagicGzip = newStdinMagicReader(magicGzip)
+		if hasMagicGzip {
```
craiglowe, line 48: I think this is the core of the functionality that is added.  If the user tells us to read from "stdin" instead of "stdin.gz" we are now smart enough to realize that they probably wanted "stdin.gz" and seamlessly make this silent change for them.
Your work is technically impressive, but I wonder if a more general implementation would have advantages.  I wonder if right now we are in a bit of a middle ground where we are adding a decent number of lines of code and complexity, to catch a mistake of "stdin" instead of "stdin.gz".  What if we sent inputs into a buffer where we can peek the magic bytes.  We can then either use that buffer if it is plain text, or if we detect magic bytes during the peek we can put it into a gzip buffer and back into a buffered reader we can call ReadString on.
Another option would be to believe the file extension when there is one (one buffer for plain text and two buffers if gzip) and otherwise we would still have one buffer for plain text or three buffers if the file with no extension (e.g. /dev/stdin) was given.  I think this could be added to the easyreader and would not need many more lines and might be a more general solution that could see more uses?  Are there performance differences?

craiglowe (commented):
Cool work!  I left a couple comments.

rimangan (approved):

craiglowe (approved):
Great work!

PR #520, opened 2021-07-19
Bgzf easy read write

dasnellings:
Creates a more straightforward reader/writer for BGZF that does not require you work in blocks (i.e. the blocks get handled under the hood). Still has the old reader/writer as BlockReader/BlockWriter.

### DIFF
--- a/bgzf/reader.go
+++ b/bgzf/reader.go
@@ -25,10 +25,10 @@ type BlockReader struct {
 // that automatically read a new block when the previous block has
 // been completely read.
 type Reader struct {
-	br 	*BlockReader
+	br           *BlockReader
 	blk          *Block
 	intermediate bytes.Buffer
-	eof bool
+	eof          bool
 }
 
 // Next returns the next n bytes from the bgzf block.
--- a/bgzf/reader.go
+++ b/bgzf/reader.go
@@ -4,20 +4,79 @@ import (
 	"bufio"
 	"bytes"
 	"compress/gzip"
+	"io"
 	"log"
 	"math"
 	"os"
 )
 
-// Reader is the decompressor for BGZF files which is at it's core
+// BlockReader is the decompressor for BGZF files which is at it's core
 // a gzip reader. The BGZF reader retains a *os.File to use for
 // seek functionality as well as a *bufio.Reader to implement the
 // io.ByteReader interface, required for reading gzip files as
 // independent concatenated blocks, rather than a multistream.
-type Reader struct {
+type BlockReader struct {
 	file         *os.File
 	buf          *bufio.Reader
 	decompressor *gzip.Reader
+}
+
+type Reader struct {
+	br 	*BlockReader
+	blk          *Block
+	intermediate bytes.Buffer
+	eof bool
+}
+
+// Next returns the next n bytes from the bgzf block.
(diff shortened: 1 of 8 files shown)


RavenYL (approved):

rimangan (approved):

bgzf/reader.go
```diff
@@ -4,22 +4,84 @@ import (
...
+	br           *BlockReader
+	blk          *Block
+	intermediate bytes.Buffer
+	eof          bool
+}
+
+// Next returns the next n bytes from the bgzf block.
+// Next handles reading new block if more bytes are
+// requested than are in the currently store block.
+func (r *Reader) Next(n int) []byte {
```
craiglowe, line 37: Can this be written in procedural style instead?
Next(r *Reader, n int)
Is there an advantage to having it be a method?

bgzf/reader.go
```diff
@@ -78,9 +148,17 @@ func (r Reader) Seek(offset int64, whence int) (ret int64, err error) {
 	return
 }
 
+// Seek wraps the Seek function for BlockReader and resets the reader buffer.
+func (r *Reader) Seek(offset int64, whence int) (ret int64, err error) {
```
craiglowe, line 152: Procedural style?

bgzf/writer.go
```diff
@@ -13,24 +14,36 @@ import (
...
+// NewWriter creates a bgzf writer that writes bytes instead of blocks
+func NewWriter(w io.Writer) *Writer {
+	var bw Writer
+	bw.bw = NewBlockWriter(w)
+	return &bw
 }
 
 // Write input bytes as a single bgzf block.
-func (w Writer) Write(p []byte) (n int, err error) {
+func (w *BlockWriter) Write(p []byte) (n int, err error) {
```
craiglowe, line 46: Procedural possible?  I think procedural fits the gonomics style better, but not sure if this is to implement an interface somewhere

bgzf/writer.go
```diff
@@ -65,8 +78,19 @@ func (w Writer) Write(p []byte) (n int, err error) {
 	return
 }
 
+// Write writes any number of bytes to a buffer then writes the buffer
+// as a block once 64KB of bytes have been stored.
+func (w *Writer) Write(p []byte) (n int, err error) {
```
craiglowe, line 83: Same procedural vs OOP question

bgzf/reader.go
```diff
@@ -4,22 +4,84 @@ import (
...
+type BlockReader struct {
 	file         *os.File
 	buf          *bufio.Reader
 	decompressor *gzip.Reader
 }
 
+// Read wraps a BlockReader and provides the Next and Read functions
+// that automatically read a new block when the previous block has
+// been completely read.
+type Reader struct {
```
craiglowe, line 27: I don't see "Reader" used right now.  I see bams using the blockReader, but I am guessing that you have a vision for using this somewhere else

craiglowe (approved):
I am approving.  I had two questions though:
1) I see the addition of a reader as well as block reader, but I only see block reader used
2) Some functions are programed as a method, but I was not sure if we could make it procedural, or if it is satisfying an interface that I didn't notice.

cshoben (approved):

PR #521, opened 2021-07-19
Remove biogo dependency

dasnellings:
This PR switches the binaryGiraf code from the biogo bgzf reader/writer to the gonomics implementation. Dependent on my last PR.

This removes the biogo dependency from gonomics, drastically reducing our dependency tree (see go.sum diff).


bgzf/reader.go
```diff
@@ -4,22 +4,81 @@ import (
...
 	"bytes"
 	"compress/gzip"
+	"io"
 	"log"
 	"math"
 	"os"
 )
 
-// Reader is the decompressor for BGZF files which is at it's core
+// BlockReader is the decompressor for BGZF files which is at it's core
```
rimangan, line 13: *its

bgzf/reader.go
```diff
@@ -4,22 +4,81 @@ import (
...
+// Read retrieves the next len(b) bytes from the reader
+// and automatically handles block reading.
+func (r *Reader) Read(b []byte) (n int, err error) {
+	retrieved := r.Next(len(b))
+	n = len(retrieved)
+	if r.eof {
+		err = io.EOF
+	}
+	copy(b, retrieved)
+	return
```
rimangan, line 79: confused by this syntax, should it be return n, err?

bgzf/reader.go
dasnellings: fixed in other PR.

bgzf/reader.go
dasnellings: This is a quirk of using named variables. The named variables are pre-allocated and may be set in the function. The empty return at the end returns whatever is the current value of n and err rather than explicitly stating `return n, err`.

RavenYL (approved):

craiglowe (approved):
Wonderful work!

PR #522, opened 2021-07-23
multiFaAcceleration

rimangan:
multiFaAcceleration performs branch length analysis of mutation rate velocity, acceleration, and constraint. This is all for some pretty new stuff, and related to my past pull request implementing Rref for solving systems of linear equations.


rimangan:
My initial observation is that this program runs very slowly and there may be some optimizations available. Significantly, the rref can be precalculated, but that requires a bit of math on our end

rimangan:
This pull request is a bit technical, so I've included a document explaining the theory in the vglDocumentation repository. Can be found at the link below.
https://github.com/vertgenlab/vglDocumentation/blob/master/riley/multiFaAccelerationREADME/multiFaAcceleration_README.pdf

dasnellings:
Nice README. You should consider adding it to the command folder in gonomics.

rimangan:
In response to Dan's comment, added the README pdf to the cmd folder itself. Not really precedent for this, but I think this is a good program to have a readme like this

rimangan:
https://www.youtube.com/watch?v=MhjSSxcGjaY&list=PLXwjzs_mabFrlRF7uALEomfGGckG0sG5y&index=14 This video, and the video following in the playlist, are a great primer on the Fitch-Margoliash method and distance matrix based methods in general for phylogenetic inference of branch lengths.

fasta/multiFa.go
```diff
@@ -166,3 +166,89 @@ func ChangePrefix(records []Fasta, prefix string) {
...
+		if seq1.Seq[i] == seq2.Seq[i] {
+			if seq1.Seq[i] != dna.Gap {
+				seq1Indel = false
+				seq2Indel = false
+				baseCount++
+			}
+		} else if seq1.Seq[i] == dna.Gap {
+			seq2Indel = false
+			if !seq1Indel {
+				seq1Indel = true
```
RavenYL, line 191: This is not related to the current PR but I was wondering if there's a way to keep track of indel sizes and use them in programs like multiFaAcceleration, like maybe in the stage where velocity gets normalized, a bigger indel size can become a multiplier that gets multiplied to the velocity (but not initial velocity) and creates a bigger acceleration (when we calculate a=v-v0)?

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+					distanceCache[currDistances] = alternatingLeastSquares(currDistances, s)
+				}
+				//no our distances should be in the cache, and we can do a simple lookup.
+				b1 = distanceCache[currDistances].B1
+				b3 = distanceCache[currDistances].B3
+
+				if !reachedEnd { //if we haven't run out of chromosome, we have another valid window to appear in our output.
+					velSum += b1
+					initialSum += b3
+					branchCacheSlice = append(branchCacheSlice, BranchCache{referenceCounter, referenceCounter + s.WindowSize, b1, b3})
```
RavenYL, line 137: I think I am not completely understanding the code - how does the sliding window work at the end of a chromosome when there are 4 species and their chromosomes are not the same length? When 1 species runs out of chromosome, does that mean no more windows will appear? And then because the multiFa format breaks the chromosome up into blocks, you repeat the sliding windows for each block?

RavenYL (dismissed):
Thank you so much for walking us through the PR at the high level in lab today! So glad that the program is working now! I just have a few questions as I'm thinking about this so I am approving with comments

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+
+	for alignmentCounter := 0; reachedEnd == false && referenceCounter < referenceLength-s.WindowSize; alignmentCounter++ {
+		if s.Verbose && alignmentCounter%1000000 == 0 {
+			log.Printf("alignmentCounter: %v\n", alignmentCounter)
+		}
+		currCount, pass = thresholdCheckPasses(s, currCount, threshold, bitArray, referenceCounter) //first we see if we have a valid window.
+		if records[0].Seq[alignmentCounter] != dna.Gap {                                            //and if we are at a reference position.
+			if pass {
+				if s.UseSnpDistance {
+					reachedEnd = fourWaySnpDistances(records, alignmentCounter, s, &currDistances)
```
dasnellings, line 122: Not a correctness problem, but this should be a target for any future optimization. Unless I am missing something, it seems like this recalculates distance from scratch for each window instead of just looking at the delta of the new bases after sliding down a base. I don't know how much of the computation is calculating the distances, but if it is significant, then the delta approach would be a massive performance boost.

cmd/multiFaAcceleration/testdata/generateTests.csh
```diff
@@ -0,0 +1,7 @@
+#!/bin/csh -ef
```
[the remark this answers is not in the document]
rimangan: yes this should go in documentation and I'm happy to remove this. Maybe a comment in the test program

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	if len(records[1].Seq) != len(records[0].Seq) || len(records[2].Seq) != len(records[0].Seq) || len(records[3].Seq) != len(records[0].Seq) {
+		log.Fatalf("Error. All records must be of the same sequence length.")
+	}
+
+	referenceLength = fasta.AlnPosToRefPos(records[0], len(records[0].Seq)-1)
+
+	if s.SearchSpaceBed != "" {
+		//if we want to limit our search to a particular set of regions, we populate the bitArray with 1s instead of 0s at positions overlapping the input bed entries.
+		searchSpace = bed.Read(s.SearchSpaceBed)
+		bitArray = make([]byte, referenceLength)
```
craiglowe, line 91: I was thinking that calling it bitArray was to let the reader know it was not being used as an array of bytes, but rather that each bit would be used for a position.  However, I am now thinking we are only using the least-significant bit in the byte.  Maybe it would be more clear if this was a slice of bools?

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+		currCount, pass = thresholdCheckPasses(s, currCount, threshold, bitArray, referenceCounter) //first we see if we have a valid window.
+		if records[0].Seq[alignmentCounter] != dna.Gap {                                            //and if we are at a reference position.
+			if pass {
+				if s.UseSnpDistance {
+					reachedEnd = fourWaySnpDistances(records, alignmentCounter, s, &currDistances)
+				} else {
+					reachedEnd = fourWayMutationDistances(records, alignmentCounter, s, &currDistances)
+				}
+
+				if _, containedInMap = distanceCache[currDistances]; !containedInMap { //if this tree has not been seen before, calculate branch lengths
```
craiglowe, line 127: I think this syntax can be hard for new programmers to read (people can chime in if it is not).  Maybe split into two lines?

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+					distanceCache[currDistances] = alternatingLeastSquares(currDistances, s)
+				}
+				//no our distances should be in the cache, and we can do a simple lookup.
```
craiglowe, line 130: no -> now

cmd/multiFaAcceleration/multiFaAcceleration.go
craiglowe: Dan has a good point if you find you want to speed this up.
Another minor thing you could think about if you need more speed would be passing in initial values for beginning the optimization.  I think the optimal values of the previous window will be a good guess at the optimal values of the current window.

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+					velSum += b1
+					initialSum += b3
+					branchCacheSlice = append(branchCacheSlice, BranchCache{referenceCounter, referenceCounter + s.WindowSize, b1, b3})
+				}
+			}
+			referenceCounter++
+		}
+	}
+
+	var averageVel, averageInitialVel, b1Normal, b3Normal float64
```
craiglowe, line 144: This is getting to be a long function.  Could this be the start of a new function here?

craiglowe (commented):
Great and really exciting work!  I have not made it all the way through, but wanted to post the comments I had so far.

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	if d1 == 0 {
+		if d2 == 0 {
+			return numbers.MaxFloat64(0, (s.ZeroDistanceWeightConstant*(d1-v1)+s.ZeroDistanceWeightConstant*(d2-v2))/(2*s.ZeroDistanceWeightConstant))
+		} else {
+			return numbers.MaxFloat64(0, s.ZeroDistanceWeightConstant*(d1-v1)+(1.0/math.Pow(d2, 2)*(d2-v2))) / (s.ZeroDistanceWeightConstant + (1.0 / math.Pow(d2, 2)))
+		}
+	} else if d2 == 0 {
+		return numbers.MaxFloat64(0, (1.0/(math.Pow(d1, 2))*(d1-v1)+s.ZeroDistanceWeightConstant*(d2-v2))/((1.0/math.Pow(d1, 2))+s.ZeroDistanceWeightConstant))
+	}
+	return numbers.MaxFloat64(0, (1.0/(math.Pow(d1, 2))*(d1-v1)+(1.0/math.Pow(d2, 2)*(d2-v2)))/((1.0/math.Pow(d1, 2))+(1.0/math.Pow(d2, 2))))
```
craiglowe, line 254: I am getting a little confused with the notation.  This function is operating on 'd' and 'v', but in the attached pdf, I see it operating on 'D' and 'v'.  However, I do see that 'D' numbers are being passed into the function, which then get re-named to 'd'.

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+		sub.vb = nonNegativeApproximation(sub.Dab, sub.Dbc, sub.va, sub.vc, s)
+	}
+	return sub.va, sub.vb, sub.vc
+}
+
+//If we constrain branch lengths to be nonNegative, we apply this correction when the minimum Q is achieved at negative branch lengths for a subtree.
+func nonNegativeApproximation(d1 float64, d2 float64, v1 float64, v2 float64, s Settings) float64 {
+	if d1 == 0 {
+		if d2 == 0 {
+			return numbers.MaxFloat64(0, (s.ZeroDistanceWeightConstant*(d1-v1)+s.ZeroDistanceWeightConstant*(d2-v2))/(2*s.ZeroDistanceWeightConstant))
```
craiglowe, line 247: Do we have these four different formulas in this function because we need to check if d1 or d2 is zero so that we can know if we need to calculate their weights or use s.ZeroDistnaceWeightConstant?  Can we declare variables for the two weights, then have two if-else statements where we check d1/d2 and then set the weight variable one with either the formula or the constant.  Then we just have one long equation where the weight is a variable that has already been set.  I think that may be more straight forward and easier to see any bugs.

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	} else if d2 == 0 {
+		return numbers.MaxFloat64(0, (1.0/(math.Pow(d1, 2))*(d1-v1)+s.ZeroDistanceWeightConstant*(d2-v2))/((1.0/math.Pow(d1, 2))+s.ZeroDistanceWeightConstant))
+	}
+	return numbers.MaxFloat64(0, (1.0/(math.Pow(d1, 2))*(d1-v1)+(1.0/math.Pow(d2, 2)*(d2-v2)))/((1.0/math.Pow(d1, 2))+(1.0/math.Pow(d2, 2))))
+}
+
+//Reduce the four species tree to the subtree containing species 0, 1, and the ancestor of 2/3.
+func pruneLeft(d Distances, b BranchLengths, sub *SubTree, s Settings) {
+	sub.Dab = d.D01
+	if d.D03 == 0 {
```
craiglowe, line 260: Similar to my above comment, I think all these long and separate equations makes it difficult to pick out bugs.  Can we make the weights variables, which get set with if-else statements and then we just have one equation?

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	} else {
+		sub.Dbc = ((1.0/math.Pow(d.D12, 2))*(d.D12-b.B4) + (1.0/math.Pow(d.D13, 2))*(d.D13-b.B5)) / ((1.0 / math.Pow(d.D13, 2)) + (1.0 / math.Pow(d.D12, 2)))
+	}
+}
+
+//Reduce the four species tree to the subtree containing 2, 3, and the ancestor of 0/1.
+func pruneRight(d Distances, b BranchLengths, sub *SubTree, s Settings) {
+	sub.Dac = d.D23
+
+	if d.D02 == 0 {
```
craiglowe, line 289: Similar to my above comment, I think all these long and separate equations makes it difficult to pick out bugs.  Can we make the weights variables, which get set with if-else statements and then we just have one equation?

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	if referenceCounter == 0 {
+		currCount = 0
+		for i := 0; i < s.WindowSize; i++ {
+			currCount += int(bitArray[i])
+		}
+	} else {
+		if bitArray[referenceCounter-1] == 1 {
+			currCount--
+		}
+		if bitArray[referenceCounter+s.WindowSize-1] > 0 {
```
craiglowe, line 364: Are just 1s and 0s in the bitArray?  I see here we check for greater than 0.

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+			currCount++
+		}
+	}
+	return currCount, currCount >= threshold
+}
+
+//Generate distances from mutation distances, which includes SNPs and INDELs, where each INDEL counts as one mutation regardless of length.
+func fourWayMutationDistances(records []fasta.Fasta, alignmentCounter int, s Settings, d *Distances) bool {
```
craiglowe, line 372: Maybe we should not use 'd' for the variable name here because I think this is different from the 'd' variable in the attached pdf

cmd/multiFaAcceleration/testdata/test.accel.searchspace.expected.bed
```diff
@@ -0,0 +1,151 @@
+chr1	175	225	3.810847e+00
```
craiglowe, line 1: Where do these "true" values come from?  They are not pretty, so I am not thinking they were human-calcualted.  Are they from another program we "trust"?

fasta/multiFa.go
```diff
@@ -166,3 +166,89 @@ func ChangePrefix(records []Fasta, prefix string) {
...
+func PairwiseMutationDistanceReferenceWindow(seq1 Fasta, seq2 Fasta, alnStart int, windowSize int) (int, bool, int) {
+	diff := 0
+	baseCount := 0
+	var seq1Indel bool = false
+	var seq2Indel bool = false
+	var reachedEnd bool = false
+	var i int = 0
+
+	for i = alnStart; baseCount < windowSize && i < len(seq1.Seq); i++ {
+		if seq1.Seq[i] == seq2.Seq[i] {
```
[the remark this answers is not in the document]
craiglowe: Good question.  How do we handle Ns?

craiglowe (commented):
Great work!  I finished my comments.

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	D01 float64
+	D02 float64
+	D03 float64
+	D12 float64
+	D13 float64
+	D23 float64
+}
+
+//The set of branch lengths corresponding to a particular distance matrix.
+type BranchLengths struct {
```
cshoben, line 54: I'm realizing it is a bit confusing to conceptualize the additions behind the distance calculations (D01, etc) because the distances are zero based and the branches are one based. Have you thought about having everything zero based for consistency?

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	//variables for normalization
+	var velSum, initialSum float64 = 0, 0
+	var branchCacheSlice = make([]BranchCache, 0)
+
+	for alignmentCounter := 0; reachedEnd == false && referenceCounter < referenceLength-s.WindowSize; alignmentCounter++ {
+		if s.Verbose && alignmentCounter%1000000 == 0 {
+			log.Printf("alignmentCounter: %v\n", alignmentCounter)
+		}
+		currCount, pass = thresholdCheckPasses(s, currCount, threshold, bitArray, referenceCounter) //first we see if we have a valid window.
+		if records[0].Seq[alignmentCounter] != dna.Gap {                                            //and if we are at a reference position.
```
cshoben, line 119: For conceptualizing; can you give an example of when we would be a position that wasn't a reference position?

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+	}
+	if d.D23 != 0 {
+		sum += math.Pow(d.D23-b.B4-b.B5, 2) / math.Pow(d.D23, 2)
+	} else {
+		sum += math.Pow(d.D23-b.B4-b.B5, 2) * s.ZeroDistanceWeightConstant
+	}
+	return sum
+}
+
+//bitArray is on reference coordinates, not alignment coordinates, so the window is simply equal to windowSize.
```
cshoben, line 350: Could you help me understand the difference between reference coordinates and alignment coordinates?

cmd/multiFaAcceleration/multiFaAcceleration.go
cshoben: I believe so. There is some code lines 88-98 in this file that talk about setting zeros to ones if a search space bed is given.

cmd/multiFaAcceleration/multiFaAcceleration.go
```diff
@@ -0,0 +1,504 @@
...
+		return 0, true
+	}
+	if referenceCounter == 0 {
+		currCount = 0
+		for i := 0; i < s.WindowSize; i++ {
+			currCount += int(bitArray[i])
+		}
+	} else {
+		if bitArray[referenceCounter-1] == 1 {
+			currCount--
```
cshoben, line 362: I dont follow why we would subtract here. Could you elaborate on this please? Is it because we are trying to get it to zero?

cshoben (commented):
Riley, the pdf document to read ahead of time was SUPER helpful! I didnt get to finish reviewing this yet but going to go ahead and publish the questions I have so far.

cmd/multiFaAcceleration/multiFaAcceleration.go
RavenYL: the when maf is converted to multi-fa format, each sequence is the same length, and insertions and deletions are noted by - in reference or one of the aligning sequences

RavenYL (dismissed):

craiglowe (approved):
Looks very nice!  Thank you for addressing many of the comments.  I am approving, but I want to leave a note that we still need test against hand-calculated examples, or output from a program we trust.

craiglowe (approved):
Nice work to get that bug fixed.  Let's continue to keep more tests in the plans

RavenYL (approved):

cshoben (approved):

PR #524, opened 2021-07-26
bedMerge mergeAdjacent

rimangan:
New option for bedMerge, a cmd that merges overlapping entries in a bed file. The new option is mergeAdjacent, which when set to true, also merges entries that are adjacent (not overlapping, but no gap between entries). Passes tests.

### DIFF
--- a/bed/compare.go
+++ b/bed/compare.go
@@ -38,6 +38,17 @@ func MergeBeds(bedFile []Bed) []Bed {
 	return bedFile
 }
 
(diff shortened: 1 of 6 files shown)


dasnellings (approved):

craiglowe (approved):
Great work!
Is there any time we would not want to merge adjacent entries?  It seems just as natural to me to merge 10-15 & 12-20 into 10-20 as it is to merge 10-15 & 15-20 into 10-20.

RavenYL (approved):

PR #525, opened 2021-07-27
bedToWig Unified Cmd

rimangan:
Gonomics main currently has a cmd for bedScoreToWig, which creates wig objects using the score field from bed entries, and bedReadsToWig, which creates wigs using the read counts (used as an intermediate step in samToWig). I wanted to introduce a third option, bedNameToWig, which parses float64 values from the name field and treats them like bedScoreToWig, as the score field does not have floating point support. 
With this many functions, I've moved to pull all three of these commands under one roof, and introduce bedToWig as the unified cmd.


wig/compare.go
```diff
@@ -40,3 +41,29 @@ func AllEqual(alpha []Wig, beta []Wig) bool {
 	}
 	return true
 }
+
+//Compare returns zero for equal wigs and otherwise returns the ordering of the two wig entries. Used for SortByCoord.
+func Compare(alpha Wig, beta Wig) int {
```
dasnellings, line 46: you might consider making this a private function

wig/compare.go
```diff
@@ -40,3 +41,29 @@ func AllEqual(alpha []Wig, beta []Wig) bool {
...
+	if len(alpha.Values) < len(beta.Values) {
+		return -1
+	}
+	if len(alpha.Values) > len(beta.Values) {
+		return 1
+	}
+	return 0
+}
+
+//SortByCoord sorts in place a slice of Wig structs by their genomic position.
```
dasnellings, line 66: might be worth saying that this function does a secondary sort on len(Values)

convert/convert.go
```diff
@@ -78,6 +78,49 @@ func SamToBedFrag(s sam.Sam, fragLength int, reference map[string]chromInfo.Chro
...
+		currentWig.Values = make([]float64, v.Size)
+		for x = 0; x < v.Size; x++ {
+			currentWig.Values[x] = Missing
+		}
+		wigSlice[i] = currentWig
+		i++
+	}
+	//loop through bed line at a time
+	file := fileio.EasyOpen(inFile)
+	defer file.Close()
```
dasnellings, line 105: I advise against using defer when closing files opened with `fileio`

convert/convert.go
```diff
@@ -78,6 +78,49 @@ func SamToBedFrag(s sam.Sam, fragLength int, reference map[string]chromInfo.Chro
...
+		startNum = common.StringToInt(words[1])
+		endNum = common.StringToInt(words[2])
+		current = &bed.Bed{Chrom: words[0], ChromStart: startNum, ChromEnd: endNum}
+		if len(words) >= 4 {
+			current.Name = words[3]
+		}
+		chromIndex = getWigChromIndex(current.Chrom, wigSlice)
+		midpoint = bedMidpoint(current)
+		if wigSlice[chromIndex].Values[midpoint] != Missing {
+			log.Fatalf("Multiple scores for one position.")
```
dasnellings, line 117: Consider making this more informative. I assume this would only occur if you have overlapping beds. Maybe indicate that people should run bedMerge?

convert/convert.go
```diff
@@ -78,6 +78,49 @@ func SamToBedFrag(s sam.Sam, fragLength int, reference map[string]chromInfo.Chro
...
+		i++
+	}
+	//loop through bed line at a time
+	file := fileio.EasyOpen(inFile)
+	defer file.Close()
+	for line, doneReading = fileio.EasyNextRealLine(file); !doneReading; line, doneReading = fileio.EasyNextRealLine(file) {
+		words := strings.Split(line, "\t")
+		startNum = common.StringToInt(words[1])
+		endNum = common.StringToInt(words[2])
+		current = &bed.Bed{Chrom: words[0], ChromStart: startNum, ChromEnd: endNum}
```
dasnellings, line 110: Why a pointer?

convert/convert.go
```diff
@@ -78,6 +78,49 @@ func SamToBedFrag(s sam.Sam, fragLength int, reference map[string]chromInfo.Chro
...
+	var chromIndex int
+	var midpoint int
+	var startNum, endNum int
+	var x int
+	var i int = 0
+	var doneReading bool = false
+	var current *bed.Bed
+	var currentWig wig.Wig
+	//generate Wig skeleton from reference
+	for _, v := range reference {
```
craiglowe, line 94: This could be a separate function (give a slice of chroms and get back a slice of wigs)

convert/convert.go
craiglowe: It looks like we are borrowing some bed-reading code.  Can we call a function that reads beds into a chan and then this loop pulls beds out of that chan?

convert/convert.go
```diff
@@ -78,6 +78,49 @@ func SamToBedFrag(s sam.Sam, fragLength int, reference map[string]chromInfo.Chro
...
+		current = &bed.Bed{Chrom: words[0], ChromStart: startNum, ChromEnd: endNum}
+		if len(words) >= 4 {
+			current.Name = words[3]
+		}
+		chromIndex = getWigChromIndex(current.Chrom, wigSlice)
+		midpoint = bedMidpoint(current)
+		if wigSlice[chromIndex].Values[midpoint] != Missing {
+			log.Fatalf("Multiple scores for one position.")
+		}
+		wigSlice[chromIndex].Values[midpoint] = common.StringToFloat64(current.Name)
```
craiglowe, line 119: Can bedNameToWig and bedScoreToWig be the same function.  I think they may be identical, besides this line where we could have an "if" to either take the number from the score or the name field.

craiglowe (commented):
Looks great!  A few small comments

convert/convert.go
rimangan: Not necessarily, in this function, two bed entries are allowed to overlap but cannot have the same midpoint, which can happen a couple of different ways. I'll make a better error message

convert/convert.go
rimangan: Yes, this function predates the bed.GoReadToChan function. will implement in the next commit

dasnellings (approved):

RavenYL (approved):

craiglowe (approved):
Great work!

PR #526, opened 2021-08-02
Pileup

dasnellings:
Function and cmd to get base counts per position from a SAM/BAM file. Similar to `samtools mpileup`. The actual function is fairly straightforward, but it got complex when I had to implement a buffer for handling the Pile struct.


cmd/pileup/pileup.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/vertgenlab/gonomics/sam"
+	"log"
+	"strings"
+)
+
+func usage() {
+	fmt.Print(
+		"pileup - Count bases from sequencing data\n\n" +
+			"Usage:\n" +
+			"  pileup [options] in.bam\n\n" +
```
craiglowe, line 20: Should we have an output file?  I am guessing this prints to stdout, but I am unsure of the format?  Maybe a quick description or a few example lines of output?

cmd/pileup/pileup.go
```diff
@@ -0,0 +1,117 @@
...
+			prevRead = read
+		}
+		close(sendChan)
+	}(reads, recycle, sendChan)
+
+	pileChan := sam.GoPileup(sendChan, header, false, nil, pileFilters(s))
+
+	var err error
+	output := fileio.EasyCreate(outfile)
+	_, err = fmt.Fprintln(output, "Chr\tPos\tA\tC\tG\tT\tN\tDEL\tINS")
```
craiglowe, line 71: I am thinking this is the header line.  Can we start this with a '#'?  I don't think there is a set standard, but I think a decent amount of file formats follow that convention.  This is also the information I think would be helpful to have in the help message

cmd/pileup/pileup.go
```diff
@@ -0,0 +1,117 @@
...
+	s.Reset()
+	s.WriteString(fmt.Sprintf("%s\t%d\t%d\t%d\t%d\t%d\t%d\t%d", h.Chroms[pile.RefIdx].Name, pile.Pos,
+		pile.Count[dna.A], pile.Count[dna.C], pile.Count[dna.G], pile.Count[dna.T], pile.Count[dna.N], pile.Count[dna.Gap]))
+	for seq, count := range pile.InsCount {
+		s.WriteString(fmt.Sprintf("\t%s:%d", seq, count))
+	}
+	return s.String()
+}
+
+var s *strings.Builder
```
craiglowe, line 92: Global variables make me nervous.  Can we avoid this?  Is it possible to pass this around?  I am guessing this was done for efficiency, but I am worried about it leading to bugs as the code is maintained.  I am wondering if it might also help with readability and maintenance to have longer variable names.

cmd/pileup/pileup.go
```diff
@@ -0,0 +1,117 @@
...
+	log.SetFlags(log.Ldate | log.Ltime)
+
+	if len(flag.Args()) != 1 {
+		usage()
+		return
+	}
+
+	infile := flag.Arg(0)
+
+	s := Settings{
```
craiglowe, line 112: I am worried that this use of scoping and shadowing can make the code harder to read/maintain.  I think what we are doing right here is that before this line we have a variable "s" that is a strings.Builder, but now on this line we shadow it, making that variable inaccessible and instead make "s" a struct of settings.

sam/pileup.go
```diff
@@ -0,0 +1,362 @@
...
+// Pile stores the number of each base observed across multiple reads.
+// The number of a particular base can be retrieved by using the dna.Base
+// as the index for Pile.Count. E.g. the number of reads with A is Count[dna.A].
+type Pile struct {
+	RefIdx   int
+	Pos      uint32
+	Count    [13]int        // Count[dna.Base] == Number of observed dna.Base
+	InsCount map[string]int // key is insertion sequence as string
+
+	touched bool // true if Count or InsCount has been modified
```
craiglowe, line 21: I am not totally sure how I would use this if maintaining the code.  Maybe an additional sentence?

sam/pileup.go
```diff
@@ -0,0 +1,362 @@
...
+// addInsertion to pileBuffer
+func addInsertion(pb *pileBuffer, refidx int, startPos uint32, seq []dna.Base) {
+	pb.addIns(refidx, startPos, dna.BasesToString(seq))
+}
+
+// ****** pileBuffer functions below ****** //
+
+// pileBuffer stores a slice of piles for efficient reuse.
+type pileBuffer struct {
+	s   []Pile // discontinuous positions are not allowed
```
craiglowe, line 129: So many variables named "s", ha ha ha.  Is the "s" for slice here?

craiglowe (commented):
Very cool!  That buffer to hold the pileup positions so that you can completely process the whole read at once is elegant.  I had some comments that were not based on correctness, but readability and maintenance.

edotau (dismissed):
LGTM

RavenYL (approved):

cmd/pileup/pileup.go
dasnellings: Yes, added

cmd/pileup/pileup.go
dasnellings: I have changed to passing the builder instead of a global variable. Also switched to longer variable names.

cmd/pileup/pileup.go
dasnellings: the builder is passed now.

sam/pileup.go
dasnellings: added

sam/pileup.go
dasnellings: yeahhhh, probably should change that. Changed to 'buf' for buffer

sam/pileup.go
```diff
@@ -0,0 +1,366 @@
...
+		newpb.buf[i] = pb.buf[pb.idx]
+		pb.idx = pb.incrementIdx()
+	}
+	*pb = *newpb
+}
+
+// incrementIdx returns the idx of the pile after pb.idx
+func (pb *pileBuffer) incrementIdx() int {
+	newidx := pb.idx + 1
+	if newidx == len(pb.buf) {
```
craiglowe, line 333: Now I am pretty sure this is a circular buffer.  I think there a decent amount of added complexity here to implement the circularity.  What do you think about a linked list of Piles?  This way you can quickly move things from the front to the end (I think this is the reason you made it circular), but you still end up with a linear data structure, so you don't need the code to deal with where the start is located (always at head) and you don't need the code to skip from the end to the beginning.

craiglowe (commented):
Very cool!  I feel like this infrastructure could have a wide range of uses and performance increases!

dasnellings:
The most recent push transitions from using a slice buffer to a circular doubly-linked list. This change has minimal effects on performance (see benchmarks below) but greatly simplifies the code. This should be much nicer to review now. 

Also, I left the old slice buffer code in the testing files. Maybe it just makes sense to delete them, but for now they exist. 

```
BenchmarkPileupSlice-8             35665             30795 ns/op          167985 B/op        138 allocs/op
BenchmarkPileupLinkedList-8        36145             32851 ns/op          168891 B/op        151 allocs/op
```

sam/pileup.go
```diff
@@ -0,0 +1,366 @@
...
+func resetPile(p *Pile) {
+	p.RefIdx = -1
+	p.Pos = 0
+	if !p.touched {
+		return
+	}
+	for i := range p.Count {
+		p.Count[i] = 0
+	}
+	p.InsCount = make(map[string]int)
```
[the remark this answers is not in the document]
dasnellings: I had some trouble with this. Since the Piles are copied to the send channel, deleting the keys from the map could lead to concurrency errors as the map is passed by reference. However, you make a good point that the map will be empty most of the time, so I have transitioned to only making the map when it is needed, which should cut down on allocations.

cmd/pileup/pileup.go
```diff
@@ -0,0 +1,116 @@
...
+		})
+	}
+
+	return filters
+}
+
+func pileup(infile string, outfile string, settings Settings) {
+	sBuilder := new(strings.Builder)
+	reads, recycle, header := sam.GoReadToChanRecycle(infile, 1000)
+	sendChan := make(chan sam.Sam) // no buffer to satisfy recycle contract
```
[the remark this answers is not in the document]
dasnellings: The problem was that the send function needs to know when a read is finished with. It seemed the simplest way to ensure the previous read had been processed was to make the channel unbuffered so that it would block until the calling function was ready for the next read (ensuring that the previous was finished).

cmd/pileup/pileup.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// goroutine sends from the read channel to an unbuffered channel
+	// then returns the previous struct to the recycle channel.
+	go func(<-chan *sam.Sam, chan<- *sam.Sam, chan<- sam.Sam) {
```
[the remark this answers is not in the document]
dasnellings: It would be easier to have the pileup function take the read and recycle channels, but I did not want to lock people into using the recycle version of GoReadToChan. This was kind of a middle ground where we got a lot of the benefit of recycling, but could still use the normal GoReadToChan which is a `chan Sam` instead of a `chan *Sam`.

One thing on my to-do list is to rework the GoReadToChanRecycle functions. I wrote them to pass pointers to structs because I wanted to reuse as many fields as possible, but because the largest fields (and the ones most likely to cause GC issues) are already passes as references, it may make mores sense to have the recycle functions operate with chans of structs rather than pointers to get around the issue in this PR of designing for both cases.

RavenYL (approved):

edotau (dismissed):

sam/pileup.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	return true
+}
+
+func pileupLinked(send chan<- Pile, reads <-chan Sam, header Header, includeNoData bool, readFilters []func(s Sam) bool, pileFilters []func(p Pile) bool) {
+	start := newLinkedPileBuffer(300) // initialize to 2x std read size
+	refmap := chromInfo.SliceToMap(header.Chroms)
+	for read := range reads {
+		if !passesReadFilters(read, readFilters) {
+			continue
```
cshoben, line 79: Just to be sure I understand the "continue"; if we hit this continue, we continue to line 81, otherwise we go back up to line 77?

cshoben (commented):
Partially done reviewing, going to submit the comments I have so far.

sam/pileup.go
```diff
@@ -0,0 +1,273 @@
...
+// list in which Pile.next/prev is the next/prev contiguous position. Discontiguous
+// positions are not allowed. When adding base information to the pile struct, the
+// pointer to the start of the read position should be kept, and a second pointer
+// will increment along the list as bases are added. The pointer to the beginning
+// of the list should only be incremented when a position is passed and set to zero.
+type Pile struct {
+	RefIdx   int
+	Pos      uint32
+	Count    [13]int        // Count[dna.Base] == Number of observed dna.Base
+	InsCount map[string]int // key is insertion sequence as string
```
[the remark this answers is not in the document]
dasnellings: The map is structured as map[insertionSequence] = numberOfReadsWithInsertionSequence

I have updated the comment to make that clearer

sam/pileup.go
```diff
@@ -0,0 +1,273 @@
...
+	// has been used. This value is used to avoid unnecessary allocations
+	// in resetPile, and is used in checkSend to avoid sending zeroed structs
+	// that happen to pass the user-defined filters.
+
+	next *Pile
+	prev *Pile
+}
+
+// GoPileup inputs a channel of coordinate sorted Sam structs and generates
+// a Pile for each base. The Pile is send through the return channel when
```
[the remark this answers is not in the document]
dasnellings: nice catch

sam/pileup.go
```diff
@@ -0,0 +1,273 @@
...
+// the Pile position is no longer being updated.
+//
+// The input readFilter functions must all be true for a read to be included.
+// The input pileFilter functions can be likewise be used to filter the output.
+// All reads/piles are included if filters == nil.
+func GoPileup(reads <-chan Sam, header Header, includeNoData bool, readFilters []func(s Sam) bool, pileFilters []func(p Pile) bool) <-chan Pile {
+	if header.Metadata.SortOrder[0] != Coordinate {
+		log.Fatal("ERROR: (GoPileup) input sam/bam must be coordinate sorted")
+	}
+	pileChan := make(chan Pile, 1000)
```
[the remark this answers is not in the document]
dasnellings: Correct, the buffer is arbitrarily large. Because we are using goroutines there are two threads operating concurrently: a 'reader' thread, and a 'worker' thread. Depending on the read either the reader or the worker may finish first. If the channel was unbuffered then the thread that finished first would be forced to wait until the other thread is finished before working on the next read. The buffer makes it so once one of the thread finishes, it can go on to the next read without waiting for the other thread.

sam/pileup.go
dasnellings: Opposite. If we hit continue we go back to line 77, else we go to line 81. 

This is basically saying "if the read does not pass filters, go ahead and move on to the next read"

cmd/pileup/pileup.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/sam"
+	"log"
+	"strings"
+)
+
+func usage() {
+	fmt.Print(
+		"pileup - Count bases from sequencing data\n\n" +
+			"pileup prints to stdout by default, but can be redirected with the -o flag.\n" +
```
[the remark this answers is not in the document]
edotau: I was never a fan of the decision to avoid stdout, but I think I was outvoted. Generating pileups from bam alignments can be thought of as a form of data wrangling and it's more common (and easier) to use stdout. This meme is mostly a joke, but I think it captures this program design pretty well. Stdout is the quicker user experience in practice, and the well-paved side walk is the fileio (to file solution).

IMG_1126

edotau (approved):

cmd/pileup/pileup.go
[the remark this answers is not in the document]
craiglowe: I think there are a few things swirling around here and I will do my best to separate them out.

I think there are three general options for writing output:
1) Write it to stdout
2) Write it to stdout, but give an option (e.g. -o) so that it can be redirected within the program to a file
3) Write it to a file, where it can be sent to stdout by giving /dev/stdout (or stdout in our case) as the filename

There are similar options for input, but I will just talk about output, where I think most things I will write are similar for the case of inputs.

If we think about programs with a single input or output file/stream, I don't really think there is a huge difference, but I am open to hearing everyone's thoughts.  The options for the two general cases are of writing to stdout or a file for the three options are:

1)
program in.sam
program in.sam > out.bed
2)
program in.sam
program in.sam -o out.bed
3)
program in.sam /dev/stdout
program in.sam out.bed

I can see how people may have some preferences, but I find them all pretty similar.

While it is not a concern with the command-line interface itself, I do worry that option 1 can tempt us to put print statements deep in functions that print directly to stdout and don't take a handler as an argument.  This can lead to the code being difficult to re-use either as the program evolves over time, or as someone else wants to use it for a different program.

I think we start to see some differentiation in style when we have multiple output files, which while uncommon in typical gnu utilities, does happen at times in genomics software.  In this case I think it is good to use filenames (or at least allow the output to be directed to a file inside the program) so that it is possible to steam any of the inputs or outputs.

While the case of multiple output files seems separate, there have been times when a program first envisioned as a small one-output program, has turned out to be useful for related things and grown to provide additional outputs.  My personal preference for option 3), with option 2) being an equal/close second is not so much about the present, but maintaining the code over longer periods of time.

I think the core of this is how strict we want the style convention to be.  As I stated above, I think most people will have a preference about the three interfaces, but at this point in time I don't think this goes beyond a preference.  Style conventions come into play in areas like this where there are equally-good ways to do things, but the team as a whole can benefit from a uniform approach.  We can discuss at code review if we want a stricter or looser style convention here; I am open to either one and interested to get everyone's input.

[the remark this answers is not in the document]
edotau: Thank you for the insightful comments! I won't be in attendance of the discussion, but I'll add my two cents here. There is a way (in golang) around the pitfalls of option 1; one can use `log.Print()` and `fmt.Print()` to differentiate print statements from data results. However, I think I would prefer options 2 and/or 3 in most cases for the reasons @craiglowe mentioned.

What I push back on is being **_too strict_** about style conventions and/or standards. While I agree that teams benefit from a uniform approach, there is still a creative and artistic aspect to programming. People sacrifice a lot of time anre are often very passionate about their work and I feel like everyone should have the freedom to implement their style preferences (given there are no functional pitfalls).

cshoben (approved):

PR #527, opened 2021-08-04
Bed filter update

rimangan:
Some updates to the bedFilter cmd. Cmd has been modernized to use chans for reading and writing (previously held output entries in memory). Options have been organized into a settings struct. Two new options: minNameFloat and maxNameFloat, for filtering entries by value when using the name field to store floats.
Also took this opportunity to write tests for this function.


dasnellings:
This seems to have a lot of the acceleration code. Since that PR is marked as 'do not merge' should this PR be reviewed yet?

craiglowe:
I am seconding Dan's comments about the acceleration code.  Should we review the acceleration code, or should this also be "do not merge" for right now?

PR #528, opened 2021-08-16
bedFilter Subset and Testing

rimangan:
bedFilter modernization, including using channels and settings structs. New option for subsetting bed files to a pseudorandom proportion of bed records.
bedFilter on main is broken, as we declare Bed structs without setting the FieldsInitialized. This is the latest in a series of cmds to be broken due to the change in bed writing that has gone undetected as bedFilter did not have tests for the cmd. As such, I wrote a set of tests for this cmd as well.

### DIFF
--- a/cmd/bedFilter/bedFilter.go
+++ b/cmd/bedFilter/bedFilter.go
@@ -7,75 +7,68 @@ import (
 	"fmt"
 	"github.com/vertgenlab/gonomics/bed"
 	"github.com/vertgenlab/gonomics/common"
+	"github.com/vertgenlab/gonomics/exception"
 	"github.com/vertgenlab/gonomics/fileio"
 	"github.com/vertgenlab/gonomics/numbers"
 	"log"
-	"strings"
+	"math/rand"
 )
 
-func bedFilter(infile string, outfile string, minScore int, maxScore int, minLength int, maxLength int, minStart int, maxStart int, minEnd int, maxEnd int, chrom string) {
-	var outlist []bed.Bed
-	var line string
-	var startNum, endNum, length int
-	var doneReading bool = false
+func bedFilter(s Settings) {
+	common.RngSeed(s.RandSeed, s.SetSeed)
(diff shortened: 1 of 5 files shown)


RavenYL (approved):
Thank you for doing maintenance on the repository code! I just wanted to ask why you added the pseudorandom subsetting. I wonder when this function is needed?

dasnellings (approved):

cmd/bedFilter/bedFilter.go
```diff
@@ -7,75 +7,79 @@ import (
...
+	for curr := range bedChan {
 		pass = true
-		words := strings.Split(line, "\t")
-		startNum = common.StringToInt(words[1])
-		endNum = common.StringToInt(words[2])
-		length = endNum - startNum
-		if len(words) > 4 {
-			if common.StringToInt(words[4]) < minScore {
+		length = curr.ChromEnd - curr.ChromStart
+		if curr.FieldsInitialized > 4 {
```
craiglowe, line 29: Right now I think records without a score pass a score threshold.  Is this desired?  Maybe there should be an "else" that gives a pass = false?

cmd/bedFilter/bedFilter.go
```diff
@@ -7,75 +7,79 @@ import (
...
+				pass = false
+			}
+		}
+		if s.Chrom != "" {
+			if curr.Chrom != s.Chrom {
+				pass = false
+			}
+		}
+		if s.SubSet < 1.0 {
+			r = rand.Float64()
```
craiglowe, line 71: This step can be expensive.  Maybe we don't need to do it if pass is already false?

craiglowe (commented):
Great work!  One quick comment about a score filter being applied when the bed does not have a score.

edotau (approved):
LGTM!

cshoben (approved):

craiglowe (approved):
Great work!

PR #530, opened 2021-08-24
bed/compare.go compareDistance

cshoben:
I'm adding a function to the bed package to compare two beds and determine their minimum distance. This first checks that they are on the same chromosome, then checks for overlap, if there is any overlap the minimum distance returned is zero. If there is no overlap, we compare distances between starts and ends, and this function returns the value for whichever distance is positive. I do not know what to return if's fail? I could have my last if statement be an else, but I like this extra check in here in case I forgot about a corner case.

### DIFF
--- a/bed/compare.go
+++ b/bed/compare.go
@@ -172,19 +172,12 @@ func CompareDistance(a Bed, b Bed) int {
 	if Overlap(a, b) {
 		return 0
 	}
-<<<<<<< HEAD
 	if a.ChromStart - b.ChromEnd >= 0 { //only positive if bed "a" is downstream of "b" bed.
 		return a.ChromStart - b.ChromEnd +1
 	}
 	if b.ChromStart - a.ChromEnd >= 0 { //only positive if bed "b" is downstream of "a" bed.
 		return b.ChromStart - a.ChromEnd +1
-=======
-	if a.ChromStart-b.ChromEnd > 0 { //only positive if bed "a" is downstream of "b" bed.
-		return a.ChromStart - b.ChromEnd
-	}
-	if b.ChromStart-a.ChromEnd > 0 { //only positive if bed "b" is downstream of "a" bed.
-		return b.ChromStart - a.ChromEnd
->>>>>>> 78955c2e47bfb178283b7ff6c9a09395e6a3f7ad
+
 	}
 	log.Panic("something went wrong with CompareDistance")
 	return -1 // maybe just return -1 here to indicate that we can't count a distance?
--- a/bed/compare.go
+++ b/bed/compare.go
@@ -159,6 +159,24 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
 	}
 	return CompareChromEnd(a, b)
 }
+//CompareDistance compares beds by chromStart and chromEnd and returns the minimum distance between the two beds.
+func CompareDistance(a Bed, b Bed) int {
+	chromComp := strings.Compare(a.Chrom, b.Chrom)
+	if chromComp != 0 {
+		return chromComp  //maybe this should flag to tell the user the chrom are not the same, otherwise it just gives back a "-1" or "1".
+		// This also a potential problem for "func compare" starting at line 110 above.
+	}
+	if Overlap (a, b) {
+		return 0
+	}
+	if a.ChromStart - b.ChromEnd > 0 { //only positive if bed "a" is downstream of "b" bed.
+		return int(a.ChromStart - b.ChromEnd)
+	}
+	if b.ChromStart - a.ChromEnd > 0 { //only positive if bed "b" is downstream of "a" bed.
+		return int(b.ChromStart - a.ChromEnd)
+	}
+	return ??? // I dont know what to return here?
+}
 
 //AllAreEqual returns true if two input slices of Beds contain Bed entries that all return true for Equal.
 func AllAreEqual(a []Bed, b []Bed) bool {
--- a/bed/compare.go
+++ b/bed/compare.go
@@ -1,6 +1,7 @@
 package bed
 
 import (
+	"errors"
 	"github.com/vertgenlab/gonomics/numbers"
 	"log"
 	"sort"
@@ -166,8 +167,8 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
 //0 (overlap, minimum distance is 0), >=0 is the minimum distance.
 func CompareDistance(a Bed, b Bed) int {
 	if a.Chrom != b.Chrom {
-		log.Panic("chromosomes do not match between input beds")
-		return -1 //when I run a test on this it only returns the panic statement and not the "-1".
+		myError := errors.New("Could not calculate distance, different chromosomes")
+		return myError
 	}
 	if Overlap(a, b) {
 		return 0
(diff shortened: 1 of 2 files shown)


bed/compare.go
```diff
@@ -159,6 +159,24 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
...
+func CompareDistance(a Bed, b Bed) int {
+	chromComp := strings.Compare(a.Chrom, b.Chrom)
+	if chromComp != 0 {
+		return chromComp  //maybe this should flag to tell the user the chrom are not the same, otherwise it just gives back a "-1" or "1".
+		// This also a potential problem for "func compare" starting at line 110 above.
+	}
+	if Overlap (a, b) {
+		return 0
+	}
+	if a.ChromStart - b.ChromEnd > 0 { //only positive if bed "a" is downstream of "b" bed.
```
cfauci, line 172: I think that you might want lines 172-176 within the Overlap if statement. You only want to do that comparison if there is overlap right?

bed/compare.go
```diff
@@ -159,6 +159,24 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
 	}
 	return CompareChromEnd(a, b)
 }
+//CompareDistance compares beds by chromStart and chromEnd and returns the minimum distance between the two beds.
+func CompareDistance(a Bed, b Bed) int {
+	chromComp := strings.Compare(a.Chrom, b.Chrom)
+	if chromComp != 0 {
```
cfauci, line 165: you could rearrange this so you can return the chromComp at the end:

if chromComp == 0 {
     if Overlap (a, b) {
         if a.ChromStart...{
}
        if a.ChromStart...{
}
}
else {
return chromComp (or whatever tells the user that the beds aren't on the same chrom
}
return var (you could store the int from the comparisons in this var instead of returning them above or something there are options)
}

cshoben: I like the idea with the chromComp. The var part isn't necessary, is it?

cfauci: just a way to have a return there if you don't have anything else im sure someone will have a way around it tho

dasnellings: For what it's worth, I like the way you have it set up already. This is and example of the 'guard clause' we were talking about yesterday in code review. The early return saves you from doing more comparisons than needed, and helps reduce nested `if` statements which I always find hard to keep track of. The caveat is that you end up with multiple return's throughout the code which some people find jarring. Both are valid styles so no right answer here.

bed/compare.go
```diff
@@ -159,6 +159,24 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
...
+	if Overlap (a, b) {
+		return 0
+	}
+	if a.ChromStart - b.ChromEnd > 0 { //only positive if bed "a" is downstream of "b" bed.
+		return int(a.ChromStart - b.ChromEnd)
+	}
+	if b.ChromStart - a.ChromEnd > 0 { //only positive if bed "b" is downstream of "a" bed.
+		return int(b.ChromStart - a.ChromEnd)
+	}
+	return ??? // I dont know what to return here?
```
dasnellings, line 178: It's maybe not ideal, but one way to handle this is to return an integer where -1 is always 'different chromosomes' and any value >=0 is the inter-chromosomal distance. You could leave the code as-is except changing line 166 to `return -1`.

dasnellings: Also, for the problem of what to return at the end. It seems like all valid bed entries should be true for one of the above if statements so it you have not returned by line 178, something bad has happened and there is likely a bug in the code calling this function. One way to handle this is:

```
// rest of the function //

log.Panic("something went wrong")
return 0 // unreachable
```

cshoben: Ahh, great point! So I would just put the "log.Panic..." in place of the return or after the return? I think i like this more than the "-1" if this is something that would replace it. @dasnellings

cshoben:
I did another push, incorporating some feedback.

bed/compare.go
dasnellings: The log.Panic will kill the program so the final return is never actually used, it is only there to satisfy the compile requirement that a function with a returned value ends in a return. I put zero, but it can be anything.

bed/compare.go
```diff
@@ -160,6 +161,27 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
 	return CompareChromEnd(a, b)
 }
 
+//CompareDistance compares beds by chromStart and chromEnd and returns the minimum distance between
+//the two beds. return options; -1 (different chromosomes, no distance calculated),
+//0 (overlap, minimum distance is 0), >=0 is the minimum distance.
+func CompareDistance(a Bed, b Bed) int {
+	chromComp := strings.Compare(a.Chrom, b.Chrom)
+	if chromComp == 0 {
```
dasnellings, line 169: I think this should be `chromComp != 0`.

bed/compare.go
```diff
@@ -160,6 +161,27 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
...
+		return -1
+	}
+	if Overlap(a, b) {
+		return 0
+	}
+	if a.ChromStart-b.ChromEnd > 0 { //only positive if bed "a" is downstream of "b" bed.
+		return a.ChromStart - b.ChromEnd
```
craiglowe, line 176: Bed is left-closed right-open, so you will want a +1 at the end of this.

bed/compare.go
```diff
@@ -160,6 +161,27 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
...
+	}
+	if b.ChromStart-a.ChromEnd > 0 { //only positive if bed "b" is downstream of "a" bed.
+		return b.ChromStart - a.ChromEnd
```
craiglowe, line 179: Bed is left-closed right-open, so you will want a +1 at the end of this.

bed/compare.go
craiglowe: Since you are not ordering the chroms, but just testing for equality, you can also do "if a.Chrom != b.Chrom {"

bed/compare.go
```diff
@@ -160,6 +161,27 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
 	return CompareChromEnd(a, b)
 }
 
+//CompareDistance compares beds by chromStart and chromEnd and returns the minimum distance between
+//the two beds. return options; -1 (different chromosomes, no distance calculated),
+//0 (overlap, minimum distance is 0), >=0 is the minimum distance.
+func CompareDistance(a Bed, b Bed) int {
+	chromComp := strings.Compare(a.Chrom, b.Chrom)
+	if chromComp == 0 {
+		return -1
```
craiglowe, line 170: I would give an error here to be more explicit that some unexpected happened, or that we are unable to do what the programmer asked.  I think you could either log.Fatal, or you could update your function to return an error along with an int.  Probably an error along with an int will be more useful in your case and others.

craiglowe (commented):
Looks like a good start!

craiglowe (approved):
Great work!

craiglowe (approved):
Great work!

dasnellings (approved):

RavenYL (approved):

PR #531, opened 2021-08-26
Fastq filter names

rimangan:
New option for fastqFilter. If you provide a line-delimited file containing read names, only fastq entries with names found in this file will be retained. For paired end reads, a read will be retained if either the forward or reverse read is found in the names file. Useful for pulling out all reads that map to a particular chromosome or construct, or those with other filter properties.


dasnellings (dismissed):

RavenYL (approved):

craiglowe (commented):
Looks really good!  I didn't see the code that implements the discardNames filter.  Maybe I just missed it?  I do think I saw your last two tests testing that feature, so it must be there.  Where does that filter get applied?

cmd/fastqFilter/fastqFilter.go
```diff
@@ -90,8 +124,15 @@ func fastqFilter(s Settings) {
 			if len(i.Seq) > s.MaxSize {
 				continue
 			}
+			if s.RetainNamesList != "" {
```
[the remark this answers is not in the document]
rimangan: Oops! I was a commit behind, implemented now

craiglowe (commented):
Great work!
I also see a bunch of bedFilter code, which I think has already been checked in?  Can we merge main into this branch?

cmd/fastqFilter/fastqFilter.go
```diff
@@ -64,6 +92,21 @@ func fastqFilter(s Settings) {
...
+					if _, RevInMap = namesMap[i.Rev.Name]; !RevInMap {
+						continue //continue if neither the forward or reverse read is in the map
+					}
+				}
+			}
+			if s.DiscardNamesList != "" {
+				if _, FwdInMap = namesMap[i.Fwd.Name]; FwdInMap {
+					continue
+				}
+				if _, RevInMap = namesMap[i.Rev.Name]; !RevInMap {
```
craiglowe, line 106: I was expecting "RevInMap" instead of "!RevInMap" towards the end of the line.

craiglowe (commented):
Wonderful to have this functionality.  I think I found one small bug

cmd/fastqFilter/fastqFilter.go
rimangan: Great catch! Copy paste error. Fixed in next commit

craiglowe (approved):
Wonderful!

PR #533, opened 2021-09-01
catMultiFa

rimangan:
catMultiFa is a new cmd that works like the Unix cat cmd for multiple alignment fasta (multiFa) format files. For multiple multifa files with the same number and order of species, the individual sequences are concatenated to retain multiFa format and produce composite sequences. Useful for downstream population analysis for multiFa alignments from multiple regions. 
Accepts an arbitrary number of multiFa files, supporting syntax like:
"~/go/bin/catMultiFa *.fa"
Default output to stdout, but the cmd accepts an -o flag to redirect output to a file.
The user can also provide a list of filenames as a newline-delimited text file. This is useful when catting thousands of files to avoid the command line character limit.
Passes tests.


cmd/catMultiFa/catMultiFa.go
```diff
@@ -0,0 +1,89 @@
...
+			"Usage:\n" +
+			"catMultiFa multi.fa ...\n" +
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var minNumArgs int = 2
+	var o *string = flag.String("o", "stdout", "Redirect the output to a user-given filename.")
+	var list *string = flag.String("list", "", "")
```
dasnellings, line 55: You should point out that this needs to be a line delimited text file. I see it from the code, but I don't think it is obvious if you are just looking at the help message.

cmd/catMultiFa/catMultiFa.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if len(flag.Args()) < minNumArgs {
+		flag.Usage()
+		log.Fatalf("Error: expecting at least %d arguments, but got %d\n", minNumArgs, len(flag.Args()))
+	}
+	var fileList []string
+	var err error
+	if *list == "" {
+		fileList = make([]string, len(flag.Args()))
```
dasnellings, line 73: `fileList = flag.Args()`?

dasnellings (dismissed):
Looks good. Though I don't really understand the use case. Making composite sequences is a bit weird.

cmd/catMultiFa/catMultiFa.go
rimangan: lol way simpler

rimangan:
> Looks good. Though I don't really understand the use case. Making composite sequences is a bit weird.

This is indeed a bit niche. The current use case is for PCA on HAQERs, where we want to cluster sequences from multiple discontinuous aligned regions at once. The idea here is to use the multiFaExtract -bed to pull all alignments of HAQERs from the genome-wide alignment, and follow up with this to make one continuous sequence that can then feed into PCA. I could imagine this being used to analyze multiple discontinuous sequences together for other types of population analysis, Dunn or Tajima for example

cmd/catMultiFa/catMultiFa.go
```diff
@@ -0,0 +1,88 @@
...
+			"Usage:\n" +
+			"catMultiFa multi.fa ...\n" +
+			"OR\n" +
+			"catMultiFa -list file.list" +
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var minNumArgs int = 2
```
cshoben, line 55: Can you talk me through the logic behind the number of args? Like between here and how we reset it to zero if the user gives a list. Is it set to 2 here so the math at line 68 always works out?

cshoben (approved):
Awesome! And also I appreciate this bite size PR. A logic questions included with this approval.

cmd/catMultiFa/catMultiFa.go
```diff
@@ -0,0 +1,88 @@
...
+	var fileList []string
+	var err error
+	if *list == "" {
+		fileList = flag.Args()
+	} else {
+		file := fileio.EasyOpen(*list)
+		fileList = make([]string, 0)
+		var line string
+		var doneReading bool
+		for line, doneReading = fileio.EasyNextRealLine(file); !doneReading; line, doneReading = fileio.EasyNextRealLine(file) {
```
craiglowe, line 81: I think there is code that does this already (take a filename and read it into a slice where each line of the file is an element in the slice.  It is fileio.Read

craiglowe (commented):
Looks great.  I had just a few very small comments

cmd/catMultiFa/catMultiFa.go
rimangan: I'm clarifying the logic in the next commit, but the idea here is if you send files as arguments you need at least 2 in order to cat them together. If you use a file list instead, you do not need any arguments (in fact, it's forbidden in the next commit)

cmd/catMultiFa/catMultiFa.go
```diff
@@ -0,0 +1,88 @@
...
+	var o *string = flag.String("o", "stdout", "Redirect the output to a user-given filename.")
+	var list *string = flag.String("list", "", "User may supply a line-delimited list of files to concatenate instead of providing them as arguments.")
+	var lineLength *int = flag.Int("lineLength", 50, "Specify the line length in the output file.")
+
+	flag.Usage = usage
+	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	flag.Parse()
+
+	if *list != "" {
+		minNumArgs = 0
```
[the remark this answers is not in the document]
rimangan: The user cannot do both, so expectedNumArgs is 0 in this case. I'll rework in the next commit

cmd/catMultiFa/catMultiFa.go
rimangan: Great! Changing in next commit

craiglowe (approved):
Great work!

dasnellings (approved):

cshoben (approved):

PR #536, opened 2021-09-08
Afs Sampler Control

rimangan:
In SimulateVcf, we use bounded rejection sampling using a beta-distributed bounding function to generate random variates from the stationarity distribution. Previously, the parameters for this beta function were abstracted from the user and hard-coded. Now they are presented as options with updated default values to accommodate  the simulation of AFS data under strong positive selection. 
Also added tests for simulateVcf.


RavenYL (approved):

dasnellings (approved):

cmd/simulateVcf/simulateVcf.go
```diff
@@ -42,5 +57,17 @@ func main() {
 	}
 	outFile := flag.Arg(0)
 
-	simulateVcf(*alpha, *numAlleles, *numSites, outFile, *randSeed, *setSeed)
+	s := Settings{
+		OutFile:         outFile,
+		Alpha:           *alpha,
```
cshoben, line 62: Jw, why do we use a pointers here?

cshoben (approved):

PR #538, opened 2021-09-09
Merge Sort Bug Fixes, Testing, and Sorting Sams by Single-Cell Barcode

rimangan:
mergeSort is a gonomics cmd that takes in either a SAM, GIRAF, AXT, BED, or VCF and sorts the output using a memory efficient merge sort algorithm. mergeSort on main does not function for BED, AXT, or VCF due to a variety of bugs introduced when we removed pointers from those packages. This loss of functionality was unnoticed as this cmd did not have tests (this is an older cmd). 
This pull request fixes the errors for sorting BEDs, AXTs, and VCFs, provides accuracy and stability tests to prevent future changes from breaking the cmd, and introduces a new sortCriteria for SAM where we can sort single-cell format sam records by their cell barcode. This is sort of a specific use-case, but it is a necessary step in the single-cell counting program I'm working on.
Dependent on the previous pull request for parse single cell from SAM.


RavenYL (approved):

dasnellings (commented):
After our discussion I looked into it more and this sort is not stable so your tests could fail even if the sort is still valid. May best to test via correctness rather than an expected file.

Also, please start branching from main.

cshoben (approved):

dasnellings (approved):

PR #539, opened 2021-09-10
scCount

rimangan:
scCount is a gonomics alternative to cellRanger count, which has the advantage of having each step partitioned into a distinct command to allow user access to intermediate files for custom analysis pipelines. scCount takes in a preprocessed single-cell SAM file and returns a tab-separated value file of the UMI counts for each gene for each cell, currently in an uncompressed matrix (although this output is sparse and we may consider smarter encoding).

For preprocessing, a 10x-style fastq must first be filtered with fastqFilter -collapseUmi to remove PCR duplicates from library preparation. Then, the Bx and UMI information is added to the name with fastqFormat -singleCell. Now these edited reads are aligned to the genome with an aligner of choice and sorted with mergesort -singleCellBx.

At this point, scCount can accept aligned unique reads sorted by single cell. For each cell, it writes a line to the TSV file containing the number of reads that aligned to the position of each gene in an input gtf format gene file.

Dependent on my other open pull requests.


cmd/scCount/scCount.go
```diff
@@ -0,0 +1,165 @@
...
+		var val float64
+		for i := range r.Counts {
+			if val, ok = normalizationMap[geneIdSlice[i]]; ok { //if the current gene is in the normalization map.
+				r.Counts[i] = r.Counts[i] * val
+			}
+		}
+	}
+	var countString string = fmt.Sprintf("%g", r.Counts[0])
+	for i := 1; i < len(r.Counts); i++ {
+		countString = countString + fmt.Sprintf("\t%g", r.Counts[i])
```
craiglowe, line 114: You can do this all in Sprintf with: fmt.Sprintf("%s\t%g", countString, r.Counts[i])

cmd/scCount/scCount.go
```diff
@@ -0,0 +1,165 @@
...
+	exception.PanicOnErr(err)
+	tree := interval.BuildTree(geneIntervals)
+	var overlap []interval.Interval
+	var currRow Row
+	var currGene string
+	var firstTime bool = true
+	var normFlag bool = false
+
+	var normalizationMap map[string]float64 = make(map[string]float64, 0)
+	if s.ExpNormalizationFile != "" {
```
craiglowe, line 55: This "if" statement feels like it could be a separate function.

cmd/scCount/scCount.go
```diff
@@ -0,0 +1,165 @@
...
+
+	for g := range genes {
+		geneIdSlice = append(geneIdSlice, g)
+	}
+	sort.Strings(geneIdSlice)
+	for c, g := range geneIdSlice { //c for count, g for gene
+		headerString += fmt.Sprintf("\t%s", g)
+		geneIndex[genes[g].GeneID] = c
+		geneIntervals = append(geneIntervals, genes[g])
+		c++
```
craiglowe, line 41: I have not seen increments mixed with "range."  Is this to do only the odd genes?

craiglowe (commented):
Looks like this is coming along well!  I added a few comments.

cmd/scCount/scCount.go
```diff
@@ -0,0 +1,165 @@
...
+func getGeneName(g *gtf.Gene) string {
+	return g.GeneID
+}
+
+func printRow(out io.Writer, r Row, normalizationMap map[string]float64, geneIdSlice []string, normFlag bool) {
+	if normFlag { //if the user passed in a normalization file, we normalize the count for each gene.
+		var ok bool
+		var val float64
+		for i := range r.Counts {
+			if val, ok = normalizationMap[geneIdSlice[i]]; ok { //if the current gene is in the normalization map.
```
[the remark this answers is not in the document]
rimangan: a user can include a normalization file for a set of genes for which a multiplier should be applied. Notably, this could be a subset of all genes, so some genes in the reference to which the sams are aligned may not have a normalizataion entry (for example, with my STARR-seq data I normalize to the input starr-seq library but do not normalize the GFP to anything). The output counts for features will not have a multiplier (identical in behavior to a normalization factor of 1). I've made a note in the usage, but I think this should be the correct behavior.

cmd/scCount/scCount.go
```diff
@@ -0,0 +1,165 @@
...
+	err = out.Close()
+	exception.PanicOnErr(err)
+}
+
+func getGeneName(g *gtf.Gene) string {
+	return g.GeneID
+}
+
+func printRow(out io.Writer, r Row, normalizationMap map[string]float64, geneIdSlice []string, normFlag bool) {
+	if normFlag { //if the user passed in a normalization file, we normalize the count for each gene.
```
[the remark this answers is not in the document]
rimangan: I considered this but it might be a bit messier as we always want to normalize each row right before we print and we have a couple of calls to the print function (within and after the loop). I can name the function something more descriptive like normAndPrintRow, which might help

RavenYL (approved):
I'm assuming that the expected formats of genes.gtf and expNormFile that will be parsed in this program are the same format they are typically generated in!

craiglowe (approved):
Great work!

PR #540, opened 2021-09-13
compareBedDistanceBasedOnName

cshoben:
This is ready for review. 

This is program is written to handle two input beds, where it will return the minimum distance when a bed entry from one file matches the other. It is written where the first bed (bed1) may be a list of coordinates of interest and the second bed (genomeBed)is something like a genome wide list of bed entries. Bed1 is cycled through line by line, and genomeBed is checked for the corresponding bed entry by comparing the Name field (column 4 as standard for a bed document). This program expects every bed entry in bed1 to have at least one corresponding bed entry in the genomeBed (when comparing by Name).  There could be multiple bed entries in genomeBed that match an entry Name in bed1, in which the distance will be calculated for each pair (the bed1 to each genomeBed entry) and the lowest distance value will be returned. The out file is bed1 with the Score field filled with the minimum distance between the two bed entry coordinates.


cmd/compareBedDistanceBasedOnName/compareBedDistanceBasedOnName.go
```diff
@@ -0,0 +1,102 @@
...
+			log.Fatalf("Unable to compare distance, error message: %s", err)
+		}
+		input[i].Score = lowestDistance
+		if input[i].FieldsInitialized < 5 {
+			input[i].FieldsInitialized = 5
+		}
+		bed.WriteToFileHandle(out, input[i])
+	}
+	err = out.Close()
+	if err != nil {
```
rimangan, line 57: Usually we call exception.PanicOnErr(err) here, which automatically checks if the err is not-nil

cmd/compareBedDistanceBasedOnName/compareBedDistanceBasedOnName.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	var inputNameField []bed.Bed
+	var found bool
+	var distance int
+	var lowestDistance int
+	var err error
+	var j int
+
+	for i := 0; i < len(input); i++ {
```
rimangan, line 30: This can be rewritten as follows:
for i := range input

both are functionally equivalent, but the range usage is stylistically preferred.

cmd/compareBedDistanceBasedOnName/compareBedDistanceBasedOnName.go
```diff
@@ -0,0 +1,102 @@
...
+		lowestDistance = 0
+		inputNameField, found = genomeMap[input[i].Name]
+		if found != true {
+			log.Fatalf("Did not find %s", input[i].Name)
+		}
+		for j = range inputNameField {
+			distance, err = bed.CompareDistance(inputNameField[j], input[i])
```
rimangan, line 37: do we need to catch this err in the loop at all? I noticed it is checked once we exit the loop, but not for each iteration

rimangan (commented):
Looks like the binary file still wound up in the pull request despite our efforts

cmd/compareBedDistanceBasedOnName/compareBedDistanceBasedOnName.go
```diff
@@ -0,0 +1,102 @@
...
+	var distance int
+	var lowestDistance int
+	var err error
+	var j int
+
+	for i := 0; i < len(input); i++ {
+		lowestDistance = 0
+		inputNameField, found = genomeMap[input[i].Name]
+		if found != true {
+			log.Fatalf("Did not find %s", input[i].Name)
```
RavenYL, line 34: I was going to comment that the log.Fatal here seems like an intense reaction to not finding the gene name, but then I read that you expect to find at least 1 corresponding gene name entry in bed 2 for each entry in bed 1, so it's probably a suitable application!

RavenYL (approved):

cmd/compareBedDistanceBasedOnName/compareBedDistanceBasedOnName.go
cshoben: does that still fatal or just report a message stating there was a problem? I actually dont want the program to keep going if piped into something downstream as I could have missing data if the file didnt close.

cshoben: updated, thanks!

dasnellings (approved):

rimangan (approved):
LGTM!

craiglowe (approved):
Great work!
Let continue to think about how this could be expanded to be a more general program.  Maybe someone at code review will have an idea.

PR #541, opened 2021-09-13
bedValueWig: bedMinWig and bedMaxWig

rimangan:
The cmd bedMaxWig, which returns the maximum value from a wig file that overlaps a bed region, has been renamed bedValueWig to reflect that the program has been generalized. By default, this program acts the same as bedMaxWig, although it now accommodates an option to calculate the minimum value of a wig that overlaps a bed entry.
The bedMinWig functionality is also compatible with Chelsea's normalization option.
Passes tests.
Edit: new commit also adds a bedAverageWig functionality, which annotates with average wig value within a bed entry.


cmd/bedValueWig/bedValueWig.go
```diff
@@ -12,12 +12,17 @@ import (
 	"log"
 )
 
-func bedMaxWig(infile string, database string, chromsizeFile string, outfile string, norm bool) {
+func bedValueWig(infile string, database string, chromsizeFile string, outfile string, norm bool, bedMinWigFlag bool, bedAverageWigFlag bool) {
```
cshoben, line 15: great idea on adding the average option!

cmd/bedValueWig/bedValueWig.go
```diff
@@ -12,12 +12,17 @@ import (
...
+	if bedMinWigFlag && bedAverageWigFlag {
+		log.Fatalf("Cannot select both min and average in the same operation.")
+	}
+
 	var records []bed.Bed = bed.Read(infile)
 	var wigData []wig.Wig = wig.Read(database)
 	var sizes []chromInfo.ChromInfo = chromInfo.ReadToSlice(chromsizeFile)
-	var outlist []bed.Bed
+	var outList []bed.Bed
```
cshoben, line 23: good catch no the non camel

cshoben (approved):

cmd/bedValueWig/bedValueWig.go
```diff
@@ -27,54 +32,68 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
...
 		sum = sum + w[i+start]
 	}
-	return (sum / float64(length))
+	return sum / float64(length)
+}
+
+func bedRangeMin(w []float64, start int, end int) float64 {
+	var min = w[start]
+	for i := start; i < end; i++ {
+		min = numbers.MinFloat64(min, w[i])
```
dasnellings, line 94: This is probably an esoteric point that no one cares about but me, so I give you license to ignore me, but you might consider avoiding importing numbers when possible. I think min/max functions probably account for >90% of numbers imports. That would normally be fine, but the numbers package is probably one of the largest packages in gonomics so when you use the min/max from numbers, you also get MCMC, monte carlo models, and log arithmetic included in your binary. The current binary for gonomics is just under 600MB. It might be worth considering ways to trim that down.

dasnellings (approved):

cmd/bedValueWig/bedValueWig.go
rimangan: this feels related to our choice to pull the min/max functions out of the lightweight 'common' package. One option is we write a bunch of small min/max helper functions without our programs, the other is we think about a more appropriate place for those functions in gonomics. Numbers is a very broad package and could also be broken

craiglowe (approved):
Great work!  I had a few small comments

cmd/bedValueWig/bedValueWig.go
dasnellings: Breaking up numbers may be a good idea. I think that this will become less of an issue when generics are implemented (december release??) but for now is something worth considering.

cmd/bedValueWig/bedValueWig.go
```diff
@@ -27,54 +32,68 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
...
+					currValue = bedRangeAverage(wigData[chromIndex].Values, records[k].ChromStart, records[k].ChromEnd)
+				} else {
+					currValue = bedRangeMax(wigData[chromIndex].Values, records[k].ChromStart, records[k].ChromEnd)
+				}
 				if norm == true {
-					maxWig = maxWig / wigTotal
+					currValue = currValue / wigTotal
 				}
-				currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%g", float64(maxWig))) // %g will
+				currentBed.Annotation = append(currentBed.Annotation, fmt.Sprintf("%g", currValue)) // %g will
```
RavenYL, line 60: Will it be beneficial to mark in some way whether we calculated max, min or average, either in the name field in each entry, or somewhere in the output file name? It might be annoying to have to remove something in each name field if we don't want it to contain a "max/min/avg" label, but I wonder if indicating "_min/max/avg" in the output file name will be useful

RavenYL (approved):
Thank you for adding these functionalities!

convert/convert.go
```diff
@@ -122,7 +122,7 @@ func BedReadsToWig(b []bed.Bed, reference map[string]chromInfo.ChromInfo) []wig.
 
 	//generate Wig skeleton from reference
 	for _, v := range reference {
-		currentWig = wig.Wig{StepType: "fixedStep", Chrom: v.Name, Start: 1, Step: 1}
+		currentWig = wig.Wig{StepType: "fixedStep", Chrom: v.Name, Start: 1, Step: 1, Span: -1}
```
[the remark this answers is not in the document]
rimangan: -1 leads to no span field in the output file. Do we care if we get a span field? I did this primarily for read/write fidelity, as most wigs made in genomics have no span field

PR #542, opened 2021-09-18
mergeSort Sam unaligned

rimangan:
When sorting sam records, we run into a log fatal when sorting unaligned reads, as the cigar.ConsumeReference function calls a fatal if the cigar rune is '*'. ConsumesReference is called in the GetChromEnd helper function in sam/methods.go. Appears to be working now for genome-scale sam files on dcc. I added a case for unaligned reads to the test for sam sorting.


sam/methods.go
```diff
@@ -17,13 +17,16 @@ func (s *Sam) GetChromStart() int {
 }
 
 func (s *Sam) GetChromEnd() int {
-	var runlength int = 0
+	var runLength int = 0
+	if s.Cigar[0].Op == '*' {
+		return s.GetChromStart()
```
craiglowe, line 22: What is the chrom start for an unaligned read?

craiglowe (approved):
Good catch!

cshoben (approved):

cfauci (approved):

PR #543, opened 2021-09-20
off-by-one bug fix in interval

craiglowe:
I believe I have tracked down an off-by-one bug fix in the interval package which is caused by interval being left-closed, right-open, but the math in our implementation being for left-closed, right-closed.  It will be good to make sure Dan double-checks this change because I think he has the deepest knowledge in this section of the code base.

### DIFF
--- a/interval/interval_test.go
+++ b/interval/interval_test.go
@@ -11,9 +11,9 @@ import (
 
 func TestQuery(t *testing.T) {
 	type testQuery struct {
-		query      *bed.Bed
-		relationship  string
-		hasOverlap bool
+		query        *bed.Bed
+		relationship string
+		hasOverlap   bool
 	}
 
 	var targetIntervals []Interval = []Interval{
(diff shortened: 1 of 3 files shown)


RavenYL (approved):

interval/interval_test.go
```diff
@@ -5,30 +5,105 @@ import (
...
 		&bed.Bed{Chrom: "1", ChromStart: 8, ChromEnd: 10},
 		&bed.Bed{Chrom: "1", ChromStart: 9, ChromEnd: 10},
+		&bed.Bed{Chrom: "chr2", ChromStart: 1, ChromEnd: 3},
+		&bed.Bed{Chrom: "chr2", ChromStart: 4, ChromEnd: 10},
+		&bed.Bed{Chrom: "chr3", ChromStart: 10, ChromEnd: 100},
+	}
+
+	var queries []testQuery = []testQuery{
+		{
+			query:        &bed.Bed{Chrom: "1", ChromStart: 4, ChromEnd: 5},
```
cshoben, line 35: Are we using "1" instead of "chr1" to test some functionality?

rimangan (approved):

dasnellings (approved):

PR #544, opened 2021-09-23
Parse Flag error in sam sorting by Bx

rimangan:
Today I was attempting to sort sam files by barcode and encountered a small bug. In the main version, we check the value of the variable "singleCellBx" before we parse the flag, so this value is always false. I have moved the parse.Flag call above when we check the value of singleCellBx, so this error is fixed. Missed in tests as this is an error in the cmd template, our tests call the mergesort function directly.


craiglowe (approved):
Good catch!

RavenYL (approved):

cshoben (approved):

PR #546, opened 2021-10-14
Find and replace

cshoben:
1) findAndReplace command created
- used to scan a file for a specified string and replace with another string
- use a two column tab delimited file where column 1 is the "find" and column 2 is the "replace"
- can scan the whole file  (default) or a specified column number (zero based)
- currently can only handle strings
- currently depends on all files being tab delimited
- test file and test data included in this PR
2) wrote a fileio.Write() function, that was used in the command above. 
- updates the fileio test file for the Read() and Write() functions.


fileio/easyio.go
```diff
@@ -80,6 +80,9 @@ func EasyOpen(filename string) *EasyReader {
 // EasyCreate creates a file with the input name. Panics if errors are encountered.
 func EasyCreate(filename string) *EasyWriter {
 	answer := EasyWriter{}
+	if filename == "" {
```
dasnellings, line 83: Nice catch. Could you consider moving this to fileio.MustCreate instead of EasyCreate so that other derivatives of MustCreate can benefit as well?

fileio/fileio.go
```diff
@@ -195,3 +195,21 @@ func ReadFileToSingleLineString(filename string) string {
 	}
 	return catInput
 }
+
+func WriteToFileHandle(file io.Writer, rec string) {
```
dasnellings, line 199: I don't think this function is necessary. It would usually be easier and faster to just do `file.Write([]byte(rec))`

fileio/fileio.go
```diff
@@ -195,3 +195,21 @@ func ReadFileToSingleLineString(filename string) string {
...
+	var err error
+	_, err = fmt.Fprintf(file, "%s\n", rec)
+	exception.PanicOnErr(err)
+}
+
+//Write writes a file
+func Write(filename string, records []string) {
```
dasnellings, line 206: I have no objections to the code, but a couple housekeeping points. 

Since this function uses the 'Easy' functions it may make sense for this to live in easyio.go instead of fileio.go. 

The function name `Write` is generally reserved for writing a slice of bytes to an open file. It may be more fitting to call this something like `EasyWriteFile` to avoid confusion.

cmd/findAndReplace/findAndReplace.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/fileio"
+	"log"
+	"strings"
+)
+
+//TODO
+// findAndReplace to handle both int and string? Currently just strings
```
RavenYL, line 12: There might be a relatively quick way to implement this on int? I'm trying to think. I thought if you just let the program interpret ints as strings it would still work? If you're looking for parts of the int, e.g. replacing the "5" in "357" to become "397", then you can further slice up the words []string so you get words [][]string, and the second layer would be individual ints interpreted as strings, and you can use a variation of the current for loop on this second layer to do the same "find and replace" function within each string, if that's the functionality you're hoping to achieve?

cshoben: Good point Raven! I tested it and it does handle "int" just fine. At this point I don't have a reason to change part of a column like your example, but that could be written into a future version.

fileio/easyio.go
cshoben: Sounds good to me. I made the switch.

fileio/fileio.go
cshoben: I don't follow how I could call the function within itself as a helper function. Could you expand?

fileio/fileio.go
cshoben: Happy to move it and rename it. I'm interested in using your suggestion with the helper function above and will wait to make this change until I improve that helper function.

RavenYL (approved):

PR #549, opened 2021-10-27
bedFormat scale nameFloat

rimangan:
Added testing for bedFormat, which previously lacked testing coverage. Implemented a new option: scaleNameFloat. In some cases, we have stored float values in the name field. This option simply allows the user to scale all the values stored in the name column by a constant multiplier.


cmd/bedFormat/bedFormat.go
```diff
@@ -40,11 +49,19 @@ func usage() {
...
+	EnsemblToUCSC  bool
+	ScaleNameFloat float64
+}
+
 func main() {
 	var expectedNumArgs int = 2
 	var ensemblToUCSC *bool = flag.Bool("ensemblToUCSC", false, "Changes chromosome format type.")
 	var UCSCToEnsembl *bool = flag.Bool("UCSCToEnsembl", false, "Changes chromosome format type.")
-	var fields *int = flag.Int("fields", 4, "Specify the number of bed fields.")
+	var scaleNameFloat *float64 = flag.Float64("scaleNameFloat", math.MaxFloat64, "If float values are held in the name field, scale those values by this constant multiplier.")
```
craiglowe, line 64: Should this scale value default to 1?

craiglowe (approved):
Great to have this.  Is 1 a better default option for scaling all numbers?  This seems like it will be more understandable to the user who sees it as the default value in the help message.

cmd/bedFormat/bedFormat.go
rimangan: We need an option to not scaleNameFloat for beds where non-float64 data is stored in the name field

cfauci (approved):

RavenYL (approved):

craiglowe (approved):
Great work!

PR #550, opened 2021-11-01
bedMath

rimangan:
We have been using the name field to store float values in bed format files in a number of programs. bedMath is a new cmd that enables arithmetic operations on these values. For example, let's say I have two bed files with corresponding entries as seen below:
file1:
chr1 10 20 4.22
file 2:
chr1 10 20 8.1
In these entries, the fourth column (name) contains float values.
The following cmd will add those name values together and produce a new bed entry:
bedMath file1.bed plus file2.bed out.bed
with the following output in out.bed:
chr1 10 20 12.32
In this output, the name is now the sum of values in the corresponding bed entries from the two input files.
bedMath supports addition, subtraction, multiplication, and division. Passes tests.


RavenYL (approved):
If I remember correctly, multiple things can get put into the name field when we process a bed in gonomics. Is it up to the user to make sure to only feed in beds where the name field only contains the desired value that can be interpreted as a float? If so, then the usage might need to specify that?

craiglowe (commented):
some comments

cmd/bedMath/bedMath.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	err = out.Close()
+	exception.PanicOnErr(err)
+}
+
+func doBedMath(a bed.Bed, b bed.Bed, Op Operation) bed.Bed {
+	var aFloat = common.StringToFloat64(a.Name)
+	var bFloat = common.StringToFloat64(b.Name)
+	if Op == Add {
+		a.Name = fmt.Sprintf("%e", aFloat+bFloat)
```
craiglowe, line 59: %g?

craiglowe (approved):
Great work!

dasnellings (approved):

cshoben (approved):

RavenYL (approved):

PR #551, opened 2021-11-02
Raven align low mem4

RavenYL:
This is a low memory version of the affineGap alignment algorithm. It is very similar to the low memory version of constGap I wrote earlier, except affineGap's data structures have another dimension that stores "alternate universe" scores in the situations where the ColType is 0 (match), 1 (insertion), or 2 (deletion), and then the actual traceback jumps between those 3 duplicate matrices depending on whether the ColType is actually 0, 1, or 2.


align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
+//				coordinates specifying the current checkerboard in rows (k1) and columns (k2), row (i) and column (j) positions of inChecker_min_Previous which describe where Step 3 (writeCigar) stopped in the previous checkerboard that just went through Step 3 (writeCigar)
+//				trace matrix which was initialized in the main ConstGap function and passed back and forth to be recycled
+//outputs: trace matrix to be recycled, row (i), column (j) and dimension (k) positions of inChecker_max which describe where Step 2 (fillTraceback) stopped in the current checkerboard that just went through Step 2 (fillTraceback)
+func fillTraceback_affineGap(alpha []dna.Base, beta []dna.Base, scores [][]int64, gapOpen int64, gapExtend int64, checkersize_i int, checkersize_j int, score_highest_i int, score_highest_j int, trace_prep_i [][][]int64, trace_prep_j [][][]int64, k1 int, k2 int, i_inChecker_min_Previous int, j_inChecker_min_Previous int, trace [][][]ColType) ([][][]ColType, int, int, ColType) {
```
rimangan, line 237: many arguments for this function might be tough for further development. perhaps we could consider an internal struct to hold all these fields that get passed around?

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
-	if alphaSize%chunkSize != 0 {
-		log.Fatalf(fmt.Sprintf("Error: the first sequence, %s, has a length of %d, when it should be a multiple of %d\n", dna.BasesToString(alpha), alphaSize, chunkSize))
-	}
-	if betaSize%chunkSize != 0 {
-		log.Fatalf(fmt.Sprintf("Error: the second sequence, %s, has a length of %d, when it should be a multiple of %d\n", dna.BasesToString(beta), betaSize, chunkSize))
-	}
-	alphaChunks := alphaSize / chunkSize
-	betaChunks := betaSize / chunkSize
+//This version of AffineGap needs additional inputs and allows customization of checkersize_i and checkersize_j
+func AffineGap_customizeCheckersize(alpha []dna.Base, beta []dna.Base, scores [][]int64, gapOpen int64, gapExtend int64, checkersize_i int, checkersize_j int) (int64, []Cigar) { //input=same as AffineGap_step1 for now, output=route
```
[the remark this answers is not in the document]
RavenYL: Keep this but reduce code so it just calls AffineGap_customizeCheckersize

rimangan (approved):

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
+func fillTraceback_affineGap(alpha []dna.Base, beta []dna.Base, scores [][]int64, gapOpen int64, gapExtend int64, checkersize_i int, checkersize_j int, score_highest_i int, score_highest_j int, trace_prep_i [][][]int64, trace_prep_j [][][]int64, k1 int, k2 int, i_inChecker_min_Previous int, j_inChecker_min_Previous int, trace [][][]ColType) ([][][]ColType, int, int, ColType) {
+	mRowCurrent := make([][]int64, 3)
+	mRowPrevious := make([][]int64, 3)
+	for k := range mRowCurrent { //k ranges through 3 numbers (0,1,2)
+		mRowCurrent[k] = make([]int64, len(beta)+1)
+		mRowPrevious[k] = make([]int64, len(beta)+1)
+		copy(mRowPrevious[k], trace_prep_i[k][k1])
+	}
+
+	//i_inChecker and j_inChecker are internal (within checkerboard) versions of i and j, e.g. for checkersize=3, i_inChecker can take on the values 0, 1, 2
```
cshoben, line 246: Honestly, I'm struggling to make sense of this note. Could you explain it more for me please?

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
-	if len(alpha.Seq) != len(beta.Seq) {
-		return 0, fmt.Errorf("Error: alignment being scored has sequences of unequal length: %d, %d\n", len(alpha.Seq), len(beta.Seq))
+//Step 3
+//inputs: trace matrix, row (i), column (j) and dimension (k) positions of inChecker_max_Previous which describe where Step 2 (fillTraceback) stopped in the previous checkerboard that just went through Step 2 (fillTraceback)
+//				the growing route describing the collection of cigars of the entire alignment, the index of the cigar that is currently being built, row (i), column (j) and dimension (k) positions of inChecker_min_Previous which describe where Step 3 (writeCigar) stopped in the previous checkerboard that just went through Step 3 (writeCigar)
+//outputs: the updated grown route describine the collection of cigars of the entire alignment, the updated index of the cigar that is currently being built, row (i), column (j) and dimension (k) positions of inChecker_min which describe where Step 3 (writeCigar) stopped in the current checkerboard that just went through Step 3 (writeCigar)
```
cshoben, line 292: minor typo in comments ;"describe"

cshoben (approved):
Various comments but nothing to hold back a merge :)

align/affineGap_highMem.go
```diff
@@ -0,0 +1,259 @@
+package align
```
craiglowe, line 1: Is this a copy of the affineGap.go file from before?  If so, in the future you can do "git mv affineGap.go affineGap_highMem.go" which will do this and it will show up as a simple one-line rename in the diff.  Currently it is difficult to tell what has just been moved to a new file and what has been modified.

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
 
 import (
-	"fmt"
 	"github.com/vertgenlab/gonomics/dna"
-	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/numbers"
 	"log"
 )
 
+//reminder that unlike constGap, affineGap is an alignment algorithm that favors having a long gap compared to multiple gaps. In the code, affineGap data structures have an additional "k" dimension, which can take on values 0,1,2 and represent the "alternative universes" of scores where the correct alignment produces Match(0), Insertion(1) and Deletion(2)
```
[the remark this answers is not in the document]
craiglowe: Constant gap means that each gap character "-" in the alignment has the same penalty.  For example, if "-" is -5, then "---" is -15.  In affine gap, there is a "opening" cost for the first gap, and then gaps following the opening have small extension penalty.  This means the aligner will prefer "A--AA" over "A-A-A", where constant gap penalty sees them as equal.

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
-				mRowCurrent[1][j], trace[1][i][j] = tripleMaxTrace(gapOpen+gapExtend+mRowCurrent[0][j-1], gapExtend+mRowCurrent[1][j-1], gapOpen+gapExtend+mRowCurrent[2][j-1])
-				mRowCurrent[2][j], trace[2][i][j] = tripleMaxTrace(gapOpen+gapExtend+mRowPrevious[0][j], gapOpen+gapExtend+mRowPrevious[1][j], gapExtend+mRowPrevious[2][j])
-			}
-		}
-		if i < mColumn-1 {
-			mRowPrevious, mRowCurrent = mRowCurrent, mRowPrevious
+	var checkersize_i, checkersize_j int
+	checkersize_i = 10000
+	checkersize_j = 10000
+
```
craiglowe, line 57: I think it is good to have this function, but I would replace it with a call to:
AffineGap_customizeCheckersize
after setting the checkersize to be 10000.  This way you don't have code duplication.
For my comments, I am going to skip ahead to the custom checkersize function.

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
+			copy(trace_prep_i[2][i/checkersize_i], mRowCurrent[2])
 			mRowPrevious, mRowCurrent = mRowCurrent, mRowPrevious
+		} else if i < mColumn-1 {
+			mRowPrevious, mRowCurrent = mRowCurrent, mRowPrevious //reuse mRow variables to save memory, but only up until the second to last row
 		}
 	}
-	maxScore, route := affineTrace(mRowCurrent, mColumn, trace)
-
-	return maxScore, route
+	score_highest_i := mColumn - 1
```
craiglowe, line 226: Right now it looks like we use the final corner spot in the matrix as the "highest score."  This would be a "global alignment" that forces all of sequence one to align to all of sequence two.  In the future, we can add an option for a "loca alignment" where would would actually keep track of the highest score in the matrix and send back that score and its coordinates in the matrix (not needed for this pull request).

craiglowe (commented):
Wonderful work!  I think there is at least one place where you have some code duplication you can remove and then I think we are all set to get this merged!

dasnellings (approved):
Great job!

align/affineGap_highMem.go
RavenYL: Yes it is a copy! I did git mv initially but then ran into git merge conflicts after go fmt, and when I resolved those conflicts the affineGap_highMem.go document got uploaded as a new file entirely. I will do that in the future

align/affineGap.go
RavenYL: Just fixed this

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
+	checkersize_i = 10000
+	checkersize_j = 10000
+
+	//Step 1: find highest score, as well as get the position (i and j) of the highest score, and materials needed to fill traceback and write cigar in checkerboards
+	score_highest, score_highest_i, score_highest_j, trace_prep_i, trace_prep_j := highestScore_affineGap(alpha, beta, scores, gapOpen, gapExtend, checkersize_i, checkersize_j)
+
+	//Make variables needed for checkerboards
+	//k1: the i-index (row-index) of the current checkerboard
+	//k2: the j-index (column-index) of the current checkerboard
+	//i_inChecker_max: the max i-index (row-index) in the current checkerboard during Step 2 (fillTraceback), either checkersize-1 or a smaller number, e.g. if the alpha/beta sequence to be aligned is not divisible into perfect checkerboards, and the remainder checkerboard is not completely filled
```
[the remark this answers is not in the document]
RavenYL: Just fixed this based on my fullscreen window!

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
+	//k_inChecker_max: the k-index (out of 0,1,2) of the max i and j in the current checkerboard during Step 2 (fillTraceback)
+	//i_inChecker_min: the min i-index (row-index) in the current checkerboard during Step 3 (writeCigar), either 0 or another number, e.g. if cigar route leaves the checkerboard at a position that is not i_inChecker==0, j_inChecker==0
+	//j_inChecker_min: the min j-index (column-index) in the current checkerboard during Step 3 (writeCigar)
+	//k_inChecker_min: the k-index (out of 0,1,2) of the min i and j in the current checkerboard during Step 3 (writeCigar)
+	//note that k1 amd k2 (coordinates for checkerboads) are not related to k_inChecker_max and k_inChecker_min (out of 0,1,2). Also note that k_inCkecer_max and k_inChecker_min do not determine traceback path (e.g. determine the next k1 and k2 to go to), but influence cigar route
+	var k1, k2, i_inChecker_max, j_inChecker_max, i_inChecker_min, j_inChecker_min int
+	i_inChecker_min = -2                                                          //initialize i_inChecker_min != 0, so that the first ever Step 3 (writeCigar) will not interfere with i_inChecker_max
```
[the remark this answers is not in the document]
RavenYL: I think I could have used -1 but I wanted to be sure I won't accidentally generate 0 somehow

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
-			mRowPrevious, mRowCurrent = mRowCurrent, mRowPrevious
+	var checkersize_i, checkersize_j int
+	checkersize_i = 10000
+	checkersize_j = 10000
+
+	//Step 1: find highest score, as well as get the position (i and j) of the highest score, and materials needed to fill traceback and write cigar in checkerboards
+	score_highest, score_highest_i, score_highest_j, trace_prep_i, trace_prep_j := highestScore_affineGap(alpha, beta, scores, gapOpen, gapExtend, checkersize_i, checkersize_j)
+
+	//Make variables needed for checkerboards
+	//k1: the i-index (row-index) of the current checkerboard
```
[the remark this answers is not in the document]
RavenYL: You're right! I made the variables k1 and k2 back when I was working on constGap and "k" wasn't a variable. Now that there's a "k" variable in affineGap it's a little confusing, but the k1 and k2 have nothing to do with k. They're numbers that keep track of which checkerboard we're in in each 2D matrix

align/affineGap.go
```diff
@@ -1,259 +1,352 @@
...
+			k2 -= 1 //go to the next checkerboard with lower j, same i
 		}
+
 	}
-	maxScore, route := affineTrace(mRowCurrent, mColumn, trace)
 
-	return maxScore, route
+	//Step 4: write the last cigar entry. Since this step doesn't need special features for affineGap, reuse constGap function
+	//This step is necessary because the row i=0 and the column j=0 are always stored in trace_prep, and never filled in in any checkerboard, but the cigar ends by the route going into 1 box/entry in either the row i=0 or the column j=0
+	if i_inChecker_min != -1 && j_inChecker_min == -1 { //indicating that Step 3 ended when k1 and k2 reached the smallest combination, and reached j=0, so the last cigar is a "2", e.g. if last cigar entry is the i=1, j=0 square
```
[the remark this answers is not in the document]
RavenYL: Since the last square that was filled in the checkerboard reached j=0 (ran out of columns before running out of rows), and the beta sequence is on the x-axis (each column corresponds to a base in the beta sequence), reaching j=0 first means the last cigar should be a "deletion" in the beta sequence

align/affineGap.go
RavenYL: In a 3x3 checkerboard structure, the i and j could be 3,4,5 ("external", true i and j), but when we divide up all the i and j into checkerboards, those same coordinates would have "internal" i and j of 0,1,2. So the external i and j will keep going up (0,1,2,3,4,5,6,...) but the internal i and j will loop within the size of the checkerboard (0,1,2,0,1,2,...)

align/affineGap.go
RavenYL: Just fixed this!

craiglowe (approved):
Great work!!

dasnellings (approved):

rimangan (approved):
Fantastic!

PR #552, opened 2021-11-10
VcfFilter subset

rimangan:
Implements a -subSet option in vcfFilter, in which the user can provide a float between 0 and 1 to retain a proportion of variants in the output file. Pseudorandom, so it will not guarantee a particular number of variants in the output. 
This serves as an alternative to sampleVcf, which already has a similar functionality for producing subsets of Vcf files. However, sampleVcf reads the whole Vcf into memory, shuffles the slice of Vcf structs, and returns a precise number of variants. While this is useful, say if you wanted precisely 1000 variants, this program is not memory efficient. To get precisely 1000 variants from a large vcf file, I would recommend first subsetting with vcfFilter to perhaps 1M variants and then further subsetting with sampleVcf to get the final product.
Passes tests.
I also reorganized a bit of code related to sampleVcf for clarity (mainly moving some code into a helper function)

### DIFF
--- a/cmd/vcfFilter/vcfFilter.go
+++ b/cmd/vcfFilter/vcfFilter.go
@@ -5,6 +5,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"github.com/vertgenlab/gonomics/common"
 	"github.com/vertgenlab/gonomics/fileio"
 	"github.com/vertgenlab/gonomics/popgen"
 	"github.com/vertgenlab/gonomics/vcf"
@@ -14,7 +15,8 @@ import (
 	"strings"
 )
 
-func vcfFilter(infile string, outfile string, c criteria, groupFile string, parseFormat bool, parseInfo bool) (total, removed int) {
+func vcfFilter(infile string, outfile string, c criteria, groupFile string, parseFormat bool, parseInfo bool, randSeed bool, setSeed int64) (total, removed int) {
+	common.RngSeed(randSeed, setSeed)
 	records, header := vcf.GoReadToChan(infile)
 	out := fileio.EasyCreate(outfile)
 	tests := getTests(c, header)
@@ -111,7 +113,7 @@ type criteria struct {
 	refStrongAltWeakOnly           bool
 	notRefWeakAltStrong            bool
 	notRefStrongAltWeak            bool
-	id                             string //raven's note: added id (rsID), can upgrade to []string in the future
+	id                             string
 	formatExp                      string
 	infoExp                        string
 	includeMissingInfo             bool
@@ -245,7 +247,7 @@ func getTests(c criteria, header vcf.Header) testingFuncs {
 	if c.notRefStrongAltWeak {
 		answer = append(answer, vcf.IsNotRefStrongAltWeak)
 	}
-	if c.id != "" { //raven's note: not sure what getTests is for and if c.id (modeled after c.chrom) is right
+	if c.id != "" {
 		answer = append(answer,
 			func(v vcf.Vcf) bool {
 				if v.Id != c.id {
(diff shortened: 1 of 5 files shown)


cmd/vcfFilter/vcfFilter_test.go
```diff
@@ -66,16 +72,18 @@ func TestVcfFilter(t *testing.T) {
...
 		}
 
-		vcfFilter(v.inputFile, "tmp.vcf", c, v.groupFile, false, false)
-		records, _ := vcf.Read("tmp.vcf")
+		vcfFilter(v.inputFile, v.tmpOutFile, c, v.groupFile, false, false, v.randSeed, v.setSeed)
+		records, _ := vcf.Read(v.tmpOutFile)
 		expected, _ := vcf.Read(v.expectedOutputFile)
 		if !vcf.AllEqual(records, expected) {
 			t.Errorf("Error in vcfFilter.")
+		} else {
```
craiglowe, line 84: That is a nice bit of logic so that we keep the temp file around when it does not pass the test.

craiglowe (approved):
Nice work!

cmd/vcfFilter/vcfFilter_test.go
rimangan: yes we should make this standard practice. Maybe worth a note in code review

cmd/vcfFilter/vcfFilter.go
```diff
@@ -252,11 +256,23 @@ func getTests(c criteria, header vcf.Header) testingFuncs {
...
+				return true
+			})
+	}
 	return answer
 }
 
 func main() {
 	var expectedNumArgs int = 2
+	var randSeed *bool = flag.Bool("randSeed", false, "Uses a random seed for the RNG.")
+	var setSeed *int64 = flag.Int64("setSeed", -1, "Use a specific seed for the RNG.")
```
dasnellings, line 275: 2 flags for seeds seems like overkill. How about if setSeed is -1 then it uses a random seed, else it uses setSeed.

rimangan: randSeed by default is pretty dangerous. If we want to move to one flag, I think we should abandon the randSeed all together, as it produces data that can never be replicated. I never use it and would suggest people don't

craiglowe: You can default setSeed to 19 (or some other number) so that it always runs with a constant seed by default and is repeatable.  Then someone can use the setSeed option to run with another seed (other than 19) if they would like to get a different sample.  I think the question here is: do we make -1 a special value of set seed that uses a random seed, or do we have a separate option to do a random seed (or do we not allow random seeds).  I am not super opinionated on this.  I think Riley makes a good point that it is probably better programming/science practice to generate random numbers, and then use these random seeds/numbers as the starting point for the analysis with setSeed, but this may be beyond some users and "randSeed" would be a quick option for them when generating their files.

dasnellings: Ah, yeah I can see how that would cause problems. Is there much value in providing a random seed option? If not, I would agree with Riley and just remove it. I don't see much reason to enable people to use a bad practice. If they really want a random seed in each run then they can wrap it in a bash script.

cmd/vcfFilter/vcfFilter.go
```diff
@@ -282,6 +298,7 @@ func main() {
 	var infoExp *string = flag.String("info", "", "Identical to the 'format' tag, but tests the info field. The values of type 'Flag' in the info field"+
 		"can be tested by including just the flag ID in the expression. E.g. To select all records with the flag 'GG' you would use the expression \"GG\".")
 	var includeMissingInfo *bool = flag.Bool("includeMissingInfo", false, "When querying the records using the \"-info\" tag, include records where the queried tags are not present.")
+	var subSet *float64 = flag.Float64("subSet", 1, "Proportion of variants to retain in output. Value must be between 0 and 1.")
```
RavenYL, line 301: Would it be helpful to add to the message here, saying that in order to get a small set of variants from a large vcf file, the recommended action is to first subset with vcfFilter, then feed the result into sampleVcf?

RavenYL (approved):

craiglowe:
I am trying to survey the status of our pull requests.  I think this one is waiting on an update to remove the option to run with a random seed.  Is that correct?  If that is true, I think we could merge without that, since it is not a bug, and a later pull request could remove the random seed option from this and possibly other programs.

dasnellings (approved):

PR #553, opened 2021-11-13
Annotate VCF with population frequency via web query

dasnellings:
I wanted to annotate some VCFs with population allele frequencies from gnomAD and 1kgp, but did not want to download the massive VCFs to do it. This cmd queries CellBase which has data from gnomAD and 1kgp (any many other databases)  and parses the response to annotate the input VCF file.


cmd/vcfWebAnnotate/vcfWebAnnotate.go
```diff
@@ -0,0 +1,195 @@
...
+			continue
+		}
+		consequence = res.Responses[i].Results[0].Consequences[0]
+
+		if res.Responses[i].Results[0].ConsequenceType != "" {
+			ann = append(ann, fmt.Sprintf("Consequence=%s", res.Responses[i].Results[0].ConsequenceType))
+		}
+
+		if consequence.GeneName != "" {
+			ann = append(ann, fmt.Sprintf("Gene=%s", consequence.GeneName))
```
craiglowe, line 111: What happens if we overlap more than one gene/transcript/protein?

dasnellings (commented):
I just pushed my changes, however it seems the CellBase server is currently down which brings up a new question. What are peoples thoughts on web dependencies? I deliberately made all of this in a cmd because I figured it was probably not a good idea to introduced a soft dependency in the main codebase that others may build off of. 

Currently this cmd fails tests as the server is down. Does anyone have thoughts on how web dependencies should be handled in gonomics?

cmd/vcfWebAnnotate/vcfWebAnnotate.go
craiglowe: Hmmm.... I think we should at least put a note in the usage statement that we do this.  I can image people wanting all vcfs that overlap a gene or something like that, and currently they will not get it (in corner cases).  Can we add an option for "allData" or something similar that will generate the huge field?  Does it still get unwieldy if you give all genes it overlaps and all protein annotations?  Then only the transcriptIds would be hidden by default?

craiglowe: That sounds like a good solution for this pull request.  This might be a good discussion topic for our next code review so that you can fully explain the situation and we can see if anyone has a good idea.

RavenYL (dismissed):
In the future, if we make more cmds that have web dependencies maybe they can be in their own directory within cmd so we're aware these things may need to be updated from time to time?

craiglowe (approved):
Great work!

RavenYL (approved):

rimangan (approved):

PR #554, opened 2021-11-16
Bam Reading Bug

dasnellings:
There was a bug when reading bam files with headers >50KB where only 1 more bgzf block was read. This change puts block reading in a for loop so that as many block as needed are read.


craiglowe (approved):
Good find!

rimangan (approved):

RavenYL (approved):

PR #555, opened 2021-11-18
added minScore to axTools

sweavs111:
i needed to sort by minScore in axTools. added.


dasnellings (approved):
Looks great!

edotau (approved):
lgtm

cshoben (commented):
Looks good overall! Happy to approve, but wanted to double check... your comment says you wanted to sort by minScore in axtTools but this new function seems to just filter on a minimum score. I assume you could just use the bash sort with the output but wanted to double checked that this function is doing what you intended?

You could also add this function to the axTools_test.go file in a future PR?

PR #556, opened 2021-11-19
triallelic sites in VCF files

dasnellings:
Our implementation of VCF can not handle triallelic sites. I don't necessarily suggest merging this PR as it would break read/write fidelity, but it may be worth revisiting the VCF struct at some point...


rimangan:
this has come up before, specifically when we did the last major overhaul of VCF. The AlleleOne and AlleleTwo fields will need to become a slice of Alleles, which will have some downstream effects but should be rather manageable. Lots of popgen theory assumes biallelic sites, so we could just ignore triallelic positions in those programs

PR #557, opened 2021-11-29
remove fa Cmds

rimangan:
Removing unnecessary cmds whose functionality is covered by faFormat


RavenYL (approved):

dasnellings (dismissed):

craiglowe:
I think faGaps should stay because it can create a "noGap" file

rimangan:
> I think faGaps should stay because it can create a "noGap" file

Great catch! I'll do a request to add the "noGap" functionality to faFormat, then we can remove faGaps

RavenYL (approved):

craiglowe (approved):
Looks good!
It looks like faFormat is expanding a little bit beyond just reformatting a faFile.  Maybe it could be called something like faUtils in the future?

dasnellings (approved):

PR #559, opened 2021-11-29
Draft VCF change to handle ploidy != 2

dasnellings:
This PR is now ready to review. 

This changes VCF to permit any ploidy and removes a bit of unused code in VCF.


vcf/vcf.go
```diff
@@ -26,10 +26,10 @@ type Vcf struct {
...
 // GenomeSample is a substruct of Vcf, and contains information about each sample represented in a VCF line.
-// AlleleOne and AlleleTwo are set to -1 if no genotype data is present. AlleleTwo is set to -1 for hemizygous sites.
+// Indexes in Alleles are set to -1 if no genotype data is present.
 type GenomeSample struct {
-	AlleleOne  int16    // First allele in genotype, 0 for reference, 1 for Alt[0], 2 for Alt[1], etc.
-	AlleleTwo  int16    // Second allele in genotype, same Number format as above.
-	Phased     bool     // True for phased genotype, false for unphased.
-	FormatData []string // FormatData contains additional sample fields after the genotype, which are parsed into a slice delimited by colons. Currently contains a dummy empty string in FormatData[0] corresponding to "GT" in Format, so indices in FormatData will match the indices in Format.
+	Alleles    []int16  // Alleles present in genotype, 0 for reference, 1 for Alt[0], 2 for Alt[1], etc.
+	Phase      []int16  // True for phased genotype, false for unphased. len(Phase) == len(Alleles). Phase[0] == true if and only if Phase[1:] == true
```
craiglowe, line 32: Should this be a []bool instead of a []int16?

craiglowe (commented):
Looks good.  I added one comment

vcf/vcf.go
dasnellings: Yes, fixed.

vcf/vcf_test.go
```diff
@@ -34,6 +35,15 @@ func TestWriteAndRead(t *testing.T) {
 		Write(tempFile, actual)
 		alpha, _ := Read(tempFile)
 		beta, _ := Read(test.filename)
+
+		for i := range alpha {
```
rimangan, line 39: empty block of code? perhaps we can delete?

vcf/io.go
```diff
@@ -54,97 +54,117 @@ func GoReadToChan(filename string) (<-chan Vcf, Header) {
...
+func parseSamples(samples []string, format []string, line string) []Sample {
 	//DEBUG: fmt.Printf("Format: %s. Format[0]: %s.\n", format, format[0])
-	if len(format) == 0 {
-		log.Fatalf("Parsing error: cannot parse notes without formatting information.")
-	}
-	//firstFormat := format[0]
-	//fmt.Println(firstFormat)
-	if format[0] != "GT" && format[0] != "." { //len(format) == 0 checks for when there is info in the notes column but no format specification
-		log.Fatalf("VCF format files with sample information must begin with \"GT\" as the first format column or be marked blank with a period. Here was the first format entry: %s.\nError parsing the line with this Notes column: %s.\n", format[0], data)
+	if format[0] == "." || len(format) == 0 {
```
rimangan, line 93: VCF specification is that format must begin with "GT". We've lost that check here. Looks like we allow io on invalid VCFs?

vcf/io.go
```diff
@@ -54,97 +54,117 @@ func GoReadToChan(filename string) (<-chan Vcf, Header) {
...
+		return nil
 	}
 
-	if format[0] == "." { //if the format column is blank, we do not need to parse further.
-		var blankAnswer []GenomeSample = make([]GenomeSample, 0)
-		return blankAnswer
+	answer := make([]Sample, len(samples))
+	for i := range samples {
+		answer[i].FormatData = strings.Split(samples[i], ":")
+		if format[0] == "GT" {
```
rimangan, line 100: so we mandate GT is in 0, but we can still read a VCF that has something other than GT in format[0]? We might want to throw an error in that case

rimangan (approved):

rimangan (approved):

vcf/compare.go
```diff
@@ -118,39 +67,76 @@ func CompareAlt(alpha []string, beta []string) int {
...
+			if alpha[i].FormatData[j] != beta[i].FormatData[j] {
+				return false
+			}
+		}
+	}
+	return true
+}
+
 //AllEqual returns true if each Vcf in a slice of Vcf structs contain identical information.
 func AllEqual(alpha []Vcf, beta []Vcf) bool {
```
RavenYL, line 141: When comparing 2 unphased vcfs, I imagine it might be possible for vcf1 and vcf2 to share allele[0] and allele[1], but they are just swapped in position so allele[0] in vcf1 is allele[[1] in vcf2 and vice vera. Does this happen in vcfs? If so, then does the current gonomics code consider this?

RavenYL (approved):

cmd/alleleSplit/toSam.go
```diff
@@ -9,12 +9,20 @@ import (
 	"strings"
 )
 
+func ASFilter(v vcf.Vcf, parentOne int16, parentTwo int16, F1 int16) bool {
```
cshoben, line 12: this is the same ASFilter() in cmd/filterGenotypes/filterGenotypes.go ?
If we are making identical functions is that when it's time to migrate it to a package? If not, when do we normally do that?

cshoben (commented):
I still have more to review. Sending through my comments so far.

vcf/filter.go
```diff
@@ -201,11 +58,23 @@ func IsSegregating(v Vcf) bool {
...
 	}
-	var firstEncountered int16 = v.Samples[0].AlleleOne
+
+	var firstEncounteredBase int16
+	var firstEncountered bool = false
 
 	for _, sample := range v.Samples {
-		if sample.AlleleOne != firstEncountered || sample.AlleleTwo != firstEncountered {
-			return true
+		if len(sample.Alleles) == 0 {
```
craiglowe, line 66: Can this happen?

vcf/methods.go
```diff
@@ -8,11 +8,16 @@ import (
...
 )
 
-// String implements the fmt.Stringer interface for easy printing with the fmt package.
+// String implements the fmt.Stringer interface for easy printing of Vcf with the fmt package.
 func (v Vcf) String() string {
 	return fmt.Sprintf("%s\t%v\t%s\t%s\t%s\t%v\t%s\t%s\t%s\t%s\n", v.Chr, v.Pos, v.Id, v.Ref, strings.Join(v.Alt, ","), v.Qual, v.Filter, v.Info, v.Format, SamplesToString(v.Samples))
 }
 
+// String implements the fmt.Stringer interface for easy printing of Sample with the fmt package.
+func (s Sample) String() string {
```
craiglowe, line 17: This string methods are really nice/useful

craiglowe (commented):
Great work!  I am ready to approve.  Looks like there was just one small question about reading in a VCF and checking for the GT field

vcf/filter.go
dasnellings: Yes, if the genotype field is not present, but other format fields are.

vcf/compare.go
dasnellings: True. I think this is mostly (entirely?) used for testing at this point so I can inclined to leave it for now.

cmd/alleleSplit/toSam.go
dasnellings: Yeah, this seemed like a very specific function to me so it felt like it would be best if it was implemented where it was used.

RavenYL (approved):

vcf/gVcf_test.go
```diff
@@ -1,26 +0,0 @@
-package vcf
-
```
cshoben, line 2: Why is this test file removed? From my end it looks like the gVcf.go file will still be around.

cshoben (commented):
I have a quick question about a test file being removed.

craiglowe (approved):
Great to have these improvements!

rimangan (approved):

vcf/gVcf_test.go
dasnellings: GVCF is being largely removed in this PR. What's left is a few functions so that we don't break older programs that still used GVCF instead of VCF, but ultimately GVCF should be removed entirely IMO.

PR #560, opened 2021-11-30
Update to gonomics cmd

dasnellings:
Now that go code can live in directories outside GOPATH the gonomics may now have trouble finding the source code. This PR adds a -setpath flag to manually set the path to the source code. It also adds additional logic to add the -setpath value to the top of the cache file so that -setpath does not need to be used on every invocation of the gonomics cmd.


craiglowe (approved):
Nice to have these improvements!

RavenYL (approved):

cshoben (approved):

PR #561, opened 2021-11-30
simulateVcf AncestorAnnotation and vcfAncestorAnnotation test coverage

rimangan:
1. Adds testing coverage to vcfAncestorAnnotation cmd.
2. simulateVcf, which creates a VCF with allele frequencies based on an input selection parameter, previously left the ancestor allele unlabeled, despite the fact that the output allele spectra are polarized. Output variants are now labeled with ancestral allele states for compatibility with other programs (mostly divergence-based ascertainment corrections require the ancestral state to be known).
3. A new cmd, simulateDivergentWindowsVcf, has been made to validate the behavior of the divergence-based ascertainment correction used in selection analysis. This program essentially recreates the experiment described in Figure 1 of the following paper: https://journals.plos.org/plosone/article/authors?id=10.1371/journal.pone.0005152. Briefly, a number of genomic regions (windows) are created with a number of segregating sites in each region (simulating Vcfs from a stationarity distribution with a fixed selection parameter alpha). In each window, we count the number of divergent sites (where Ref != the ancestral allele). Then, we sort windows by the number of divergent sites and return the top 1% or bottom 1% of windows by divergence. Even though all windows were made with the same selection parameters, the divergence-ascertainment will create a bias in the site frequency spectrum, which we can then correct for.


craiglowe (approved):
Great to have more tests!

dasnellings (dismissed):

RavenYL (approved):

dasnellings (approved):

cshoben (approved):

PR #562, opened 2021-12-05
Selection mle by condition

rimangan:
Just for testing, validating the divergence-based ascertainment bias correction using selectionMLE where our bias correction applies only to bases in the ascertainment condition, instead of evenly across all bases.


rimangan:
Closing this branch. Empirical testing shows that this approach doesn't work
Screen Shot 2021-12-05 at 12 39 41 PM
.

PR #563, opened 2021-12-07
Vcf info update and testing

rimangan:
vcfInfo is a cmd that returns information about the variants to the user, including the number of each type of transition or transversion, information about INDELs, and variants involving N positions.

This request adds an option to return the number of divergent or non-divergent sites relative to the ancestral allele annotation and adds testing to this program.
EDIT: Lots of files changed, as this PR is dependent on my last one.


cmd/vcfInfo/testdata/test.vcf
```diff
@@ -0,0 +1,30 @@
+chr13	113593242	13:113593242:G:A	G	A	255	PASS	AA=G	GT	0|0	0|0	0|0	0|0	0|0	0|0
```
dasnellings, line 1: Is everything in the ID field just for debug?

dasnellings (approved):

cmd/vcfInfo/testdata/test.vcf
rimangan: Arbitrary ID data, just copied from another test file elsewhere in the code base.

RavenYL (approved):

cshoben (approved):

PR #564, opened 2021-12-09
Maf filter test

rimangan:
1. Removed redundant cmd mafFilterScore, which was identical to mafFilter.
2. Added testing coverage for mafFilter.


dasnellings (approved):

RavenYL (approved):

cshoben (approved):
I have to say, I'm a huge fan of these mini PRs. So consumable. 
Also good catch on the redundancy and excited to see some a testing addition.

PR #566, opened 2021-12-10
bedMinimumDistanceName -- calculate minimum distance between bed coordinates when Name field matches

cshoben:
Updates notes: 12/13/21

This PR is now ready for review. 

bedMinimumDistanceName.go take in two bed files. One is the inputBed and one is the genomeBed. Each entry of the inputBed is matched to an entry in the genomeBed based on the Name field. The minimum distance between the two beds is reported, with any overlap reported as zero. This is in the outputBed Score field. 
New functionality: The inputBed entry is compared to the genomeBed entry to determine if it sits upstream or downstream of the genomeBed coordinates. If any bases of the inputBed sit upstream, the entry is considered upstream. This is of course dependent on the genomeBed Strand (which defines that entries directionality). Upstream is '-' and downstream is '+'. 
I have included my drawing of my logic.  

------------------------------------
This is a draft. 
Looking for advice on how to get my "func determineUpstreamDownstream()" in bedMinimumDistanceName.go to work. 
Currently, my code is mishandling negative sense genes.        Lines 84-86 were only added for troubleshooting. This option will later be removed. 
My expected output beds have been included as well as the outputBeds produced by the test files. 

func determineUpstreamDownstream() is a helper function to determine if the coordinates from the inputBed are upstream or downstream of the genomeBed. 
I am using the "Strand" field for the inputBed to capture if it was upstream or downstream. Considering the TSS as 0, being upstream is '-' and being downstream is '+'. Logic is included on lines 70-83
Image from iOS
.


cmd/bedMinimumDistanceName/testdata/genomeBed3.bed
```diff
@@ -0,0 +1,6 @@
+chr1	30	40	gene1	0	+
+chr2	1	13	gene2	0	+
+chr2	2	4	gene3	0	-
+chr1	20	30	gene1	0	+
+chr1	30	56	gene4	0	-
+chr1	55	59	gene4	0	-
```
dasnellings, line 6: From my interpretation, it seems the workflow is to figure out the closest TSS then detect where your input bed is relative to the position of the TSS. For test 3 it seems like the error here is the result of bed.MinimumDistance returning 0 for both TSS on line 5 and 6, so the program defaults to using the first bed entry which shows the input bed as being upstream whereas it is downstream to the TSS on line 6. Maybe this could be resolved by merging overlapping beds before you calculate distance?

RavenYL: I wonder how this will look in practice with Chelsea's tandem repeats and gene transcripts. I imagine there will only be 1 gene transcript for each gene to compare to each tandem repeat? If there are multiple gene transcripts for each gene, then as long as the Name field specifies the transcript variant, then there won't be ambiguous entries, so there might not be a need to merge in practice?

cshoben: @dasnellings Close on the workflow. We are not looking genome wide for the closest TSS, but matching on the Name field of the two beds, and then determine the minimum distance between them. I will be inputing a genome wide bed of all genes with the coordinates being only the TSS of each gene. 
Great catch! You point out something in that I didnt quite account for which is there being matching genes where the TR ends up being in one instance upstream and another downstream. I wrote in this case because of my realization that you can have many ensembl transcripts for one ensembl gene, and I liked the idea of being able to handle them both. However, I do not anticipate needing this for how I've already pulled the data for my TSS genomeBed. So I think it will be best to remove this case from my test file and make a note in my code that all Names in the genomeBed must be unique.

cshoben: @RavenYL Yes! You are correct. There does appear to be only one transcript whose expression matches to any given eSTR. But the naming convention is neither normal ensembl nor UCSC so I was working through an odd case that I'm now thinking isnt worth my time. 

I'm going to remove this case from my code and make sure my usage statement covers this expectation. 

Thank you both for your help!

bed/compare.go
```diff
@@ -162,10 +162,10 @@ func CompareChromEndByChrom(a Bed, b Bed) int {
 	return CompareChromEnd(a, b)
 }
 
-//CompareDistance compares beds by chromStart and chromEnd and returns the minimum distance between
+//MinimumcDistance compares beds by chromStart and chromEnd and returns the minimum distance between
```
rimangan, line 165: small typo here

rimangan (approved):

RavenYL (approved):

bed/compare.go
cshoben: fixed! requires a new approval please.

dasnellings (approved):

RavenYL (approved):

rimangan (approved):

PR #568, opened 2021-12-11
Fasta seek reading

dasnellings:
Allows for random access of fasta files by using a index file. This way we don't need to read in the entire fasta before processing. Dependent on my previous PR.


RavenYL (dismissed):

fasta/seeker.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Close the Seeker.
+func (rs *Seeker) Close() error {
+	return rs.file.Close()
+}
+
+// TODO
+// readSeekCloser is a TEMPORARY github actions is updated as the io.ReadSeekCloser
+// was added to the builtin io package in 2020.
```
craiglowe, line 24: Am I correct in that this is extra code to make the code pass tests because the testing setup uses an older version of go?  Can we update the testing environment to use a newer version of Go?

fasta/seeker.go
```diff
@@ -0,0 +1,100 @@
...
+func SeekByName(sr *Seeker, chr string, start, end int) []dna.Base {
+	idx, ok := sr.idx.nameMap[chr]
+	if !ok {
+		log.Fatalf("ERROR: could not find sequence for fasta record '%s'\n", chr)
+	}
+	return seek(sr, sr.idx.chroms[idx], start, end)
+}
+
+// SeekByIndex returns a portion of a fasta sequence identified by chromosome index (order in file).
+// Input start and end should be 0-based start-open end-closed.
```
craiglowe, line 62: To make this match other places in go and most of gonomics.  Can we make this start-closed, end-open?

fasta/seeker.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil && err != io.EOF {
+		log.Panic(err)
+	}
+	data = data[:bytesRead] // in case of read past EOF
+
+	answer := make([]dna.Base, end-start)
+	var j int
+	for i := range data {
+		if data[i] == '>' { // in case of read into next fasta record
+			break
```
craiglowe, line 91: Should this be an error?  I think this person is trying to read off the end of the chrom

fasta/seeker.go
```diff
@@ -0,0 +1,100 @@
...
+	_, err := sr.file.Seek(int64(startOffset), io.SeekStart)
+	exception.PanicOnErr(err)
+
+	data := make([]byte, endOffset-startOffset)
+	var bytesRead int
+	bytesRead, err = sr.file.Read(data)
+	if err != nil && err != io.EOF {
+		log.Panic(err)
+	}
+	data = data[:bytesRead] // in case of read past EOF
```
craiglowe, line 85: Should this be an error if they ask to read past the end of the file?

craiglowe (commented):
Very cool!  I had a few small comments

fasta/seeker.go
dasnellings: Ah, it actually is. I just screwed up the comment. Fixed.

fasta/seeker.go
dasnellings: That is correct. I looked briefly into the problem, but could not figure out how to update the go version in the github action.

fasta/seeker.go
```diff
@@ -0,0 +1,101 @@
...
+	idx, ok := sr.idx.nameMap[chr]
+	if !ok {
+		log.Fatalf("ERROR: could not find sequence for fasta record '%s'\n", chr)
+	}
+	return seek(sr, sr.idx.chroms[idx], start, end)
+}
+
+// SeekByIndex returns a portion of a fasta sequence identified by chromosome index (order in file).
+// Input start and end should be 0-based start-closed end-open.
+func SeekByIndex(sr *Seeker, chr, start, end int) []dna.Base {
```
dasnellings, line 64: TODO Add check to make sure the start position is in the requested chromosome

dasnellings: Also, return an error if start is within chromosome and end goes past the requested chromosome

rimangan (dismissed):

RavenYL (approved):

rimangan (approved):

craiglowe (approved):
Great!

PR #573, opened 2021-12-15
multFaVisualizerTests

rimangan:
Testing coverage for multFaVisualizer


dasnellings (dismissed):

cshoben (dismissed):

RavenYL (dismissed):

craiglowe (commented):
These tests look great for making sure that other changes in the code base don't break this program, but it is hard as a human to verify correctness and make sure that the start/end is not one base off, or bugs like that.  Can we also have a small tests where the sequences are just 10 or 20 bp long so that we can see easily verify that the output matches what we would expect?

cshoben (approved):

dasnellings (approved):

craiglowe (approved):
Great!

PR #574, opened 2021-12-15
delete these commands? Riley is going to check old scripts

cshoben:


rimangan (approved):
Good to delete. Nothing here can't be replaced by a few unix commands. I believe I used bedAnnotate during my rotation for some epigenomic annotation stuff but there are more modern methods i've learned since then and this was checked into the code base long after it was used.

RavenYL (approved):

craiglowe (approved):
Great progress!

PR #578, opened 2021-12-16
vcfFormat testing coverage

rimangan:
Small bug where fixDash return was not caught in the cmd rendered the fix option broken. Fixed here with testing coverage


cmd/vcfFormat/vcfFormat.go
```diff
@@ -16,7 +17,7 @@ import (
 func vcfFormat(infile string, outfile string, ensemblToUCSC bool, UCSCToEnsembl bool, fixVcfRecords bool, ref string, clearInfo bool) {
 	ch, header := vcf.GoReadToChan(infile)
 	out := fileio.EasyCreate(outfile)
-	defer out.Close()
+	var err error
 
 	var refMap map[string][]dna.Base
```
RavenYL, line 22: Because a map is involved here, does v need to be sorted before writing to a vcf, or is that not necessary?

craiglowe (approved):
Great!

cmd/vcfFormat/vcfFormat.go
rimangan: we just reference this map, we don't write anything from it, so no need to sort

dasnellings (approved):

cshoben (approved):

PR #583, opened 2021-12-16
Raven cleanup SamToFa

RavenYL:
- Added testing coverage for samToFa
- Modified samToFa.go to specify that the vcfFile is an output not an input
- Replaced "defer file close" with close + error check for samToBed and samToFa. Will do samToWig in the next PR when I get samToWig updates
- Remaining question: Should []*voteBase and &voteBase{0,0,0,0,0} remain pointers or should I remove the pointers? If I remember correctly, we've talked about removing the slices of pointers and just using slices of variables?


cmd/samToFa/samToFa.go
```diff
@@ -170,9 +177,9 @@ type voteBase struct {
 
 func usage() {
 	fmt.Print(
-		"samToFa - Generate a fasta file from a sam over a refrence sequence. Uncovered sequences are converted to lowercase reference sequences.\n" +
+		"samToFa - Generate a fasta file and accompanying vcf from a sam over a refrence sequence. Uncovered sequences are converted to lowercase reference sequences.\n" +
```
craiglowe, line 180: A little more info might be good here.  I see that our test file has some insertions in the read mapping, but I think this is only doing SNPs?  I could image a "samToFa" that extracts the read sequences out of a sam file, but this is really a mutation caller.  Maybe samToSnps or samToVcf would be a better name?

dasnellings (dismissed):

rimangan (approved):

cshoben (approved):

dasnellings (approved):

PR #584, opened 2021-12-16
girafSort testing

dasnellings:


RavenYL (approved):

cmd/girafSort/girafSort_test.go
```diff
@@ -0,0 +1,42 @@
...
+	topoOrder := girafSort(readsFile, graphFile, 100, outFile)
+	actual := giraf.Read(outFile)
+	relabelIdBySortOrder(topoOrder, actual)
+
+	var lastId uint32
+	for i := range actual {
+		if i == 0 {
+			lastId = actual[i].Path.Nodes[0]
+		}
+		if actual[i].Path.Nodes[0] < lastId {
```
craiglowe, line 22: Should "<" be "<=" since ids should be unique?

cmd/girafSort/girafSort_test.go
```diff
@@ -0,0 +1,42 @@
...
+			t.Errorf("problem with girafSort")
+		}
+		lastId = actual[i].Path.Nodes[0]
+	}
+
+	fileio.EasyRemove(outFile)
+	fileio.EasyRemove(outFile + ".idx")
```
craiglowe, line 29: I am not seeing where this file gets created

cmd/girafSort/girafSort_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/giraf"
+	"testing"
+)
+
+func TestGirafSort(t *testing.T) {
+	graphFile := "../../sort/testdata/mini.gg"
+	readsFile := "../../sort/testdata/miniReads.giraf"
+	outFile := "miniReads_sorted_test.giraf"
```
craiglowe, line 12: Would probably be good to keep files be read or written during testing to the testdata directory

craiglowe (commented):
Some comments would help me with this one.  I am having trouble following all the logic

cmd/girafSort/girafSort_test.go
dasnellings: I opted for this since there is no testdata directory, but I see your point. Would you feel ok with ioutil.TempFile?

cmd/girafSort/girafSort_test.go
dasnellings: duplicate Ids are allowed. This is sorting reads aligned to a graph, but not the graph itself. Mulitple reads can align to the same node.

cmd/girafSort/girafSort_test.go
dasnellings: It is created by the index function in sortGiraf.

cshoben (approved):

rimangan (approved):

craiglowe (approved):

PR #591, opened 2021-12-16
continuous function testing in plotFunctions

rimangan:
plotFunctions samples points from a distribution function for plotting. This test covers all continuous functions in the program, and discrete functions have yet to be done in the package (logged as an issue). 

I plotted each output in R to confirm the right shape. Here's the normal distribution (0, 0.5) as an example.
Screen Shot 2021-12-16 at 3 04 07 PM


RavenYL (approved):

cmd/plotFunctions/testdata/expected.beta.txt
```diff
@@ -0,0 +1,101 @@
+X	f(X)
+0.001000	10.070879
```
craiglowe, line 2: Are these numbers from a trusted source, such as R?  I am wondering what we will do if we miss a test by a little bit.  How do we know which (newer/older) is more accurate.

craiglowe (commented):
These look good.  Where are the numbers from?

cshoben (approved):

craiglowe (approved):

PR #596, opened 2021-12-17
no longer compare vcf headers (date mismatch)

craiglowe:
The tests fails because the current date is put in the vcf header, which causes the generated file to not match the expected test file.  I changed the test to not compare comment lines.  I think this is an improvement, but probably not the optimal solution since some comment lines in the vcf format are not really comment lines, but influence the parsing of the data lines, like the sample names line.

### DIFF
--- a/cmd/samConsensus/samConsensus_test.go
+++ b/cmd/samConsensus/samConsensus_test.go
@@ -28,7 +28,7 @@ func TestSamConsensus(t *testing.T) {
 		}
 		exception.PanicOnErr(err)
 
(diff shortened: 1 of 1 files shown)


dasnellings (approved):

cshoben (approved):
I made updates to this is a PR yesterday so that the date line wasnt being generated anymore. I think maybe you were working with a version of gonomics main that didnt have this update yet? But either way, should be fixed now. 

See Test coverage multiFaToVcf #593  changes to vcf/header.go

craiglowe:
If this is fixing an outdated problem then I will close it.  Closing now.

PR #597, opened 2021-12-17
use log1p

craiglowe:
Changed code to use Log1p(x) instead of Log(1+x)

### DIFF
--- a/numbers/logSpace.go
+++ b/numbers/logSpace.go
@@ -37,12 +37,12 @@ func AddLog(x float64, y float64) float64 {
 	}
 	if x >= y {
 		if LogCanConvert(y - x) {
-			return x + math.Log(1+math.Exp(y-x))
+			return x + math.Log1p(math.Exp(y-x))
 		}
 		return x
 	}
 	if LogCanConvert(y - x) {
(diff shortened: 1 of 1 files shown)


numbers/logSpace.go
```diff
@@ -37,12 +37,12 @@ func AddLog(x float64, y float64) float64 {
 	}
 	if x >= y {
 		if LogCanConvert(y - x) {
-			return x + math.Log(1+math.Exp(y-x))
+			return x + math.Log1p(math.Exp(y-x))
```
cshoben, line 40: What does "Log1p" do instead?

cshoben (approved):
The miniPR's are the best!!!

numbers/logSpace.go
craiglowe: I am not sure how the underlying approximation is different, but Log1p is a method in the standard library that has been specifically designed to more accurately handle taking the log of 1 plus a variable.  I believe the approximation is especially tuned for when the variable is small.  It looks like this does not noticeably change any of our test calculations, but is an improvement in theory.

RavenYL (approved):

rimangan (approved):

PR #599, opened 2021-12-17
Remove nils sam consensus

cshoben:
samConsensus output contained "nils" that do not meet standard VCF requirements. 
https://samtools.github.io/hts-specs/VCFv4.2.pdf
I've done my best to change the codebase to ensure output is meeting VCF requirements including changes to samConsensus and multiFaToVcf. 
The only thing I am not confident on is if putting a "." for the FORMAT field is okay. Having what we had there before ("SVTYPE=SNP") was not meeting requirements, and the use of "nil" is not standard format. It might be best if we have the FORMAT field removed from the VCF if all rows contain "."? This could be something for a future PR but worth discussing now I think.


cmd/samConsensus/samConsensus.go
```diff
@@ -83,23 +79,21 @@ func samConsensus(samFileName string, refFile string, outFile string, vcfFile st
...
-	for i = 0; i < len(ref); i++ {
-		for k = 0; k < len(ref[i].Seq); k++ {
+	for i := range ref {
+		for k := range ref[i].Seq {
 			current = voter(votingMatrix[ref[i].Name][k], maxList)
 			if current == dna.N && ref[i].Seq[k] != dna.N {
 				current = dna.ToLower(ref[i].Seq[k])
 			} else if current != ref[i].Seq[k] {
-				fmt.Fprintf(outVcfFile, "%s\t%d\t%s\t%s\t%s\t%s\t%d\t%s\t%s\t%s\t%s\n", ref[i].Name, int64(k+1), "nil", dna.BaseToString(ref[i].Seq[k]), dna.BaseToString(current), "nil", 0, "nil", "SVTYPE=SNP", "nil", "nil")
+				fmt.Fprintf(outVcfFile, "%s\t%d\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n", ref[i].Name, int64(k+1), ".", dna.BaseToString(ref[i].Seq[k]), dna.BaseToString(current), ".", ".", ".", ".")
```
RavenYL, line 89: What if instead of "." we just have "", since that is also a valid string? Would that create an output file that looks ok and  fits vcf requirements?

RavenYL (commented):
Thank you for catching this! I didn't even think to check if our outputs fit file format requirements

rimangan (approved):
Great fix!

cshoben:
I totally credit Craig for catching this!

cmd/samConsensus/samConsensus.go
cshoben: I mirrored what was in the VCF documentation. I was under the impression it couldn't be blank but let me know if you read something that suggests otherwise.

RavenYL: I see! If nothing says we can't have "." then it's probably fine?

RavenYL (approved):

craiglowe (approved):
Great work!

PR #600, opened 2022-01-07
adding prior to mcmc

craiglowe:
This is a proposal to add priors to the mcmc calculations.  I think there is room for improvement here, but I also think this is a good first step.


rimangan (dismissed):

popgen/mcmcSelection.go
```diff
@@ -97,10 +119,13 @@ func GenerateCandidateThetaPrime(t Theta, data Afs, binomCache [][]float64, s Mc
 	} else {
 		likelihood = AfsLikelihood(data, alphaPrime, binomCache, s.IntegralError)
 	}
+
+	prior = priorProb(muPrime, sigmaPrime, s)
```
craiglowe, line 123: One option would be to move this prior calculation up to before the t.alpha loop that calculates the likelihood.  That would allow us to not calculate the likelihood (report -inf or 0) if the prior density is -inf/0.  Maybe it is nice to have the likelihood even when the prior excludes the proposed parameters?  Maybe it would be worth the speed-up to not do the likelihood calculation?  Could even make it an option.

craiglowe (commented):
Looks great Riley!  Just a couple small notes.

popgen/mcmcSelection.go
rimangan: in next commit

rimangan (dismissed):

cshoben (approved):

cshoben (approved):

rimangan (approved):

RavenYL (approved):

PR #601, opened 2022-01-13
proximityBlockVcf

rimangan:
proximityBlockVcf is a new cmd to control for the effects of linkage disequilibrium when analyzing allele frequencies in variant data. Briefly, this program takes an input Vcf file, shuffles the variants, and randomly selects variants to be retained in the output file. If a randomly selected variant falls within a user-specified distance of another variant already selected for the output, it is discarded. Thus, this program could for example pseudorandomly select all variants in a file that are no closer than 10kb of one another. Passes tests.

### DIFF
--- a/cmd/proximityBlockVcf/proximityBlockVcf.go
+++ b/cmd/proximityBlockVcf/proximityBlockVcf.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/common"
+	"github.com/vertgenlab/gonomics/exception"
+	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/vcf"
+	"log"
+	"math/rand"
+)
+
+func proximityBlockVcf(inFile string, outFile string, distance int, randSeed bool, setSeed int64) {
+	var err error
+	common.RngSeed(randSeed, setSeed)
+	records, header := vcf.Read(inFile)
+	rand.Shuffle(len(records), func(i, j int) {records[i], records[j] = records[j], records[i]})
+
+	//make an output slice and put the first entry in it.
+	var retainedVcfs []vcf.Vcf = make([]vcf.Vcf, 1)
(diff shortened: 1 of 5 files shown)


dasnellings (approved):

craiglowe (approved):
Wow!  Amazing how fast you were able to write this.  It is nice coding on your part and also nice to see how useful the gonomics library has become.

cmd/proximityBlockVcf/proximityBlockVcf.go
```diff
@@ -0,0 +1,87 @@
...
+	flag.Parse()
+
+	if len(flag.Args()) != expectedNumArgs {
+		flag.Usage()
+		log.Fatalf("Error: expecting %d arguments, but got %d\n",
+			expectedNumArgs, len(flag.Args()))
+	}
+	inFile := flag.Arg(0)
+	outFile := flag.Arg(1)
+	distance := common.StringToInt(flag.Arg(2))
```
RavenYL, line 85: Is there a reasonable distance to set as a default?

rimangan: hard to say. Since I have it configured as an argument we don't need one, but I was running this with 10k on real data

cmd/proximityBlockVcf/proximityBlockVcf.go
```diff
@@ -0,0 +1,87 @@
...
+			"that do not fall within a user-specified distance" +
+			"of variants already chosen. Output is returned in a shuffled order.\n" +
+			"Usage:\n" +
+			"proximityBlockVcf input.vcf output.vcf distance" +
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var randSeed *bool = flag.Bool("randSeed", false, "Uses a random seed for the RNG.")
```
cshoben, line 70: I ended up needing to google around to figure out that RNG stands for 'random number generator'. Not a big deal, probably most know this acronym, but I think generally it'd benefit new users to type out acronyms within usage statements.

cshoben (approved):

PR #602, opened 2022-01-19
Revoking randSeed Privileges

rimangan:
i don't trust you


dasnellings:
I hear you, but it is important for some of these functions. Specifically the simulate functions. "Sure gonomics can simulate a random VCF, but you only get ONE..."

rimangan:
> I withdraw my comment. However, lets kill the common.RngSeed function altogether. Just use the rand.Seed function instead. We don't need a middleman.

This is a good point. I initially had the common function to handle the multiple arguments, so now it doesn't make sense lol. Will fix on next commit

rimangan:
By removing common.RngSeed and working with rand.Seed we have one small behavior change. Previously, "-1" serves as a default value for rand.Seed that tells the program not to use a random seed, and so it would use the default seed of "1". Obviously this is pretty wonky so it's fixed here, so some seeds in programs changed as a result to comply with tests.

RavenYL (approved):
Looks good to me! Looks like we're removing the pseudorandom number generator in gonomics (gonomics/common/math.go > RngSeed) and replacing it with the rand in the standard golang math package (math/rand)

dasnellings (approved):

craiglowe (approved):
Thank you!

PR #605, opened 2022-01-23
Craig cut update

craiglowe:
This is a draft proposal of some code that may speed up the "faFindFast" program for genome-wide runs, especially on larger window sizes.  It is a branch off another draft pull request with improvements to faFindFast


craiglowe:
This is now ready to review

rimangan (dismissed):
Looks great! Do you know yet if there was anything in the chromosome wide diff from the beds made by the two algorithms? If so, should we go ahead and remove the slow version now?

craiglowe:
Yes, there are genome-wide diffs; however, from the sets I have run so far, they are all cases were there are Ns in the query are aligned to gaps in the reference.  The newer algorithm counts these as an N in the alignment, while the older version does not.  I think this newer output is more what the user would expect, so I am allowing for this difference.
This pull request eliminates the code for the older version; it will always be in older versions of the repo if we want to go back to it.

rimangan (approved):

cshoben (approved):

dasnellings (approved):

PR #606, opened 2022-01-24
Craig primate recon update

craiglowe:
Draft of code to do an mle reconstruction of the hca


cmd/primateRecon/primateRecon.go
```diff
@@ -5,29 +5,133 @@ package main
...
+
+	for i = range likes {
+		total += likes[i]
+		if i != int(humanBase) {
+			nonHumanTotal += likes[i]
+		}
+	}
+
+	for i = range likes {
+		if likes[i]/total >= probThreshold && likes[i] > bestProb && nonHumanTotal/total >= nonHumanProbThreshold {
```
rimangan, line 39: it might be good in the documentation to explain how the nonHumanThreshold and propThreshold interact. I wonder if you might have some unintuitive results for what is accepted or not accepted given that we have an AND in this statement

cmd/primateRecon/primateRecon.go
```diff
@@ -5,29 +5,133 @@ package main
...
+		return true
+	}
+	return false
+}
+
+func hcaIsPresent(human, bonobo, chimp, gorilla, organutan dna.Base) bool {
+	if baseIsPresent(human) && (baseIsPresent(bonobo) || baseIsPresent(chimp)) {
+		return true
+	}
+	if (baseIsPresent(human) || baseIsPresent(bonobo) || baseIsPresent(chimp)) && (baseIsPresent(gorilla) || baseIsPresent(organutan)) {
```
rimangan, line 58: in the manual algorithm, we would return N if human and gorilla were present but not bonobo and chimp, with the worry of estimating over the longer branch. Does the usage of branch lengths in the ML help us here? Right now it looks like we'll evaluate a base at a position with no bonobo or chimp

cmd/primateRecon/primateRecon.go
```diff
@@ -5,29 +5,133 @@ package main
...
+		return true
+	}
+	return false
+}
+
+func reconHcaBase(root, humanNode, nodeToRecon *expandedTree.ETree, position int, probThreshold float64, nonHumanProbThreshold float64, messyToN bool) {
+	reconstruct.SetState(root, position)
```
rimangan, line 65: Is reconstruct all normal space floats? Should we consider a switch to log space floats? Won't big trees hit underflow pretty quickly since we're calculating a product?

rimangan (approved):

dasnellings (approved):

dasnellings:
When this gets merged it may be worth making the first release of gonomics so in the paper you can refer people to `gonomics v0.1` instead of `gonomics commit fb2bf53`

cmd/primateRecon/primateRecon.go
```diff
@@ -5,29 +5,175 @@ package main
...
+func primateReconMle(inFastaFilename string, inTreeFilename string, humanBias bool, probThreshold float64, nonHumanProbThreshold float64, outputFastaFilename string) {
+	var tree, humanNode, humanAltNode, bonoboNode, chimpNode, gorillaNode, orangutanNode, hcaNode *expandedTree.ETree
+	var err error
+	var i int
+
+	tree, err = expandedTree.ReadTree(inTreeFilename, inFastaFilename)
+	exception.FatalOnErr(err)
+
+	// roll call to make sure everyone is here and will need them later
+	humanNode = expandedTree.FindNodeName(tree, "hg38")
```
RavenYL, line 108: Is this program specific to certain reference genome assemblies (e.g. hg38, gorGor5)? To me it looks like the user needs to provide inputs where the program can find the specific genome assemblies. Does the specific genome assembly requirement need to be specified in either usage or the documentation?

RavenYL (approved):

PR #607, opened 2022-01-25
MultFaToVcf Alt Starts With Gap Panic Fix

rimangan:
MultiFaToVcf is a program that takes in a pairwise multiple alignment file in multiFa format and produces a vcf file with one line for each difference observed between the two sequences in the multiFa.
Seth noticed that the following alignment produces a panic:
>Ref
ATCAAC
>Alt
---AAG
The Reason for this is we would call the base preceding the first A of the ref as the position of the deletion in Alt, so it would attempt to look at the base identity in slice position [-1] of the Ref sequence, producing a panic. A small change fixed this here.

### DIFF
--- a/cmd/multiFaToVcf/multiFaToVcf_test.go
+++ b/cmd/multiFaToVcf/multiFaToVcf_test.go
@@ -19,7 +19,7 @@ var MultiFaToVcfTests = []struct {
 	{"testdata/inputMulti.fa", "chr2", "testdata/output.vcf", "testdata/expectedSubOnly.vcf", true, false},
 	{"testdata/inputMulti.fa", "chr2", "testdata/output.vcf", "testdata/expectedRetainN.vcf", false, true},
 	{"testdata/inputStartWithGap.fa", "chr2", "testdata/startGapTmp.vcf", "testdata/expectedStartGap.vcf", false, false},
-	{"testdata/inputAltStartWithGap.fa", "chr2", "testdata/startGapAltTmp.vcf", "testdata/expectedAltStartsWithGap.vcf", false, false},//TODO: This expected file does not contain a deletion at position 1, as our parser cannot handle this case.
+	{"testdata/inputAltStartWithGap.fa", "chr2", "testdata/startGapAltTmp.vcf", "testdata/expectedAltStartsWithGap.vcf", false, false}, //TODO: This expected file does not contain a deletion at position 1, as our parser cannot handle this case.
 }
 
 func TestMultiFaVcf(t *testing.T) {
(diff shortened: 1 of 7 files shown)


convert/convert.go
```diff
@@ -229,7 +229,7 @@ func PairwiseFaToVcf(f []fasta.Fasta, chr string, out *fileio.EasyWriter, substi
 		} else if deletion {
 			pastStart = true
 			deletion = false
-			if !substitutionsOnly {
+			if !substitutionsOnly && deletionAlnPos >= 0 {
```
dasnellings, line 232: Instead of ignoring this corner case, is it feasible to match the VCF specs here?

From the specs:

"Strings must include the base before the event (which must be reflected in the POS field), unless the event occurs at position 1 on the contig in which case it must include the base after the event"

dasnellings (changes requested):

convert/convert.go
rimangan: Do you have an idea of what that looks like for the alignment I showed above?
Would that work like this:
chr1 1 ATCA A

dasnellings: Yes, that is my understanding.

craiglowe:
I am noticing that the chain package appears to have hit an index out of range while doing coverage testing, which is odd.  Maybe that is a bug in our code, or maybe it is a random error that is not related to our code?

dasnellings:
> I am noticing that the chain package appears to have hit an index out of range while doing coverage testing, which is odd. Maybe that is a bug in our code, or maybe it is a random error that is not related to our code?

That is very strange. It passed the test but failed the coverage (which as I understand it just runs the test). I reran and they all passed. Could there be some instability in the chain package?

convert/convert.go
rimangan: Added a TODO and updated the usage to reflect that we're not fixing this. This deletion in position 1 will break our parser I think

craiglowe: Do we need to update the parser to treat mutations at position 1 as a special case?  Maybe not needed immediate, but it would be good to not lose site of this inability to handle deletions of the first base.  Can we discuss this at code review on Monday to try and understand how big of a fix it would be to handle these types of mutations?  I am ok to check in this improvement now, and have that discussion on Monday about a fix to the larger issue

craiglowe (dismissed):
Thank you for fixing this crash!  I want to make sure we don't lose site of a larger VCF fix that I think we need sometime in the future

dasnellings (dismissed):

cmd/multiFaToVcf/multiFaToVcf_test.go
```diff
@@ -18,6 +18,8 @@ var MultiFaToVcfTests = []struct {
 	{"testdata/inputMulti.fa", "chr2", "testdata/output.vcf", "testdata/expected.vcf", false, false},
 	{"testdata/inputMulti.fa", "chr2", "testdata/output.vcf", "testdata/expectedSubOnly.vcf", true, false},
 	{"testdata/inputMulti.fa", "chr2", "testdata/output.vcf", "testdata/expectedRetainN.vcf", false, true},
+	{"testdata/inputStartWithGap.fa", "chr2", "testdata/startGapTmp.vcf", "testdata/expectedStartGap.vcf", false, false},
```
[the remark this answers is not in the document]
rimangan: In latest commit

PR #608, opened 2022-01-26
Dunn testing

rimangan:
dunnIndex command on main has the bed package "FieldsInitialized" error. Fixed here. Added testing.
Also added a small check with a fatal condition: we fatal there is already information in the annotation section
of a bed entry, as dunnIndex would overwrite this.


craiglowe (approved):
Great to have these tests!  Looks like it may also have files from the pull request on multiFaToVcf

cshoben (approved):
Waiting to see if the multiFaToVcf file changes were intended before merging.

PR #609, opened 2022-01-26
Remove tajimaD

rimangan:
Two years ago I made an implementation of Tajima's D on our code base, a heuristic for selection based on the shape of the derived allele frequency spectrum. We then transitioned to the MCMC model and never used this code again. I'm also not sure if it works and I don't see myself returning to it. This is one situation where I think it is better to remove a program than write tests for it. I've kept the implementation as a comment block in dunn.go in the popgen package.


popgen/dunn.go
```diff
@@ -73,3 +73,331 @@ func FindMinInter(g []*Group, subFa []fasta.Fasta) int {
 	}
 	return answer
 }
+
+//below is an implementation of the calculations for Tajima's D. Not currently used in the
```
craiglowe, line 77: Should this be removed?  It will always be in the history of the repository, so it will not be lost.  Having it be commented out in dunn.go, might not be a place that someone will find it, unless they know about it, and would probably re-write themselves.  If you want to keep it commented out, maybe create a file in popgen that is something like "oldCode.go" which has commented out code?

craiglowe (commented):
Great to have this package getting polished!  I would consider either getting rid of the code, or putting it in a file with a better fitting name.  This could be on a separate branch from the vcf bug fix.

popgen/dunn.go
cshoben: I like the idea of "oldCode" because that is more easily searchable than knowing where to go back in the codebase history to find something. There could even be a title page with the name of each function and usage statements to help people find what they are looking for.

cshoben (approved):

dasnellings:
Merge main?

rimangan:
> Merge main?

Fixed thanks

dasnellings (approved):

RavenYL (approved):

PR #610, opened 2022-01-26
Variant Calling via Pileup

dasnellings:
First, I apologize. This is a major overhaul so it may be a bit of a bear to review. This is updating the variant caller in various ways to 

1) make the output VCF spec compliant (though still a couple corner cases I am working on)

2) call variants from bam files (previous implementation required sam files)

3) call variants based on the pileup code which is faster and likely more stable than the previous implementation I used. 

4) enable me to delete old packages of mine that are not worthy of testing (in following PR). 

The testing could be better. I am currently working on a sam simulation program that should make testing a bit more robust than it currently is.


cmd/callVariants/testVariant.go
```diff
@@ -0,0 +1,359 @@
...
+
+	// pull the base before, as well as the current base in case an anchor is need for an indel
+	var seekStart, seekEnd int // Note that the Pos in pile is 1-base
+	seekStart = int(bkgd.Pos) - 2
+	seekEnd = int(bkgd.Pos)
+	if seekStart == -1 {
+		seekStart = 0 // TODO get some logic to add the base after a start-of-chromosome deletion for anchor
+	}
+	refBases, err := fasta.SeekByIndex(ref, bkgd.RefIdx, seekStart, seekEnd) // The Pos in pile is 1-base
+	dna.AllToUpper(refBases)                                                 // very important to get the correct index in Pile.Count since lowercase is not allowed in sam/bam
```
craiglowe, line 54: I am constantly tempted by a reduction in the dna package that would eliminate the lower case letters

cmd/callVariants/testVariant.go
```diff
@@ -0,0 +1,359 @@
...
+	} else { // if no normals are present then use experimental samples for comparison
+		bkgd = sumPiles(exp)
+	}
+
+	// pull the base before, as well as the current base in case an anchor is need for an indel
+	var seekStart, seekEnd int // Note that the Pos in pile is 1-base
+	seekStart = int(bkgd.Pos) - 2
+	seekEnd = int(bkgd.Pos)
+	if seekStart == -1 {
+		seekStart = 0 // TODO get some logic to add the base after a start-of-chromosome deletion for anchor
```
craiglowe, line 51: This can also be included in our code review discussion about vcf handling the deletion of the first ref base.

cmd/callVariants/testVariant.go
```diff
@@ -0,0 +1,359 @@
...
+			pVals[i] = passingAltPvalues[i][sIdx]
+		}
+	}
+
+	for i := range alleleCounts {
+		if alleleCounts[i] > 0 {
+			genotypeAlleles = append(genotypeAlleles, int16(i))
+		}
+	}
+	if len(genotypeAlleles) == 1 { // make homozygotes diploid //TODO kind of a hack to assume diploidy, how do others determine ploidy
```
craiglowe, line 179: This seems like a reasonable way to do it.

cmd/callVariants/testVariant.go
```diff
@@ -0,0 +1,359 @@
...
+		genotypeAlleles = append(genotypeAlleles, genotypeAlleles[0])
```
craiglowe, line 180: Why do we make haploids into diploids?

cmd/callVariants/testVariant.go
```diff
@@ -0,0 +1,359 @@
...
+	case a+c < minCoverage:
+		p = 1
+
+	// If the allele frequency is less than the threshold then p is noted as 1 so as to be excluded
+	case float64(c)/float64(c+a) < minAf:
+		p = 1
+
+	// If no exclusion conditions are met, then calculate p value
+	default:
+		if a < c && b < d { // alt is common allele
```
craiglowe, line 352: Do we need to multiply these pvalues by 2 since we are checking to see if a lower tail or upper tail test would give a more extreme value (essentially doing both tests and reporting the one that is most extreme).

craiglowe (commented):
Looks like great work!  Wonderful to have this!  I had a few comments and maybe you can present this at code review for a few minutes?  I think there is some cool stuff here that could help others.

cmd/callVariants/callVariants.go
```diff
@@ -18,133 +19,130 @@ import (
...
-	case ".sam":
-		*alleleChans = append(*alleleChans, alleles.GoCountSamAlleles(file, ref.([]fasta.Fasta), minMapQ))
-		*samFilesPresent = true
-		if isNormal == true {
-			normalIDs[file] = true
-		} else {
-			normalIDs[file] = false
+		piles[i] = sam.GoPileup(samChan, headers[i], false, nil, nil)
+	}
+	return
```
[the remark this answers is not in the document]
dasnellings: Even with named variables you need a return at the end of the function, just the variables to return are pre-defined

cmd/callVariants/callVariants.go
```diff
@@ -18,133 +19,130 @@ import (
...
-	}
-	if len(alleleChans) < 2 {
-		log.Fatalln("ERROR: Must input at least two samples (between experimental and normal) to facilitate comparisons. Only", len(alleleChans), "sample was submitted")
-	}
+// inputFiles is a custom type that gets filled by flag.Parse()
+type inputFiles []string
 
-	syncedAllelesChan := alleles.SyncAlleleStreams(ref, memBufferSize, alleleChans...)
+// String to satisfy flag.Value interface
+func (i *inputFiles) String() string {
```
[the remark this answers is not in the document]
dasnellings: Sounds good

cmd/callVariants/testVariant.go
dasnellings: Yes, though I think we really just need to start using 3bit. I'll be the first to admit I have been pretty bad about it. I need to spend some time to familiarize myself with it and figure out where I should be using 3bit instead of Base.

cmd/callVariants/testVariant.go
dasnellings: This was added when I was debugging, but I found a different bug which made it unnecessary. It is now changed so aSmall is always true.

craiglowe (commented):
This looks good to me, but having a mini presentation on this pull request would help me get a full understanding of its capabilities.

cmd/callVariants/callVariants.go
```diff
@@ -18,133 +19,130 @@ import (
...
+
+	isExperimental := make([]bool, len(experimentalFiles)+len(normalFiles))
+	for i := 0; i < len(experimentalFiles); i++ {
+		isExperimental[i] = true
 	}
-	alleleStream, normalIDs := startAlleleStreams(ref, expSamples, normSamples, minMapQ, memBufferSize)
-	answer := alleles.FindNewVariation(alleleStream, normalIDs, afThreshold, sigThreshold, minCov)
 
-	lastProgressReport := time.Now()
+	err := checkHeadersMatch(append(expHeaders, normHeaders...))
```
cshoben, line 43: What does this "..." do at the end here?

cmd/callVariants/callVariants.go
```diff
@@ -18,133 +19,130 @@ import (
...
-	var graphReference *string = flag.String("gr", "", "Graph reference used for alignment [.gg].")
-	var experimentalSamples *string = flag.String("i", "", "Input experimental sample(s) [.sam, .giraf, .txt]. Can be a file or a txt file with a list (must have .txt extension) of sample paths.")
-	var normalSamples *string = flag.String("n", "", "Input normal sample(s) [.sam, .giraf, .txt]. Can be a file or a txt file with a list (must have .txt extension) of sample paths. If no normal samples are given, each experimental sample will me measured against the other experimental samples.")
-	var minMapQ *int64 = flag.Int64("minMapQ", 20, "Exclude all reads with mapping quality less than this value")
-	var memBufferSize *int = flag.Int("memBuffer", 100, "Maximum number of allele records to store in memory at once")
-	var minCov *int = flag.Int("minCoverage", 10, "Minimum number of covering reads to be considered a valid variant")
-	flag.Usage = usage
-	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	var experimentalFiles, normalFiles inputFiles
+	flag.Var(&experimentalFiles, "i", "Input experimental files. May be declared more than once (.bam, .sam)")
```
cshoben, line 131: Just to check, does this "may be declared more than once" mean that the program can handle a combination of both .bam and .sam files?

cshoben (approved):

cmd/callVariants/callVariants.go
dasnellings: It can handle a combination of files types, but the main purpose is for joint variant calling for multiple samples. It will process them at the same time and use the other samples to determine p values.

cmd/callVariants/callVariants.go
dasnellings: the append function takes a slice (expHeaders) and adds single elements to the slice. But when you want to append one slice to another it doesn't work. You can get around that by expanding your slice to individual values using `...`. This line is equivalent to 
`append(expHeaders, normHeaders[0], normHeaders[1], normHeaders[2], normHeaders[n])`

craiglowe (approved):
Wonderful!

PR #611, opened 2022-01-26
Delete alleles package

dasnellings:
This is dependent on my previous PR. All it does is delete the alleles package and the countAlleles cmd. It also adds a comment to pileup.go about the position being 1-based.


rimangan (approved):

craiglowe (approved):
Great to have the new version of this!

RavenYL (approved):

PR #612, opened 2022-01-27
prcompFormat testing

rimangan:
Testing for prcompFormat, a function that converts a multiFa alignment into a data matrix that can be easily read into the "prcomp" PCA function built into R. This program only gives the first half of the result (data formatting) so in the testdata directory I include the R script that makes the PCA plot from the actual data. Not sure what people think about adding a bit of R code into the code base, I can always refile this under vglDocumentation if that is preferable. 
I made a small multiple alignment in which some species were more closely related to each other. This phylogenetic organization of species was reflected in the pdf generated by the r script.
Passes tests.


RavenYL (dismissed):
I think it's fine to have R code in the repository since it doesn't interfere with go tests

cmd/prcompFormat/prcompFormat.go
```diff
@@ -42,9 +41,11 @@ func prcompFormat(infile string, outfile string) {
 		allValid = true
 		if !(currentBase == dna.N || currentBase == dna.Gap) {
 			for j := 1; j < len(records); j++ {
-				//fmt.Printf("i = %d, j = %d\n", i, j)
 				if records[j].Seq[i] != currentBase && records[j].Seq[i] != dna.Gap && records[j].Seq[i] != dna.N {
-					fmt.Printf("i = %d, j = %d\n", i, j)
+					if verbose {
```
RavenYL, line 45: Should there be tests for verbose?

cmd/prcompFormat/testdata/pca.r
```diff
@@ -0,0 +1,17 @@
+library("ggplot2")
+
```
cshoben, line 2: If the R isnt necessary for the test files to run, then probably best to keep it in vglDocumentation.

dasnellings: I agree. That said, I would support adding a gonum dependency and making some bona fide plotting packages in gonomics.

rimangan: It's a bit more than plotting, in this case R is calculating the actual principal components, the go program just formats the input data from a multiple alignment. I can move the script to vgl

rimangan: Removed in the next commit, and I've added the PCA script to the vglDocumentation repository and added a readme in the testdata folder to direct the user there.

dasnellings: gonum also does PCA ;)

dasnellings (approved):

cmd/prcompFormat/prcompFormat.go
rimangan: i don't think so, it just gives us extra prints to the screen for debugging.

cmd/prcompFormat/testdata/readme.txt
```diff
@@ -0,0 +1,3 @@
+Please see the vglDocumentation repository for a detailed explanation of
```
craiglowe, line 1: Should this be in the usage message of the program?  My concern is that people trying to figure out how to run this, will not look for documentation in "testdata".  How extensive are the R commands, can you add them as a line or two in the usage message for this program?

craiglowe (commented):
Great to have tests.  I had a comment about putting documentation in "testdata."  I would either put it in the usage message, or consider removing it.

dasnellings (approved):

dasnellings (approved):

craiglowe (approved):
Great!

RavenYL (approved):

PR #614, opened 2022-01-31
Simulate Sam Alignments

dasnellings:
Functions and a cmd for simulating sam alignments to an input reference file. Currently the only support is for illumina-style 150bp paired end reads, however the foundation is there should others wish to expand this for other read types. Also it currently only supports random coverage across the input file, however I may also make a version in the future that supports peaks of coverage.


cmd/simulateSam/simulateSam.go
```diff
@@ -0,0 +1,89 @@
...
+	var totalLen int
+	for i := range ref {
+		totalLen += len(ref[i].Seq)
+	}
+
+	var proportionOfReads float64
+	var readsPerContig int
+	var reads []sam.Sam
+	for i := range ref {
+		proportionOfReads = float64(len(ref[i].Seq)) / float64(totalLen)
```
dasnellings, line 49: fix so read name does not reset on new contig

simulate/sam.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// simulatePairedSam generates a pair of sam reads randomly distributed across the input ref.
+func simulatePairedSam(refName string, ref []dna.Base, numPairs, readLen, avgInsertSize int, avgInsertSizeStdDev float64) []sam.Sam {
+	reads := make([]sam.Sam, numPairs*2)
+
+	var insertSize, midpoint, startFor, startRev, endFor, endRev int
+	for i := 0; i < len(reads); i += 2 {
+		insertSize = int(numbers.SampleInverseNormal(float64(avgInsertSize), avgInsertSizeStdDev))
+		midpoint = numbers.RandIntInRange(-2*readLen, len(ref)+(2*readLen))
```
craiglowe, line 25: I am having trouble following the logic here.  I would have thought that the midpoint would have to come in from the ends (to allow space for the sides of the fragment, without going off the end of the chrom), but it seems like here we are allowing the midpoint to be beyond both ends of the chromosome.

simulate/sam.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		reads[i].PNext = reads[i+1].Pos
+		reads[i+1].PNext = reads[i].Pos
+	}
+	return reads
+}
+
+// generateSamReadNoFlag generates a sam record for the input position.
+// Soft clips sequence that is off template and does not generate Flag, RNext, or PNext.
```
craiglowe, line 53: I think I understand better now a section above where I made a comment.  I think this simulation is assuming that there is sequence beyond the ends of the fragment, that it is more like an assembly gap, and less like the actual end of a DNA fragment.

simulate/sam.go
```diff
@@ -0,0 +1,143 @@
...
+
+	s.MapQ = uint8(numbers.RandIntInRange(30, 40))
+	s.RName = refName
+
+	// generate random seq if off template
+	var realSeqStartIdx, realSeqEndIdx int
+	for realSeqStartIdx = start; realSeqStartIdx < 0; realSeqStartIdx++ {
+		s.Seq[realSeqStartIdx-start] = dna.Base(numbers.RandIntInRange(0, 4))
+	}
+	for realSeqEndIdx = end; realSeqEndIdx > len(ref); realSeqEndIdx-- {
```
craiglowe, line 82: Should this be ">=" instead of ">"?

craiglowe (commented):
Looks great!  I had only a couple minor comments/questions.

cmd/simulateSam/simulateSam.go
```diff
@@ -0,0 +1,89 @@
...
+		readsPerContig = int(float64(numReads) * proportionOfReads)
+		reads = simulate.IlluminaSam(ref[i].Name, ref[i].Seq, readsPerContig)
```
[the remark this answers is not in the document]
dasnellings: I'm not sure I follow the relationship between coverage and fragment length. Currently reads are distributed between chroms based on the length of the contig. A contig of size 2 gets twice as many reads as a contig of size 1. How does the fragment length factor in, and could/should it be added in the main function instead of the IlluminaSam function?

sam/io.go
```diff
@@ -234,17 +234,31 @@ func readNextRecycle(reader *fileio.EasyReader, s *Sam) bool {
 // such that storing the entire file in memory is not feasible. Most
 // sam files should be read using the GoReadToChan function which
 // streams sam records so only a small portion of the file is kept
-// in memory at any given time.
+// in memory at any given time. Reads as a bam if .bam suffix is present.
 func Read(filename string) ([]Sam, Header) {
+	var err error
 	var alignments []Sam
+	if strings.HasSuffix(filename, ".bam") {
```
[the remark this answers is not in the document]
dasnellings: added test case

simulate/sam.go
dasnellings: This was some gymnastics I did to get even coverage across the contig. My initial code generated a gradient where coverage increases at the start of the contig and decreases slowly towards the end of the contig. This lets a read start past the ends of the contig (and get trimmed if it goes off the contig) to ensure even coverage.

simulate/sam.go
dasnellings: That wasn't the goal per se, but that is one way of looking at it.

simulate/sam.go
dasnellings: The interval is `[realSeqStartIdx, realSeqEndIdx)` so `realSeqEndIdx == len(ref)` is a valid index so I believe > is correct.

simulate/sam.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	copy(s.Seq[realSeqStartIdx-start:len(s.Seq)-(end-realSeqEndIdx)], ref[realSeqStartIdx:realSeqEndIdx])
+
+	// generate other values
+	s.Pos = uint32(realSeqStartIdx) + 1
+	s.TLen = int32(realSeqEndIdx - realSeqStartIdx)
+
+	// assemble cigar
+	if realSeqStartIdx > start {
+		s.Cigar = append(s.Cigar, &cigar.Cigar{RunLength: realSeqStartIdx - start, Op: 'S', Sequence: nil})
```
RavenYL, line 93: If I understood correctly, the sequences that would be marked with S in the output sam cigar are those that went beyond the start and end of a contig/chromosome. For downstream applications, would you need to remove those sequences?

RavenYL (dismissed):

simulate/sam.go
dasnellings: The S in the cigar stands for "Soft clip". It is part of the SAM specs and is commonly done by mapping software. Most programs should be able to handle it without a problem.

cmd/simulateSam/simulateSam.go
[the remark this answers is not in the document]
craiglowe: I think there are two things here that are related, but mixed together.
1) How fragment length factors in.  I was thinking that you need to use fragment length to know the probability of getting a read from a section of DNA of a certain length.  For example, I would have thought that a 100bp chromosome would have a 0 percent chance of giving you a fragment that was 300bp long, but I now realize that in this case you add random bases to the end to allow a chromosome of any size to give a fragment of any size.  I should have gone back to my comments after realizing this and modified it.
2) Exact same number of reads on each chromosome.  This will likely be variable in real life and by fixing the simulation to always give the most likely number of reads (and not vary) the simulation may not have as much variance in read coverage as one might see in real world data.  For example.  I think chromosome 22 is about 1 to 2% of the human genome.  If I am understanding the code correctly, if I were to repeatedly simulate 50 reads from the human genome, I may never get a read from chromosome 22.  This is probably made to simulate very large numbers of reads, so this is not a huge concern, but there may be some corner case where the behavior is a bit odd by fixing the simulation to always give the expected number.

craiglowe (commented):
Replying to a few comments

simulate/sam.go
dasnellings: Hmm I think it might depend on your fragmentation method. With something like motif-based enzymatic fragmentation you would probably get major depletion of the fragment ends, but with sonication-based fragmentation you may get a bit of depletion but I would guess the ends would be mostly intact. Either way, I've changed the code to take the midpoint from [0, len(ref)) but left in a commented line for the old method in case it becomes relevant.

cmd/simulateSam/simulateSam.go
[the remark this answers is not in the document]
dasnellings: Ah I understand now. The most recent commit distributes reads randomly, but weighted by contig length.

cmd/simulateSam/simulateSam_test.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	bamA, _ := sam.Read("testdata/actual.bam")
+	bamB, _ := sam.Read("testdata/expected.bam")
+
+	if len(bamA) != len(bamB) {
+		t.Error("problem simulating bam")
+	} else {
+		for i := range bamA {
+			if !sam.Equal(bamA[i], bamB[i]) {
```
cshoben, line 28: Would you expect to find the files equal when they are not the same size? Since the .bam is created by this program I expect it has a very standard output.

cmd/simulateSam/simulateSam_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+	"testing"
+)
+
+func TestSimulateSam(t *testing.T) {
+	rand.Seed(1)
+	simulateSam("testdata/test.fa", "testdata/actual.sam", 100, false)
+	simulateSam("testdata/test.fa", "testdata/actual.bam", 100, true)
+
+	if !fileio.AreEqual("testdata/actual.sam", "testdata/expected.sam") {
```
cshoben, line 17: My understanding is that the output sam involves a random sampling. So how does our output sam end up being our expected every time?

cshoben (approved):

rimangan (approved):

cfauci (approved):

PR #615, opened 2022-02-03
Simulate Determinism

dasnellings:
The simulate package makes often requires the use of the rand package to get random values for data simulation. This has been causing problems where our test functions are not deterministic leading to some lines being tested some of the time. This has been highlighted by the Codecov reports where this behavior led to spurious accusations that my PRs were reducing coverage. In this PR I set the seed to 1 for all tests so Codecov will not slander out future PRs inaccurate reports.

### DIFF
--- a/simulate/simulate_test.go
+++ b/simulate/simulate_test.go
@@ -1,9 +1,9 @@
 package simulate
 
 import (
-	"fmt"
-	"github.com/vertgenlab/gonomics/dna"
 	"github.com/vertgenlab/gonomics/fasta"
+	"math/rand"
+	"os"
 	"testing"
 	//"fmt"
 )
@@ -16,6 +16,12 @@ var RandGeneTests = []struct {
 	GC     float64
 }{
 	{"testingRandGene", 102, GCcontent},
+}
+
+func TestMain(m *testing.M) {
+	rand.Seed(1)
+	code := m.Run()
+	os.Exit(code)
 }
 
 func TestRandGene(t *testing.T) {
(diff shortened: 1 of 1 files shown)


craiglowe (approved):
Cool!  I had not known about the TestMain function.  I had thought the seed was one by default?  Is it random by default?

dasnellings:
You're right. I looked into this a bit more and the problem isn't that the seed was being randomized, its that the order in which tests are executed is not guaranteed. Below are some tests I did using `go test -shuffle on` to force randomize the order tests are run

Tests:
```
func TestA(t *testing.T) {
   fmt.Println("A:", rand.Int())
}

func TestB(t *testing.T) {
   fmt.Println("B:", rand.Int())
}

func TestC(t *testing.T) {
   fmt.Println("C:", rand.Int())
}
```

Test 1
```
A: 5577006791947779410
B: 8674665223082153551
C: 6129484611666145821
```

Test 2
```
B: 5577006791947779410
C: 8674665223082153551
A: 6129484611666145821
```

Test 3
```
A: 5577006791947779410
C: 8674665223082153551
B: 6129484611666145821
```

You can see the numbers being pulled are the same, but the tests pulling them vary. 

My initial push with TestMain does not fix this since it sets the seed before the tests begin. My most recent commit changes it so the seed is set to 1 at the start of each test function and this is the result.

```
C: 5577006791947779410
A: 5577006791947779410
B: 5577006791947779410
```

This should fix the issue

rimangan (approved):

craiglowe (approved):
Great work tracking this down!

cfauci (approved):

PR #616, opened 2022-02-04
vcfToFa

craiglowe:
This is a program that will edit a fasta file based on the mutations seen in a vcf file.  Currently it edits by using the alternate allele for all sites and only handles biallelic SNPs.  People will probably want additional options (e.g. edit based on an individual, edit more complex mutations), but this is the first step that I need now.


cmd/vcfToFa/vcfToFa.go
```diff
@@ -0,0 +1,73 @@
...
+		"vcfToFa - Use the variant data in the vcf to edit an input fasta of the reference.\n\n" +
+			"Usage:\n" +
+			"  vcfToFa [options] input.vcf input.fa output.fa\n\n" +
+			"Options:\n\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var expectedNumArgs int = 3
+	var useAlt *bool = flag.Bool("useAlt", false, "Retains only biallelic variants in the output file.")
```
rimangan, line 56: a bit confused by your vision for this program, should useAlt be the default behavior? Right now this program would spit out an identical output if this is not checked, right? Perhaps you want to make it flexible for multiple alt fields? Or can we remove this argument and clarify the usage?

fasta/fasta.go
```diff
@@ -103,6 +103,22 @@ func Write(filename string, records []Fasta) {
 	exception.PanicOnErr(file.Close())
 }
 
+// Write a fasta to input filename. Output fastas have line length of 50.
+func WriteMap(filename string, records FastaMap) {
```
rimangan, line 107: non deterministic output order from this function? Should FastaMap map chrom names to structs, including the sequence and the index?

cmd/vcfToFa/vcfToFa.go
```diff
@@ -0,0 +1,73 @@
...
+		if seqsLookup[v.Chr][v.Pos-1] != dna.StringToBase(v.Ref) {
+			log.Fatal("Error: base in fasta didn't match ref base from VCF record\n")
+		}
+
+		if useAlt {
+			seqsLookup[v.Chr][v.Pos-1] = dna.StringToBase(v.Alt[0])
+		}
+	}
+
+	fasta.WriteMap(outFaFilename, seqsLookup)
```
dasnellings, line 42: The fasta map just has pointers to the sequences in the slice so you could just use `fasta.Write(seqsOrdered)` and get the properly ordered output unless there is a specific reason you want to write from the map.

fasta/fasta.go
dasnellings: I think we should just stick with Write unless there is a specific reason to write from a map. I think it would be good practice to always keep around the slice of ordered fasta if you ever plan to write them to a file. I don't love the idea of mapping to structs.

cmd/vcfToFa/vcfToFa.go
rimangan: So we can still take advantage of map lookup while doing normal slice based writing! Such a solution

craiglowe: Yes, you are correct.  I had thought about doing that, but I was thinking that in the future people might expand to insertions, which will then have the potential to change where the DNA is stored in memory, if the space needs to be expanded.  I could leave this for that future person to figure out by putting a comment.  Another option would be to add an (optional) ordering to the WriteMap function, or some function that "updates" the slice by looping through slice, hitting the map, and setting the seq of the fasta equal to the seq of the map.  Do people have a preference?

cmd/vcfToFa/vcfToFa.go
craiglowe: My vision was that "useAlt" might be a somewhat rare case in the final version of this program, but you are correct that it is now confusing.  I will add a catch such that someone who does not use "useAlt" will get an error, rather than the unintuitive result of nothing happening.

fasta/fasta.go
craiglowe: To address these comments and others, I think I will write a function that copies seq back from a map back to a slice, maintaining the order of the slice, but updating with seq from the map.  Or maybe I will just make a note in the code that this would be needed for future extensions to handle insertions.

craiglowe (commented):
Response to comments

cmd/vcfToFa/vcfToFa.go
dasnellings: Good point. 

Personally I prefer the update option.

dasnellings (approved):

rimangan (approved):

dna/convert_test.go
```diff
@@ -36,6 +36,40 @@ func TestBasesToString(t *testing.T) {
...
+	{"g", LowerG},
+	{"t", LowerT},
+	{"n", LowerN},
+}
+
+func TestStringToBase(t *testing.T) {
+	for _, test := range equalStringAndBase {
+		actual := StringToBase(test.character)
+		if actual != test.base {
+			t.Errorf("StringToBase(%s): expected %d, actual %d", test.character, test.base, actual)
```
cshoben, line 59: I was wondering if there was a reason to print the "test.character" here in this error message instead of using the "actual" variable created on line 57.

cshoben (approved):

PR #617, opened 2022-02-04
IncludeRef in vcfAfs

rimangan:
When constructing derived allele frequency spectra (dAFS) from VCF variant data, the new option "includeRef"
will add the reference genome allele as one more observed sequence in the dAFS for each site. So if the reference genome has the derived allele in a position, the allele frequency at the segregating site will be (i+1) / (n+1), where i/n is the allele frequency without the reference position. This change impacts several programs, including selectionMcmc, selectionMle, and vcfAfs.

I also removed afsPmf, an unused program for making density distributions from dAFS data which I now conduct primarily in R. We have all the functionality we need with just vcfAfs, so I'm comfortable deleting this older program.


craiglowe (commented):
I had to wrestle with the code in VcfSampleToSegSite a little bit and write out some examples by hand, but it looks to work well!  Is it possible to add a few hand written tests of VcfSampleToSegSite to make sure we get the counts we expect with IncludeRef as true and false?  It looks like right now we don't test that function for correctness, only with downstream stability tests.

rimangan:
> I had to wrestle with the code in VcfSampleToSegSite a little bit and write out some examples by hand, but it looks to work well! Is it possible to add a few hand written tests of VcfSampleToSegSite to make sure we get the counts we expect with IncludeRef as true and false? It looks like right now we don't test that function for correctness, only with downstream stability tests.

Next commit adds in these tests.

craiglowe (approved):
Great work!

cshoben (approved):

Issue #434, opened 2021-03-19
dotToNewick is Broken


cfauci:
Last time i tried to run this command it didn't output a newick format that our parser could read

PR #618, opened 2022-02-04
dotToNewick Testing

rimangan:
Testing for dotToNewick, which draws Newick trees from DOT format trees.
https://en.wikipedia.org/wiki/Newick_format
https://en.wikipedia.org/wiki/DOT_(graph_description_language)

### DIFF
--- a/cmd/dotToNewick/dotToNewick.go
+++ b/cmd/dotToNewick/dotToNewick.go
@@ -10,9 +10,8 @@ import (
 	"log"
 )
 
-func dotToNewick(inFile string, outFile string) {
-	t := tree.ParseDot(inFile)
-
+func dotToNewick(inFile string, outFile string, verbose bool) {
+	t := tree.ParseDot(inFile, verbose)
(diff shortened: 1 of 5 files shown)


rimangan:
#434

rimangan:
This program outputs a newick tree that looks correct from what I can tell, but I notice that drawNewickTree is unable to parse it, like Christi noted in the issue she raised. However, drawNewickTree is also untested, so I don't know if it's the parser there or an issue here.

cfauci:
Parser on draw newick is good I've used it successfully a whole bunch and last I left it everything was working great. I was also desperately confused because the newicks out of dotToNewick looked right to me too.

rimangan:
> Parser on draw newick is good I've used it successfully a whole bunch and last I left it everything was working great. I was also desperately confused because the newicks out of dotToNewick looked right to me too.

Hmm good to know maybe we can take a look at this together next week.

cshoben (approved):

cfauci (approved):

dasnellings (approved):

PR #619, opened 2022-02-07
popgenTestingPart1

rimangan:
Testing for Dunn and MaximumLikelihood selection analysis. These tests are redundant with the tests in the corresponding cmds, mostly copied and pasted into the popgen package so we get better test coverage.


craiglowe (approved):
Great to have these tests!

RavenYL (approved):

PR #620, opened 2022-02-08
pileup memory leak fix

dasnellings:
I encountered a memory leak in the pileup code when a read started with an insertion (not sure why a mapper would do that). Because insertions are assigned to the base prior to the insertion during pileup, the data had already been written to file, and the code was endlessly expanding the pileup linked list to find an already passed base. This led to massive memory bloat (hundreds of GB before the program was killed). This PR fixes the particular issue and adds a test to catch a memory leak from this particular case.


dasnellings:
I also found a bug where any positions after the start of the last read were not output. This is also fixed in this PR.

rimangan (dismissed):

RavenYL (approved):

cshoben (approved):
Good catch!

rimangan (approved):

PR #624, opened 2022-02-11
Cigar pointers and function culling

dasnellings:
Removed pointers from cigar and culled functions where necessary. I also did a rewrite of sam consensus to get rid of some unique functions, but the tests all pass and the codes is a bit more simplified. Tests for cigar will be in my next PR.


rimangan (dismissed):

cmd/samConsensus/samConsensus.go
```diff
@@ -19,149 +18,99 @@ import (
...
-						SeqIndex++
-						RefIndex++
-					}
-				} else if aln.Cigar[i].Op != 'H' {
-					SeqIndex = SeqIndex + aln.Cigar[i].RunLength
-				}
-			}
-		}
+	for i := range ref {
+		dna.AllToLower(ref[i].Seq)
```
craiglowe, line 25: Should this be all to upper?  Is this because sam uses lowercase?

cmd/samConsensus/samConsensus.go
```diff
@@ -19,149 +18,99 @@ import (
...
 		outBase = dna.Gap
 	}
 
 	//now we check for ties
-	var maxCount int32 = 0
-
-	if v.A == max {
-		maxCount++
-		maxList = append(maxList, dna.A)
+	candidateBases := make([]int, 0, 5)
```
craiglowe, line 92: It is probably a rare case that you need to do this allocation.  One option would be to keep track in the calculations if you have a tie.  If you don't have a tie, then just return the max base, if you do have a tie then go through the following code to build the slice of bases that tie and choose a winner.
Above, tie would start out false.  If you get a tie you mark tie as true.  If you "beat" the current max, then you not only set max, but you also set tie to false.

cmd/samConsensus/samConsensus.go
```diff
@@ -19,149 +18,99 @@ import (
...
-	return maxList[numbers.RandIntInRange(0, len(maxList))]
-}
-
-type voteBase struct {
-	A   int32
-	C   int32
-	G   int32
-	T   int32
-	Gap int32
+	return dna.Base(candidateBases[numbers.RandIntInRange(0, len(candidateBases))])
```
craiglowe, line 113: Why not have candidate bases as []dna.Base instead of int?  I see us turning the bases into ints and then turning them back into bases when they come out.

craiglowe (approved):
Great work!

cmd/samConsensus/samConsensus.go
dasnellings: True. fixed.

rimangan (approved):

cshoben (approved):

RavenYL (approved):

PR #626, opened 2022-02-11
Raven maf coverage

RavenYL:
Added some coverage to the maf package
Added functions to compare if maf structs are equal
Overall coverage might have decreased


maf/compare.go
```diff
@@ -1,10 +1,105 @@
 package maf
 
 import (
+	"github.com/vertgenlab/gonomics/dna"
 	"sort"
 	"strings"
 )
 
+func MafSLinesAreEqual(a MafSLine, b MafSLine) bool {
+	if strings.Compare(a.Src, b.Src) != 0 {
```
craiglowe, line 10: This can also be "a.Src != b.Src"

maf/compare.go
```diff
@@ -1,10 +1,105 @@
...
+		return false
+	}
+	if dna.CompareSeqsCaseSensitive(a.Seq, b.Seq) != 0 {
+		return false
+	}
+	return true
+}
+
+func MafILinesAreEqual(a MafILine, b MafILine) bool {
+	if strings.Compare(a.Src, b.Src) != 0 {
```
craiglowe, line 32: Like above, this can use "!="

maf/compare.go
```diff
@@ -1,10 +1,105 @@
...
+		return false
+	}
+	if a.RightCount != b.RightCount {
+		return false
+	}
+	return true
+}
+
+func MafELinesAreEqual(a MafELine, b MafELine) bool {
+	if strings.Compare(a.Src, b.Src) != 0 {
```
craiglowe, line 51: This can also use != between the strings

maf/compare.go
```diff
@@ -1,10 +1,105 @@
...
+		return false
+	}
+	if a.Status != b.Status {
+		return false
+	}
+	return true
+}
+
+func MafSpeciesAreEqual(a MafSpecies, b MafSpecies) bool {
+	if strings.Compare(a.Src, b.Src) != 0 {
```
craiglowe, line 73: !=

craiglowe (dismissed):
Wonderful work!  There are a few places where we could use the simpler "!=" between strings instead of strings.Compare

maf/compare.go
```diff
@@ -1,10 +1,105 @@
 package maf
 
 import (
+	"github.com/vertgenlab/gonomics/dna"
 	"sort"
 	"strings"
 )
 
+func MafSLinesAreEqual(a MafSLine, b MafSLine) bool {
```
rimangan, line 9: A number of these compare functions may be able to be made private

rimangan (approved):
LGTM!

craiglowe (approved):
Wonderful work!  Great to have these tests!  Riley had a good idea that some of these can likely become private functions (starting with a lowercase letter in the name of the function and therefore not accessible outside of the maf package), but I am also happy to merge all of this improvements now and have further improvements as a later pull request on a later day.

dasnellings (approved):
Great to have tests. I agree that most of these should probably be private since they are only used for testing at the moment.

PR #627, opened 2022-02-12
simRecon cleanup

cfauci:
Added testing for simRecon and removed extra files from testdata directory


cmd/simRecon/simRecon_test.go
```diff
@@ -0,0 +1,55 @@
...
+	SimRecon(root, newick, gene, simOut, leafOut, reconOut, accOut, baseAcc)
+	simOk := fasta.AllAreEqualIgnoreOrder(fasta.Read(simOut), fasta.Read("testdata/simOutT.fasta"))
+	leafOk := fasta.AllAreEqualIgnoreOrder(fasta.Read(leafOut), fasta.Read("testdata/leafOutT.fasta"))
+	reconOk := fasta.AllAreEqualIgnoreOrder(fasta.Read(reconOut), fasta.Read("testdata/reconOutT.fasta"))
+	accOk := accEqual(accOut, "testdata/accOutT.txt")
+	baseAccOk := accEqual(baseAcc, "testdata/baseAccOutT.txt")
+	if simOk && leafOk && reconOk && accOk && baseAccOk != true {
+		t.Errorf("one or more testing files did not match expected values, SimMatch: %v, LeafMatch: %v, ReconMatch: %v, AccuracyMatch: %v, BaseAccuracyMatch: %v", simOk, leafOk, reconOk, accOk, baseAccOk)
+	}
+	fileio.EasyRemove("testdata/simOut.fasta")
```
rimangan, line 50: An optional thing, but one good practice is to put the remove statements in an "else" so if the test fails the output files stick around so you can quickly debug.

rimangan (dismissed):
Looks great!

cmd/simRecon/simRecon_test.go
```diff
@@ -0,0 +1,55 @@
+package main
+
+import (
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/fileio"
+	"strings"
+	"testing"
+)
+
+func accEqual(a string, b string) bool {
```
craiglowe, line 10: I am not sure the logic of this function is what you are looking for.  If I am reading it correctly, it takes two file names and if there is at least one line in file A, for which there is also a line in file B that has the first two columns match then we return true.
I am wondering if you might be looking for all the lines to match, but want to allow a different ordering of the lines?

craiglowe (commented):
Great to see work on upgrading the testing in this package!  I had some questions about a comparison function, which I think could use a little bit of documentation or a modification of the logic.

cmd/simRecon/simRecon_test.go
cfauci: That's a good idea, I'll update it.

cmd/simRecon/simRecon_test.go
cfauci: Yeah I got hung up on checking all the names first but now that you point it out it isn't totally necessary. Instead of splitting on the tab should I just leave the line as a single "word" and check them against each other?

rimangan (approved):

dasnellings (approved):

cshoben (approved):

PR #632, opened 2022-02-14
Vcf to table

dasnellings:
This PR adds a `-csv` flag to vcfFormat to output vcf records for spreadsheet-based analyses. It turns out to be a fair bit more complicated than it initially seems, but it is working. There are tests, though there are so many corner cases in VCF right now that there are almost certainly some files where this will break.


cmd/vcfFormat/vcfFormat.go
```diff
@@ -55,18 +72,20 @@ func usage() {
...
-			"vcfFormat input.vcf output.vcf\n" +
+			"vcfFormat input.vcf\n" +
 			"options:\n")
 	flag.PrintDefaults()
 }
 
 func main() {
-	var expectedNumArgs int = 2
+	var expectedNumArgs int = 1
+	var outfile *string = flag.String("o", "stdout", "Output file")
```
rimangan, line 82: I'm not sure I love the default stdout in most cases and would be curious to hear people's thoughts at code review. Certainly inconsistent with the style of much of the code base.

cmd/vcfFormat/vcfFormat_test.go
```diff
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"encoding/csv"
```
rimangan, line 4: not sure if i've seen the csv package in gonomics, nice!

craiglowe (approved):
This will be nice functionality for people who want to look at VCF data, but are not programmers.

vcf/queryInfo.go
```diff
@@ -81,6 +81,10 @@ func parseValue(v Vcf, s []string, k Key) interface{} {
 		for i := range s {
 			stringValues = strings.Split(s[i], ",")
 			for j := range stringValues {
+				if stringValues[j] == "." {
```
cshoben, line 84: Can you point me in the direction of where I can see an example of this input s []string?

cmd/vcfFormat/table.go
```diff
@@ -0,0 +1,231 @@
...
+	_, err := fmt.Fprintln(outfile, s.String())
+	exception.PanicOnErr(err)
+}
+
+// getMaxAltCount reads through the input vcf file to determine the maximum number of alternate alleles present
+func getMaxAltCount(infile string) int {
+	var maxAlts int
+	records, _ := vcf.GoReadToChan(infile)
+	for v := range records {
+		if len(v.Alt) > maxAlts {
```
cshoben, line 106: How does this work if the alt is a structural variant? Where the length might be greater but it is truly only one alt? I'm wondering if len() is a good way to count alt or if parsing on the "," and counting is more accurate?

cmd/vcfFormat/table.go
```diff
@@ -0,0 +1,231 @@
...
+	case "A": // == num alt alleles
+		return maxAlts
+
+	case "R": // == num ref + alt alleles
+		return maxAlts + 1
+
+	case "G": // one value for each possible genotype
+		return 1 //TODO once parser is updated
+
+	case ".": // wildcard. they never make it easy do they...
```
cshoben, line 124: lol

cshoben (approved):
Just a few questions but nothing to hold up approval

rimangan (approved):

rimangan (changes requested):
Fix -o flag

craiglowe (approved):
Great!

rimangan (approved):

PR #634, opened 2022-02-15
Pileup update

dasnellings:
Going through results from the variant caller I noticed that I was getting calls where all of the alt reads are only present on a single strand. This is a common exclusion criteria for evaluating variants. In preparation to adding strand detection to the variant caller, this PR modifies the Pileup code to save information about read strand. I also went ahead and added support for multi-base deletions. It is a bit messy as the multi-base deletions do not contribute to the total read counts (that is handled by dna.Gap) but I think it is important functionality to have. 

This PR does not make any major changes to the variant caller. That will come in a future PR dependent on this one.

### DIFF
--- a/cmd/pileup/pileup.go
+++ b/cmd/pileup/pileup.go
@@ -92,11 +92,11 @@ func fmtOutput(pile sam.Pile, header sam.Header, builder *strings.Builder) strin
 	builder.WriteString(fmt.Sprintf("%s\t%d\t%d\t%d\t%d\t%d\t%d\t%d", header.Chroms[pile.RefIdx].Name, pile.Pos,
 		pile.CountF[dna.A]+pile.CountR[dna.A], pile.CountF[dna.C]+pile.CountR[dna.C], pile.CountF[dna.G]+pile.CountR[dna.G],
 		pile.CountF[dna.T]+pile.CountR[dna.T], pile.CountF[dna.N]+pile.CountR[dna.N], pile.CountF[dna.Gap]+pile.CountR[dna.Gap]))
-	for seq, _ := range pile.InsCountF {
+	for seq := range pile.InsCountF {
 		builder.WriteString(fmt.Sprintf("\t%s:%d", seq, pile.InsCountF[seq]+pile.InsCountR[seq]))
 	}
 	var presentInForward bool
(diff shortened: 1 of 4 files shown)


rimangan (approved):

rimangan:
At code review i'd love to hear a bit about the technical reasons why you can get all the reads on a single strand, cool consideration, and well implemented here!

cmd/pileup/pileup.go
```diff
@@ -87,9 +90,17 @@ func pileup(infile string, outfile string, settings Settings) {
 func fmtOutput(pile sam.Pile, header sam.Header, builder *strings.Builder) string {
 	builder.Reset()
 	builder.WriteString(fmt.Sprintf("%s\t%d\t%d\t%d\t%d\t%d\t%d\t%d", header.Chroms[pile.RefIdx].Name, pile.Pos,
-		pile.Count[dna.A], pile.Count[dna.C], pile.Count[dna.G], pile.Count[dna.T], pile.Count[dna.N], pile.Count[dna.Gap]))
-	for seq, count := range pile.InsCount {
-		builder.WriteString(fmt.Sprintf("\t%s:%d", seq, count))
+		pile.CountF[dna.A]+pile.CountR[dna.A], pile.CountF[dna.C]+pile.CountR[dna.C], pile.CountF[dna.G]+pile.CountR[dna.G],
+		pile.CountF[dna.T]+pile.CountR[dna.T], pile.CountF[dna.N]+pile.CountR[dna.N], pile.CountF[dna.Gap]+pile.CountR[dna.Gap]))
+	for seq := range pile.InsCountF {
+		builder.WriteString(fmt.Sprintf("\t%s:%d", seq, pile.InsCountF[seq]+pile.InsCountR[seq]))
```
craiglowe, line 96: Is the trick here that sometimes pile.InsCountR[seq] will miss in the map, but that during those misses we get 0 as the default value?

craiglowe (approved):
Great to have this additional functionality!

cmd/pileup/pileup.go
dasnellings: Yes, that's right. That is why there is a loop on line 99 to get anything we missed.

cshoben (approved):

PR #635, opened 2022-02-16
bedMerge lowmem

rimangan:
Low memory version of bedMerge using channels. bedMerge requires sorted input, so bedMerge on main
reads in the entire bed file, sorts it, and then performs the bedMerge operation. With the "lowmem" user option,
we can read beds with a channel and perform bedMerge without the built in sorting step, enabling the analysis of large bed files. Critically, this means input bed files must be presorted to be compatible with this option, as is indicated in the usage.

### DIFF
--- a/cmd/bedMerge/bedMerge.go
+++ b/cmd/bedMerge/bedMerge.go
@@ -6,11 +6,50 @@ import (
 	"flag"
 	"fmt"
 	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/exception"
+	"github.com/vertgenlab/gonomics/fileio"
 	"github.com/vertgenlab/gonomics/numbers"
 	"log"
 )
 
-func bedMerge(infile string, outfile string, mergeAdjacent bool) {
+func bedMerge(infile string, outfile string, mergeAdjacent bool, lowMem bool) {
+	if lowMem {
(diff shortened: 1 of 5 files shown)


dasnellings (approved):

cmd/bedMerge/bedMerge.go
```diff
@@ -6,11 +6,50 @@ import (
...
 	"github.com/vertgenlab/gonomics/numbers"
 	"log"
 )
 
-func bedMerge(infile string, outfile string, mergeAdjacent bool) {
+func bedMerge(infile string, outfile string, mergeAdjacent bool, lowMem bool) {
+	if lowMem {
+		bedMergeLowMem(infile, outfile, mergeAdjacent)
+	} else {
+		bedMergeHighMem(infile, outfile, mergeAdjacent)
```
craiglowe, line 19: Is there an advantage to the highmem version?  Should the lowmem version replace the highmem version?

craiglowe: I see now that highmem does not require the input to be sorted.  An idea for sometime in the future would be to add a "unSorted" flag where the program could read in the bed, sort it, and then send that slice to a channel that would run the lowmem code.  This might reduce our redundant code.

craiglowe (approved):
This is a nice addition!

cfauci (approved):

PR #637, opened 2022-02-18
Variant Caller Upgrades

dasnellings:
This PR introduces several upgrades to the variant caller and is dependent on my pileup PR below. I added several flags to the caller to filter calls by allele frequency, mapping quality, alternate read depth, and strand bias. I have also added support for multi-base deletions. This also has made me feel that we need a general solution for generating vcf records rather than having each application handle it independently as we currently do. There are many considerations for writing vcfs, especially regarding sites with multiple alleles and it is hard to get right (as you can see from my commit history iteratively fixing bugs).


cmd/callVariants/callVariants.go
```diff
@@ -27,15 +27,15 @@ func usage() {
 	flag.PrintDefaults()
 }
 
-func callVariants(experimentalFiles, normalFiles []string, refFile, outFile string, maxP, minAf float64, minCoverage, workers int) {
+func callVariants(experimentalFiles, normalFiles []string, refFile, outFile string, maxP, minAf, maxAf, maxStrandBias float64, minCoverage, minMapQ, minAltReads, workers int) {
```
rimangan, line 30: perhaps enough options for a Settings struct, especially if this cmd is likely to grow in complexity

rimangan (approved):

cmd/callVariants/callVariants.go
dasnellings: I thought about it, but they are being used in various different places downstream, so I opted to not use a struct.

cfauci (approved):

craiglowe (approved):
This is great!

rimangan (approved):

PR #638, opened 2022-02-20
vcfToFa MultiFa Compatibility

rimangan:
vcfToFa edits an input fasta file based on information from an input vcf. Currently, it can change the base in the reference genome to the "alt" states defined in the vcf (though a useful program from the future would be to make 'individual' sequences from the Sample fields). This PR adds a "multiFaMode", in which a multiFa can be provided instead of a normal reference fasta, and VCF alt states can be substituted in the correct reference positions, with respect to gaps.
Passes the tests I wrote, will make it a real PR once I can test on genome-wide data and we discuss the logic at code review.

Critically, there was a small bug in RefPosToAlnPosCounter, but it seems that none of our tests covered the rare case where it matters (consecutive vcf records with a gap between them).


cmd/vcfFilter/vcfFilter.go
```diff
@@ -16,11 +17,34 @@ import (
...
+	var header vcf.Header
+	var err error
+	var mapContains bool
+	var sitesSeen map[string]map[int]uint8 = make(map[string]map[int]uint8, 0) //uint8 is the number of times this site is seen in the vcf file.
+
+	if c.biAllelicOnly {
+		records, _ = vcf.GoReadToChan(infile)
+		for v := range records {
+			if _, mapContains = sitesSeen[v.Chr]; mapContains { //if an entry in the map for the current chrom exists
+				if _, mapContains = sitesSeen[v.Chr][v.Pos]; mapContains {
```
dasnellings, line 30: maps are pretty smart when it comes to memory allocation. If you check a map for a key that is not present it will return the zero value for the relevant type so you don't need to worry about checking if it is present then allocating (at least for data types without pointers). So you could replace lines 30-34 with just `sitesSeen[v.Chr][v.Pos]++`

cmd/vcfFilter/vcfFilter.go
```diff
@@ -16,11 +17,34 @@ import (
 
 func vcfFilter(infile string, outfile string, c criteria, groupFile string, parseFormat bool, parseInfo bool, setSeed int64) (total, removed int) {
 	rand.Seed(setSeed)
-	records, header := vcf.GoReadToChan(infile)
+	var records <-chan vcf.Vcf
+	var header vcf.Header
+	var err error
+	var mapContains bool
+	var sitesSeen map[string]map[int]uint8 = make(map[string]map[int]uint8, 0) //uint8 is the number of times this site is seen in the vcf file.
```
dasnellings, line 24: Not sure performance is a top priority here, but you might consider making a single map instead of a map of maps. You could wrap `v.Chr` and `v.Pos` in a struct and use that as the map key. 1 lookup is always faster than 2.

cmd/vcfFilter/vcfFilter.go
```diff
@@ -46,6 +69,15 @@ func vcfFilter(infile string, outfile string, c criteria, groupFile string, pars
 			v = vcf.ParseFormat(v, header)
 		}
 
+		if c.biAllelicOnly {
+			if sitesSeen[v.Chr][v.Pos] < 1 {
```
dasnellings, line 73: In my sisyphean struggle to convince people that `switch` > `else if` I will point out that in addition to its obvious stylistic superiority, here it would also have the added benefit of only 2 lookups instead of 4.

```
switch sitesSeen[v.Chr][v.Pos] {
     case 0:  // Panic
     case 1:  // continue
     default: // remove site
}
```

dasnellings (approved):

cmd/vcfToFa/vcfToFa.go
```diff
@@ -11,31 +11,76 @@ import (
...
+
 		if !(vcf.IsBiallelic(v) && vcf.IsSubstitution(v)) {
 			log.Fatal("Error: currently we only handle biallelic substitutions\n")
 		}
 
-		// check to make sure the fa matches the vcf before we edit
-		if seqsLookup[v.Chr][v.Pos-1] != dna.StringToBase(v.Ref) {
-			log.Fatal("Error: base in fasta didn't match ref base from VCF record\n")
-		}
+		if multiFaMode {
```
craiglowe, line 60: I think what happened here is that there are two separate functions that have been combined into one function, but they only share about 4 lines of code and the other 100 lines are wrapped in "if" statements where we do one thing if we are in multiFa mode and another set of code if we are not.  One option that I think will be easier to read/review/maintain would be to add a separate function that does mutliFa mode.  This will not be duplicate code since the combined version does not really share code.
The other option would be to narrow the scope of this pull request to "improve" the code to allow gaps in the input sequence.
I think there are two things being added here: 1) we are allowing the reference sequence to have gaps in it.  The code that allows for gaps should also be fine for the case where we have no gaps, so I think we can just "improve" the current code rather than needing a new function 2) We are adding the ability to add a sequence to the original sequences in the file instead of needing to "cat" the input and output files.  In this case I think it would be better to just run "cat" on the input and output after running the program.  The workflow would be to faFilter the human sequence out of the multiFasta.  Run vcfToFa on the human fasta, which has gaps to edit it.  Then run "cat" on the multiFasta and the output to have the original alignment with the edited sequence added.
I worry that what could be a few commands are being put into one command, which could cause this command to be more difficult to a review/maintain.

craiglowe (commented):
Great to have this functionality!  I included a few high-level comments

cmd/vcfToFa/vcfToFa.go
rimangan: i went the first route in the next commit, split this function into two. However, both are still housed in the same cmd.

dasnellings (approved):

RavenYL (approved):

cshoben (approved):

cfauci (approved):

dasnellings (approved):

PR #640, opened 2022-02-23
rm Interval interface write dependency

dasnellings:
The `interval.Interval` interface requires the method `WriteToFileHandle(io.Writer)` but this method is not used in the interval package, only in downstream utilities like sequelOverlap and lift. I wanted to remove this dependency so interval can be used on abstract data types that do not have a file representation. This PR removes the method from interval and adds the `fileWriter` interface in sequelOverlap and lift that requires the method. When it is time to write the file, the cmd does a type assertion from `Interval` to `fileWriter` and does the write. I was worried about the performance impact of type assertions, but it seems to be negligible. 

```
goos: darwin
goarch: amd64
pkg: github.com/vertgenlab/gonomics/cmd/sequelOverlap
cpu: VirtualApple @ 2.50GHz
BenchmarkAssertWrite-8           1000000              1112 ns/op             112 B/op          2 allocs/op
BenchmarkAllocWrite-8            1000000              1082 ns/op             112 B/op          2 allocs/op

```


rimangan (approved):

RavenYL (approved):
Now that the interval.Interval interface does not require writing to file, does that open it up to be used in more programs? What would that look like?

cshoben (approved):

cfauci (approved):

PR #641, opened 2022-03-06
overlapEnrichment, lift package, and method testing

rimangan:
1. bedEnrichments has been renamed overlapEnrichments, and accepts Lift interface-compatible file types. Now we can do enrichments between beds and vcfs, beds and beds, or vcfs and vcfs.
2. the cmd lift has been renamed liftCoordinates, do distinguish from the lift interface, and has been edited to reflect changes in the lift interface.
3. The Lift interface, used for overlapEnrichments and liftCoordinates, has been elevated to its own package due to sufficient complexity.
4. Some additional testing coverage has been provided for Lift-related methods.


interval/relationship.go
```diff
@@ -126,3 +128,32 @@ func PrintRelationships() {
 	fmt.Println("\"end\"    = f + fi + e")
 	fmt.Println("\"equal\"  = e")
 }
+
+//matchOverlapLen returns the number of bases shared between two start and end points. Used in MatchProportion.
+func matchOverlapLen(start1 int, end1 int, start2 int, end2 int) int {
+	return numbers.Max(0, numbers.Min(end1, end2)-numbers.Max(start1, start2))
+}
+
+//MatchProportion returns the proportion of bases in the target and query that can be lifted for a particular interval as a pair of floats (propT, propQ)
```
dasnellings, line 137: Should this be in the new lift package?

bed/methods.go
```diff
@@ -14,17 +14,18 @@ func (b Bed) GetChrom() string {
...
+	return b.ChromStart
 }
 
 func (b Bed) GetChromEnd() int {
-	return int(b.ChromEnd)
+	return b.ChromEnd
 }
 
-func (b Bed) UpdateLift(c string, start int, end int) {
+func (b Bed) UpdateCoord(c string, start int, end int) interface{} {
```
dasnellings, line 24: This is totally fine, though I imagine the empty interface return can get a bit unwieldy. This is not a necessary change, but you might consider instead something like `func (b *Bed) SetCoord(...)`. That way you don't need a return value. I'm not positive, but I think pointer here would go on the stack, not the heap, so the performance would be pretty much identical between the two options.

dasnellings (approved):

bed/methods.go
rimangan: i was going to ask you about this, because i tried that first and then it said Bed could not implement Lift because it seemed like it couldn't recognize the function (b *Bed) as taking in a Bed. How did you get around this?

dasnellings: That is a good point. I did not think of that when I commented. In looking a bit deeper it seems that the general consensus is to use pointers for all methods in a set if any of them require pointers. This may be be worth a larger conversation. I'll copy below an excerpt from the FAQ on go.dev

Should I define methods on values or pointers?
func (s *MyStruct) pointerMethod() { } // method on pointer
func (s MyStruct)  valueMethod()   { } // method on value
For programmers unaccustomed to pointers, the distinction between these two examples can be confusing, but the situation is actually very simple. When defining a method on a type, the receiver (s in the above examples) behaves exactly as if it were an argument to the method. Whether to define the receiver as a value or as a pointer is the same question, then, as whether a function argument should be a value or a pointer. There are several considerations.

First, and most important, does the method need to modify the receiver? If it does, the receiver must be a pointer. (Slices and maps act as references, so their story is a little more subtle, but for instance to change the length of a slice in a method the receiver must still be a pointer.) In the examples above, if pointerMethod modifies the fields of s, the caller will see those changes, but valueMethod is called with a copy of the caller's argument (that's the definition of passing a value), so changes it makes will be invisible to the caller.

By the way, in Java method receivers are always pointers, although their pointer nature is somewhat disguised (and there is a proposal to add value receivers to the language). It is the value receivers in Go that are unusual.

Second is the consideration of efficiency. If the receiver is large, a big struct for instance, it will be much cheaper to use a pointer receiver.

Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used. See the section on [method sets](https://go.dev/doc/faq#different_method_sets) for details.

For types such as basic types, slices, and small structs, a value receiver is very cheap so unless the semantics of the method requires a pointer, a value receiver is efficient and clear.

rimangan: yes certainly looks like a longer conversation. I'm open to switching to pointers, though it would be quite a rewrite for interval and other packages

RavenYL (approved):

cmd/liftCoordinates/liftCoordinates.go
```diff
@@ -138,12 +146,12 @@ func minMatchPass(overlap *chain.Chain, i interval.Interval, minMatch float64) b
 
 func usage() {
 	fmt.Print(
-		"lift - Moves an interval interface compatable file format between assemblies.\n" +
+		"liftCoordinates - Lifts a Lift interface compatable file format between assembly coordinates.\n" +
```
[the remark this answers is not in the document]
rimangan: the lift ReadToChan function will fatal if the filename is neither .bed, .bed.gz, .vcf, or .vcf.gz. Maybe a good idea to add that it's bed and vcf for the usage here

interval/relationship.go
rimangan: Only uses intervals, could go either way

craiglowe: It does seem like this should be in lift, but maybe that causes a circular dependency to have interval in lift (is lift already in interval?)?

craiglowe (approved):
Great to see this getting added to main!

PR #642, opened 2022-03-07
vcf samples to multiAln

craiglowe:
This adds functionality so that we can use the samples present in a vcf to add on to a fasta multiple alignment

### DIFF
--- a/cmd/vcfToFa/vcfToFa.go
+++ b/cmd/vcfToFa/vcfToFa.go
@@ -95,7 +95,7 @@ func vcfToMultiFa(inVcfFilename string, inFaFilename string, outFaFilename strin
 		}
 
 		if !(vcf.IsBiallelic(v) && vcf.IsSubstitution(v)) {
-			log.Fatal("Error: currently we only handle biallelic substitutions\n")
+			log.Fatal("Error: currently we only handle biallelic substitutions.  You can filter the vcf for these with vcfFilter.\n")
 		}
 		if v.Chr == chromName { //only consider variants with the correct chrom name.
 			currAlnPos = fasta.RefPosToAlnPosCounter(seqsOrdered[outIndex], v.Pos-1, prevRefPos, prevAlnPos)
@@ -154,7 +154,7 @@ func vcfToFa(inVcfFilename string, inFaFilename string, outFaFilename string, us
 
 func usage() {
 	fmt.Print(
-		"vcfToFa - Use the variant data in the vcf to edit an input fasta of the reference.\n\n" +
+		"vcfToFa - Use the variant data in the vcf to edit an input fasta of the reference.  Assumes diploid organisms unless the haploid flag is given..\n\n" +
 			"Usage:\n" +
 			"  vcfToFa [options] input.vcf input.fa output.fa\n\n" +
 			"Options:\n\n")
--- a/cmd/vcfToFa/vcfToFa.go
+++ b/cmd/vcfToFa/vcfToFa.go
@@ -11,36 +11,80 @@ import (
 	"log"
 )
 
-func vcfToMultiFa(inVcfFilename string, inFaFilename string, outFaFilename string, chromName string, useAlt bool) {
+// addAnotherCopyOfRef takes an alignment and a name.  The function will copy the reference (first sequence
+//  in the alignment and place this copy at the end of the alignment with the name that is given.
+func addAnotherCopyOfRef(aln []fasta.Fasta, nameOfCopy string) []fasta.Fasta {
+	var refSeqCopy []dna.Base
+	if len(aln) == 0 {
+		log.Fatalf("Input fasta must contain at least one record in multiFaMode.")
+	}
+	refSeqCopy = make([]dna.Base, len(aln[0].Seq))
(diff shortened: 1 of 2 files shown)


cmd/vcfToFa/vcfToFa.go
```diff
@@ -117,11 +163,13 @@ func usage() {
...
 func main() {
 	var expectedNumArgs int = 3
-	var useAlt *bool = flag.Bool("useAlt", false, "Retains only biallelic variants in the output file.")
+	var useAlt *bool = flag.Bool("useAlt", false, "Creates a new sequence using the alternative base at every site (only uses biallelic SNPs).")
+	var useSamples *bool = flag.Bool("useSamples", false, "Creates new sequences using the genotype of that individual (only uses biallic SNPs).")
 	var multiFaMode *bool = flag.Bool("multiFaMode", false, "Use a multiFa reference chromosome. Variants will be placed in reference"+
 		"positions with respect to gaps. Input vcf must be sorted when using multiFaMode. Appends a new sequence with the updated sequence from the reference sequence,"+
 		"defined as the first sequence in the multiFa. Currently only supports useAlt.")
 	var multiFaChromName *string = flag.String("multiFaChromName", "", "Identifies the chrom of the multiFa file. Only variants matching this chrom name will be used.")
+	var haploid *bool = flag.Bool("haploid", false, "Only make one sequence per individual in the vcf, instead of two.  The first genotype will be used.")
```
RavenYL, line 172: Is it worth specifying that only diploid and haploid are accepted? If I remember correctly, at some point Craig was helping someone use gonomics to parse vcfs of organisms with other ploidies.
And what is "the first genotype"? I thought if a sample is haploid there would only be 1 entry in its genotype column?

cmd/vcfToFa/vcfToFa.go
```diff
@@ -53,16 +97,18 @@ func vcfToMultiFa(inVcfFilename string, inFaFilename string, outFaFilename strin
 		if !(vcf.IsBiallelic(v) && vcf.IsSubstitution(v)) {
```
RavenYL, line 97: Does this mean that if any vcf record contains more than 2 alleles, e.g. ref=A, alt=AT, the whole program would log fatal? Is there a situation where it would be easier for the user if the program had given a warning instead, e.g. if they  have more than 2 alleles somewhere in their vcf, but still wanted to run this program without specifically removing those records?

RavenYL (approved):

cmd/vcfToFa/vcfToFa.go
rimangan: In the current PR, yes, we only handle haploid or diploid samples. For other polyploid samples, we'll need an additional argument that parses samples in a for loop for an arbitrary number of alleles. This is sort of a quicker fix for some specific short term use cases

cmd/vcfToFa/vcfToFa.go
rimangan: Yes. one could simply remove the offending records with vcfFilter and run this program again. I was initially worried about continuing because it could produce some silent errors, the warning might be a viable compromise.

rimangan (dismissed):

cmd/vcfToFa/vcfToFa.go
craiglowe: I am always nervous to have the user tell us to do something and us to just silently not do it.  The warning is a good idea, but I am worried that the user may not see it with cluster jobs.  I have updated the error message to let the user know they can use vcfFilter to update their vcf file (if they are ok with only biallelic substitutions.

cmd/vcfToFa/vcfToFa.go
craiglowe: I have added a note to the usage statement that we are assuming diploid, unless the haploid flag is given.

craiglowe (commented):
Added responses to comments.

cmd/vcfToFa/vcfToFa_test.go
```diff
@@ -13,21 +13,23 @@ var VcfToFaTests = []struct {
...
 	useAlt             bool
+	useSamples         bool
+	haploid            bool
 	multiFaMode        bool
 	multiFaChromName   string
 }{
 	//{"testdata/testInput.vcf", "testdata/testInput.fa", "testdata/actual.fa", "testdata/expected.fa", true, false, ""},
 	//{"testdata/testInput.vcf", "testdata/testInput.fa", "testdata/actualNoAlt.fa", "testdata/testInput.fa", false, false, ""},//should just produce the input file.
 	//{"testdata/testMultiInput.vcf", "testdata/testMultiInput.fa", "testdata/actualMultiNoAlt.fa", "testdata/expectedMultiNoAlt.fa", false, true, "chr1"},//should also produce the input file
 	//{"testdata/testMultiInput.vcf", "testdata/testMultiInput.fa", "testdata/actualMultiAlt.fa", "testdata/expectedMultiAlt.fa", true, true, "chr1"},
```
rimangan, line 24: uncomment

cshoben (approved):

rimangan (approved):

RavenYL (approved):

PR #643, opened 2022-03-11
Bed methods testing coverage

cshoben:
I've started to write a test file for bed/methods.go

Currently, my test for the bed UpdateLift() method is not working. My understanding is that this method changes the receiving bed but I might be mistaken. Looking for advice on resolving this. 

Also added some comments for the go docs.


bed/methods_test.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, v := range BedStructTests {
+		if v.chromEnd != v.bed.GetChromEnd() {
+			t.Errorf("Error in bed methods.go GetChromEnd(); Expected chromEnd: %v. Actual chromEnd: %v.", v.chromEnd, v.bed.GetChromEnd())
+		}
+	}
+}
+
+func TestBed_UpdateLift(t *testing.T) {
+	for _, v := range BedStructTests {
+		v.bed.UpdateLift("chr4", 3, 4)
```
RavenYL, line 44: Maybe you need to create another variable v_updated, and assign v_updated = v.bed.UpdateLift("chr4", 3, 4)?
I found this page that's Python but kind of relevant: https://stackoverflow.com/questions/53249829/python-keep-changes-on-a-variable-made-within-a-function
I was wondering how permanent the function output is. Maybe when you did v.bed.UpdateLift it worked but just didn't save

cshoben:
These changes were moved over to a new pull request to accommodate changes that Riley made. His changes were already merged in.

PR #647, opened 2022-03-14
Raven align anchor2

RavenYL:
Fixed bug and added features
- output bed for the gap regions that do not get aligned (due to large size or invalid start/end)
- output each alignment into bed
- fixed bug in align package. The runlength of the last cigar was assumed to be 1, but should be calculated


align/constGap.go
```diff
@@ -277,14 +277,29 @@ func writeCigar(trace [][]ColType, i_inChecker_max_Previous int, j_inChecker_max
...
-//inputs: the growing route describing the collection of cigars of the entire alignment except the last cigar, the index of the cigar that is currently being built, the ColType (M=0,I=1,D=2) of the last cigar entry
+//inputs: lengths of original sequences alpha and beta, the growing route describing the collection of cigars of the entire alignment except the last cigar, the index of the cigar that is currently being built, the ColType (M=0,I=1,D=2) of the last cigar entry
 //outputs: the updated final route describing the collection of cigars of the entire alignment including the last cigar, the updated and index of the cigar that is currently being built
-func lastCigar(route []Cigar, routeIdx_current int, Op_end ColType) ([]Cigar, int) {
+func lastCigar(len_alpha int, len_beta int, route []Cigar, routeIdx_current int, Op_end ColType) ([]Cigar, int) {
+	//Calculate the size of final runlength
+	var TotalRunLength, LastRunLength int64
+	TotalRunLength = 0
+	for routeIdx := range route {
+		TotalRunLength += route[routeIdx].RunLength
```
craiglowe, line 287: I am not sure we want to count insertions, deletions, and matches here.  I am wondering if (depending on whether we use len_beta or len_alpha) we want to either count only insertions+matches, or deletions+matches

align/view_test.go
```diff
@@ -19,6 +19,7 @@ var alignTests = []struct {
 	{"ACT", "ACGT", "AC-T\nACGT\n"},
 	{"CGCGCGCGCG", "CGCGCGTTTTCGCG", "CGCGCG----CGCG\nCGCGCGTTTTCGCG\n"},
 	{"CGCGCGCGCG", "CGAAAACGCGTTTTCGCG", "CG----CGCG----CGCG\nCGAAAACGCGTTTTCGCG\n"},
+	{"AA", "GGGAATT", "---AA--\nGGGAATT\n"},
```
craiglowe, line 22: This relates to my comment above, but I think we might want a test that has both insertions and deletions in the final alignment.

craiglowe (commented):
Great to see these improvements and bug fix!  There was one place where I was not sure about the logic.  Should we add a test that includes something going into the "doNotCalculate" file?

align/constGap.go
RavenYL: also make gapSize threshold customizable to test "doNotCalculate"

RavenYL:
> Great to see these improvements and bug fix! There was one place where I was not sure about the logic. Should we add a test that includes something going into the "doNotCalculate" file?

Yes, just made gapSizeLimit customizable and added a test with non-empty doNotCalculate files

align/view_test.go
RavenYL: added tests

align/constGap.go
```diff
@@ -277,14 +277,36 @@ func writeCigar(trace [][]ColType, i_inChecker_max_Previous int, j_inChecker_max
...
-//inputs: the growing route describing the collection of cigars of the entire alignment except the last cigar, the index of the cigar that is currently being built, the ColType (M=0,I=1,D=2) of the last cigar entry
+//inputs: lengths of original sequences alpha and beta, the growing route describing the collection of cigars of the entire alignment except the last cigar, the index of the cigar that is currently being built, the ColType (M=0,I=1,D=2) of the last cigar entry
 //outputs: the updated final route describing the collection of cigars of the entire alignment including the last cigar, the updated and index of the cigar that is currently being built
-func lastCigar(route []Cigar, routeIdx_current int, Op_end ColType) ([]Cigar, int) {
+func lastCigar(len_alpha int, len_beta int, route []Cigar, routeIdx_current int, Op_end ColType) ([]Cigar, int) {
+	//Calculate the size of final runlength
+	var TotalRunLength, LastRunLength int64
+	TotalRunLength = 0
+	switch Op_end {
+	case 1: //Op_end==1=ColI=Insertion=unique sequence in sequenceTwo beta
```
dasnellings, line 287: It took me a second to figure out what the cases were here. The comments are very helpful but it may be easier to read if you make match, insertion, and deletion constants so you can do 
```
switch Op_end {
case Insertion:
case Deletion:
default:
}
```

dasnellings (approved):

craiglowe (approved):
Great work!

PR #652, opened 2022-03-17
basic generics

dasnellings:
Now that Go 1.18 is out I have been playing around the generics. This PR removes the typed min/max functions in favor of a single generic min and max function in the numbers package. This is a fairly big change that would require everyone to update to 1.18 and introduces a new dependency to golang.org/x/exp as the generic packages are in googles experimental branch until generics get more field testing, however the expectation is these will move to the standard library in a future version of go. I think this would be worth a larger discussion before adding generics to gonomics so I am adding a do not merge tag for the time being.


craiglowe:
This is exciting!  What is the down side?  I know you mentioned the dependency on exp and also needing everyone to be on 1.18 or greater.  Those seem acceptable to me (other people's thoughts?).  Any other read/potential downside that you see?

numbers/minMax.go
```diff
@@ -1,135 +1,34 @@
...
-	for i := 1; i < len(a); i++ {
-		if a[i] > answer {
-			answer = a[i]
+import (
+	"golang.org/x/exp/constraints"
+	"log"
+)
+
+// Max returns the maximum of the input values of an ordered type.
+func Max[E constraints.Ordered](s ...E) E {
```
craiglowe, line 9: What do your benchmarks look like for this compared to just MinInt(a,b)?  I am wondering if turning the two ints into a slice slows things down?
Even if it does slow things down, maybe we could keep it but have two separate functions for min/max of two numbers, and min/max of any amount?

numbers/minMax.go
```diff
@@ -1,135 +1,34 @@
...
+	if len(s) == 0 {
+		log.Panic("empty argument to max function")
+	}
+	var max E = s[0]
+	for i := range s {
```
craiglowe, line 14: very minor, but since you are initializing to s[0], can we start the loop at 1 instead of 0?

craiglowe (commented):
Lots of great stuff here.  I favor moving in this direction.  I added two small comments.

craiglowe:
Does anyone have concerns about this?  Biggest thing I see would be making sure this doesn't break things for people when they go to run/compile main, but don't have >=1.18.  Can we have everyone upgrade their systems over the next couple days before we merge this so that once this gets merged, everyone is already on 1.18?

common/math_test.go
```diff
@@ -0,0 +1,77 @@
...
+func intContains(s []int, v int) bool {
+	for i := range s {
+		if s[i] == v {
+			return true
+		}
+	}
+	return false
+}
+
+func genericContains[E comparable](s []E, v E) bool {
```
RavenYL, line 18: If we can go over what E is that would be helpful for me!

numbers/minMax_test.go
```diff
@@ -162,7 +85,7 @@ var tripleMaxTests = []struct {
 
 func TestTripleMax(t *testing.T) {
 	for _, test := range tripleMaxTests {
-		actual := TripleMax(test.a, test.b, test.c)
```
RavenYL, line 165: There might be other packages that need fixing. I remember using TripleMaxTrace in the align package, which compares scores in the 2x2 grid to decide which direction to proceed in the alignment matrix

RavenYL (commented):
Although I'm always scared to go through updates and changes I agree that this is the direction we should go in

common/math_test.go
RavenYL: A Tour of Go's new section of generics has been helpful: https://go.dev/tour/generics/1

dasnellings: Yes! There is also a more in-depth tutorial that is pretty good https://go.dev/doc/tutorial/generics.

You probably figured it out from the Tour of Go, but the E was just a placeholder for the type name. In the blog I was looking at they used 'E'. In the Tour of Go they use 'T', but it could be anything.

numbers/minMax_test.go
dasnellings: I don't think the tripleMaxTrace function uses the TripleMax function so it shouldn't be affected.

craiglowe (approved):
Wonderful!  Looks fantastic!

RavenYL (approved):

RavenYL:
I've upgraded to golang1.18 both locally and on DCC and am ready to merge when others are

PR #653, opened 2022-03-17
Craig find

craiglowe:
I need to find and replace thousands of strings (sequence names) in a file.  I was not sure how to do this with *nix tools without doing many passes over the file (which is large), so I wrote a small program to do this with a single pass over the file.

### DIFF
--- a/cmd/findReplace/findReplace.go
+++ b/cmd/findReplace/findReplace.go
@@ -1,0 +1,80 @@
+// Command Group: "General Tools"
+
+package main
+
+import (
+	"flag"
+	"fmt"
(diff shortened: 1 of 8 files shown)


rimangan:
There is already a cmd in gonomics called "FindAndReplace". Is it possible to make this function an option of the other one, or can we find a more descriptive name for either of these to avoid confusion going forward?

craiglowe:
That is a good point.  Let me see if I can incorporate this functionality into the existing program

craiglowe:
I am closing this to instead add functionality to an existing program

PR #654, opened 2022-03-17
add Luke, update go docs, add testing coverage

cshoben:
1) Adds Luke as an author
2) Adds documentation for bed/methods.go go docs
3) Adds a bit more testing coverage for the bed/methods.go


bed/methods.go
```diff
@@ -30,15 +35,22 @@ func (b Bed) UpdateCoord(c string, start int, end int) interface{} {
...
 
+// Push and pop() satisfy the interface for the heap method
+// Push pushes to the heap.
 func (b *BedSlice) Push(x interface{}) {
 	answer := x.(*Bed)
 	*b = append(*b, answer)
 }
 
+// Pop and push() satisfy the interface for the heap method
+// Pop pushes to the heap.
```
dasnellings, line 53: I think this should be something like "Pop returns a value from the heap" rather than pushing to the heap.

bed/methods_test.go
```diff
@@ -66,3 +74,28 @@ func TestBed_UpdateCoord(t *testing.T) {
 		}
 	}
 }
+
+//	Tests for methods that use: type BedSlice []*Bed
+func TestBedSlice_Len(t *testing.T) {
+	var allBed BedSlice
+	for _, v := range BedMethodTests {
+		pointer := &v.b //must create a pointer because type BedSlice is []*Bed
+		allBed = append(allBed, pointer)
```
dasnellings, line 83: No issue here, but just for reference you can append a pointer without going through an intermediate variable.
`allBed = append(allBed, &v.b)`

bed/methods.go
```diff
@@ -9,18 +9,23 @@ import (
...
 func (b Bed) GetChromStart() int {
 	return b.ChromStart
 }
 
+// GetChromEnd grabs the end coordinates of the bed struct
 func (b Bed) GetChromEnd() int {
 	return b.ChromEnd
 }
 
+// UpdateCoord will change the bed struct to the user defined chrom string,
```
craiglowe, line 27: I believe changing/modifying the bed struct was the original plan, but this function right now creates a copy of the bed struct and returns that copy with modified coordinates (and the original is unchanged).

bed/methods.go
```diff
@@ -9,18 +9,23 @@ import (
 // Current methods satisfy requirements for the following interfaces:
 // bed.BedLike
 
+// GetChrom grabs the chrom of the bed struct
```
craiglowe, line 12: changing "grabs" to "returns" may help people who are starting out with coding by using more precise (and google-able) language.

bed/methods.go
craiglowe: One option would be to also mention that the element (along with being returned) is also removed from the heap.

bed/methods.go
```diff
@@ -57,10 +71,14 @@ func (b BedSlice) Write(file string) {
 	exception.PanicOnErr(err)
 }
 
+// WriteToFileHandle will write a bed struct to the
+// io.Writer
 func (b Bed) WriteToFileHandle(file io.Writer) {
 	WriteToFileHandle(file, b) //adaptive field writing seems most flexible for the method
 }
 
+// NextRealRecord keeps track of if we are done reading through
```
craiglowe, line 80: This comment is true, but we should also add that it sets b equal to the next bed entry in the file (along with keeping track of if we are done reading the file).

bed/methods.go
```diff
@@ -84,6 +103,8 @@ type ByGenomicCoordinates struct {
 	BedSlice
 }
 
+// Less sorts by genomic coordinate and satisfies the sort.Slice
```
craiglowe, line 106: May be helpful to stay that it sorts by chromosome, and then entries within the same chromosome are sorted from smallest to largest by genomic coordinates (first by start coordinate and then ties are broken by the end coordinate).

craiglowe (commented):
Great to have this added documentation!  I made a few tiny comments while reading through it.

RavenYL (dismissed):

craiglowe (approved):
Great work!

dasnellings (approved):

PR #657, opened 2022-03-21
Generic sort

dasnellings:
This PR makes the mergeSort command a generic function. The most important changes are in sort/mergeSort.go, but there are several other changes made as a result of this. The main benefit of this is that we no longer need to implement a laundry list of methods in order to do an external merge sort (e.g. push, pop, len, swap, write, etc.). All we need to perform a merge sort on any data type is to input a `Less(a, b Type) bool` function to the sort function and the rest is handled internally.


go.mod
```diff
@@ -3,3 +3,5 @@ module github.com/vertgenlab/gonomics
 go 1.18
 
 require golang.org/x/image v0.0.0-20210220032944-ac19c3e999fb
+
+require golang.org/x/exp v0.0.0-20220317015231-48e79f11773a // indirect
```
RavenYL, line 7: Are the changes to go.mod and go.sum the same as in the generics MinMax PR?

sort/mergeSort.go
```diff
@@ -2,204 +2,157 @@ package sort
...
-	"path"
 	"sort"
 )
 
-const (
-	// Newly added filetypes should be added to the following valid types list
-	validFileTypes     = "axt, bed, vcf, sam, giraf"
-	maxTmpFilesAllowed = 1000
-)
+type heapBuf []any
```
RavenYL, line 12: You save so many lines of code

RavenYL: I'm learning here that the generic type can just be declared with this 1 line. I thought you had to specify it somehow, but it seems like that's just because the example in A Tour of Go was a "singly-linked list" that requires a pointer and a value

sort/sort.go
```diff
@@ -13,6 +13,10 @@ import (
 	"sync"
 )
 
+const (
+	maxTmpFilesAllowed = 1000
```
RavenYL, line 17: Why do we have this limit?

sort/mergeSort.go
```diff
@@ -2,204 +2,157 @@ package sort
...
+	// and key them to the decoder for the origin file. The memory addresses for the buffer
+	// is kept static throughout the heap loop below and therefore can be used as a stable key
+	// to determine the file of origin
+	memoryAddressMap := make(map[*E]*gob.Decoder)
+
+	curr := make([]*E, len(tmpFiles)) // this functions as a buffer for hb with static memory addresses
+	for i := range curr {
+		curr[i] = new(E) // allocate mem for curr pointers
+	}
+	hb := make(heapBuf, len(tmpFiles))
```
RavenYL, line 75: I'm going to take notes on this because this type of make() works, but when I tried to make or use a variable with a generic type I got a "can't use generic type without instantiation" error

RavenYL (dismissed):

sort/mergeSort.go
dasnellings: I am still learning generics so I am not positive but I suspect the compiler makes exceptions for the `any` type. In most cases you need to instantiate when using a generic type but `any` is an alias for an empty interface `interface{}` so maybe that is why this is ok?

sort/sort.go
dasnellings: I was running into some errors that I exceeded the maximum number of open files on my machine so I put a cap on it.

craiglowe (approved):
Great to have these improvements!

PR #661, opened 2022-03-25
multiFaToVcf Runtime Optimization

rimangan:
multiFaToVcf uses AlnPosToRefPos, which converts a multiFa alignment position to the corresponding reference coordinate, each time a variant is encountered along a chromosome. I've implemented AlnPosToRefPosCounter, which allows us to start our search for the next AlnPos/RefPos pair midway through a chromosome. This feature exists for RefPosToAlnPos, but not for AlnPosToRefPos on main.
Should be a substantial speedup for genome-wide data, by orders of magnitude. Back of the envelope calculation for chromosome 2 suggests about a 60 million times speedup

### DIFF
--- a/fasta/multiFa.go
+++ b/fasta/multiFa.go
@@ -29,15 +29,22 @@ func RefPosToAlnPosCounter(record Fasta, RefPos int, refStart int, alnStart int)
 //AlnPosToRefPos returns the reference position associated with a given AlnPos for an input Fasta. If the AlnPos corresponds to a gap, it gives the preceeding reference position.
 //0 based.
 func AlnPosToRefPos(record Fasta, AlnPos int) int {
-	var RefPos int = 0
-	for t := 0; t < AlnPos; t++ {
+	return AlnPosToRefPosCounter(record, AlnPos, 0, 0)
+}
+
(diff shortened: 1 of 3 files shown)


fasta/multiFa.go
```diff
@@ -29,15 +29,22 @@ func RefPosToAlnPosCounter(record Fasta, RefPos int, refStart int, alnStart int)
...
+		refStart, alnStart = 0, 0 //in case the alnStart was improperly set (greater than the desired postion, we reset the counters to 0.
+	}
+	for t := alnStart; t < AlnPos; t++ {
 		if t == len(record.Seq) {
 			log.Fatalf("Ran out of chromosome.")
 		} else if record.Seq[t] != dna.Gap {
-			RefPos++
+			refStart++
 		}
 	}
```
RavenYL, line 46: Will it be even faster if you won't need to loop through the positions from alnStart to AlnPos, but update refStart in 1 step? Is there an efficient 1-step solution to count the number of dna.Gap in record.Seq[alnStart : AlnPos-1], and subtract that number of dna.Gap when you add (AlnPos-1-alnStart) to refStart?

RavenYL (approved):
Glad to see the algorithm speed up so much!!!

cfauci (approved):

PR #663, opened 2022-04-06
Vcf info/format parsing bugfix

dasnellings:
The vcf info/format parser panics if you have multiple samples in the records and some samples have multiple value fields (as might happen for polyallelic sites) and some samples have a nil '.' value. This PR updates the logic for handling '.' values in the format and info fields to avoid this error.


cshoben (approved):

craiglowe (approved):
Looks good!  Wow, the vcf file format is so loose; parsing it is not easy.

RavenYL (approved):

PR #664, opened 2022-04-12
Raven align anchor3

RavenYL:
globalAlignmentAnchor3
- Fixed bugs, such as dividing by 0 when the size of a gap is 0 and the program needed to divide gap sizes to check if one is too big compared to the other
- Added a new feature, which is to recognize insertions that don't need to be aligned (e.g. in chimp, the gap between previously aligned chunks is 0; in human, the gap between previously aligned chunks is >0, meaning the corresponding sequence in humans is an insertion, or vice versa for chimp insertions). Added a go test case for this. Using this new feature, was able to recognize the METTL14 rainbow region as an insertion


rimangan (approved):

cshoben (approved):
I'm interested in exploring that "path" package you used here.

PR #665, opened 2022-04-14
the beginning of slurmCheck

cshoben:
This is the beginning of my implementation of the Parasol Job list check (https://genecats.gi.ucsc.edu/eng/parasol.html) for SLURM. 

The overall goal is to have a way to check if all jobs submitted through a job array on the SLURM scheduler finish to completion. 

The end goal of my code for slurmCheck is to take in a job array text file that includes some extra documentation on each line (termed a "fancy" job array file), parse this file and create a normal SLURM job array file, create the necessary sbatch file, submit the jobs to SLURM, and then complete a "check" on all expected output files. The program will then output a new fancy job array text file that contains ONLY the jobs that did NOT pass the "check". The user could then look deeper into why these particular jobs did not run as expected, and the new fancy file can be fed back into slurmCheck for another interaction, etc. 

This "fancy" file must tell slurmCheck which file is the output file, and which type of check to perform. We use "{" and "}" to help with parsing. This decision is mirroring how Parasol functions, which is useful in that a user could have a fancy file that would be functional for either job scheduler. 

There are four check options:

1.    exists - file must exist
2.    exists+ - file must exist and be non-empty
3.    line - file must end with a complete line
4.    line+ - file must end with a complete line and be non-empty

The type of check will be specified by the user in their fancy job array file. 

Here's an example of a blat job spec:

  blat human.fa mouse2.fa {check out line+ hm2.psl}

"check out" tells slurmCheck that the file inside the squiggles is the expected output file. 
"line+" tells slurmCheck which type of check to perform on this output file. 

In this PR slurmCheck is only able to parse a fancy array file. This PR includes testing for this feature. I know the testing is not the way we normally do it, I do plan to update this to match other gonomics testing files in a future PR.


cmd/slurmCheck/slurmCheck.go
```diff
@@ -0,0 +1,143 @@
...
+		"slurmCheck - Used to check for completion of SLURM job arrays. Takes in a 'fancy' version of a job array text file.\n" +
+			"Outputs a new fancy version of a job array text file with all jobs that did NOT meet the user specified\n" +
+			"check.\n" +
+			"The four types of checks:\n" +
+			"'exists'\tfile must exist\n" +
+			"'exists+'\tfile must exist and be non-empty\n" +
+			"'line'\t\tfile must end with a complete line\n" +
+			"'line+'\t\tfile must end with a complete line and be non-empty\n" +
+			"Usage:\n" +
+			"inputFancyJobArrayFile.txt\n")
```
RavenYL, line 89: Is the Usage actually "slurmCheck inputFancyJobArrayFile.txt"?

cmd/slurmCheck/slurmCheck_test.go
```diff
@@ -0,0 +1,41 @@
...
+		actual, err := os.Create(v.actualOutputParseTheInput)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+		fmt.Fprintf(actual, "begin: %s \n out: %s \n check: %s \n end: %s \n", begin, out, check, end)
+		if !fileio.AreEqual(v.actualOutputParseTheInput, v.expectedOutputParseTheInput) {
+			t.Errorf("Error in slurmGeneralizedCheck, expected %s, actual %s", v.expectedOutputParseTheInput, v.actualOutputParseTheInput)
+		} else {
+			err = os.Remove(v.actualOutputParseTheInput)
```
RavenYL, line 35: There's a file called "testdata/actualOutputParseTheInput" that is not removed in this PR, but looks the same as the expectedOutputParseTheInput.txt. Maybe "testdata/actualOutputParseTheInput.txt" was generated and removed, but "testdata/actualOutputParseTheInput" was not removed. Is this intentional?

cmd/slurmCheck/slurmCheck.go
```diff
@@ -0,0 +1,143 @@
...
+	//"io"
+	"log"
+	"strings"
+	//"time"
+)
+
+type slurmCheckArray struct {
+	begin      string // whatever is in front of the squiggles
+	outToCheck string // this is the file to be checked
+	checkType  string // this is the code for how we check the output files
```
RavenYL, line 17: What I understood was that the inputFancyFile.txt already has info on whether each slurm job output file meets the checks (empty, line, etc.), and this program parses inputFancyFile.txt so you can see that info more clearly. Is this right? Or is this program doing the checks (empty, line, etc.)?

cmd/slurmCheck/slurmCheck_test.go
cshoben: Good catch. I'm going to remove it. Thank you

cmd/slurmCheck/slurmCheck.go
cshoben: Yes, another great catch. Thank you!

cmd/slurmCheck/slurmCheck.go
cshoben: This program is going to submit the job array to SLURM, and do the check. Since there are check options and depending on the software being used, the output file could be argument 2, 3, etc, the input file needs to be different than a normal job array. We use squiggles "{ }" in this specialized job array file to tell slurmCheck where to look, this is the file I'm calling the "fancyFile.txt". The user must tell slurmCheck which file is the expected output file and which type of "check" the user wants to be performed.

RavenYL: Thanks! Are you still working on the part where the program interacts with SLURM to submit the job array and do the check? Or did you already write it in this PR and I missed it (I didn't find sbatch or anything like that in the uncommented code)?

cshoben: I wanted to submit my PRs in increments to introduce the idea and gett feedback as I'm going versus at the end. This is only parsing the file, which will likely also update as I write the code for the other features.

RavenYL: OK got it! Thanks for the explanation! Excited to see how gonomics will now be programmed to interact with SLURM

RavenYL (approved):

cmd/slurmCheck/slurmCheck.go
```diff
@@ -0,0 +1,143 @@
...
+	//separate the string into "words" aka fields. White space defined the boundary of each field
+	fields := strings.Fields(parsedSecondSquiggle[0])
+
+	//third field will be the type of check to perform
+	slurmLineParsed.checkType = fields[2]
+
+	//fourth field is the output file we will be checking
+	slurmLineParsed.outToCheck = fields[3]
+
+	slurmLineParsed.end = parsedSecondSquiggle[1]
```
rimangan, line 65: this line will panic if the output file is at the end of the cmd line. you can check the len(parsedSecondSquiggle) as the conditional to avoid this.

rimangan (approved):

craiglowe (approved):
Wonderful!

PR #667, opened 2022-04-19
first commit multiFaReduce

craiglowe:
This command removes all non-variable columns from a multi fasta alignment


cmd/multiFaReduce/multiFaReduce.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/dna"
+	"github.com/vertgenlab/gonomics/fasta"
+	"log"
+)
+
+func makeEmptyFastas(records []fasta.Fasta) []fasta.Fasta {
```
rimangan, line 13: I think there is a function in fasta called SegregatingSites that performs a similar or identical task. We might want to see if we can consolidate to reduce redundancy

craiglowe: good point.  we don't want to maintain two versions of the same thing.  i will look into this.

cmd/multiFaReduce/testdata/expected.mfa
```diff
@@ -0,0 +1,8 @@
+>one
+AACC
```
RavenYL, line 2: How can this output file be used downstream? Intuitively I think it might be confusing that after removing the non-variable columns, the remaining variable columns are smashed together, so we don't know their original positions in the original mfa. I would have thought mfas would show the alignments like this:
>one
----AACC--
>two
----AAAA--

RavenYL (approved):

craiglowe:
I have updated this pull request to call the existing function in fasta (fasta.SegregatingSites).  I updated the fasta.SegregatingSites function have slightly less complexity by breaking two the the tasks it does into subfunctions.

cmd/multiFaReduce/multiFaReduce.go
craiglowe: I have now fixed this and call the existing function (I have eliminated the duplicate that I wrote).

cmd/multiFaReduce/testdata/expected.mfa
craiglowe: You make a good point that after reducing an alignment to only the segregating sites, we can not recover the original alignment, or even map the segregating sites back to the original alignment.  I think the downstream analyses will be to calculate statistics on these segregating sites and they will not be brought back to the bigger alignments, but we could consider some way to keep track of things so that more of the original alignment could be recovered.

rimangan (approved):

PR #669, opened 2022-04-20
Craig mfa extract

craiglowe:
This adds some minor functionality to the code

### DIFF
--- a/cmd/multiFaExtract/multiFaExtract.go
+++ b/cmd/multiFaExtract/multiFaExtract.go
@@ -15,30 +15,27 @@ func multiFaExtract(s Settings) {
 	var ans []fasta.Fasta
 	records := fasta.Read(s.InFile)
 	if s.Bed == "" {
-		if !(s.Start < s.End) {
-			log.Fatalf("Invalid arguments, start must be lower than end")
-		}
-		ans = extractMultiHelper(records, s.Start, s.End)
-		if s.RemoveGaps {
-			ans = fasta.RemoveGaps(ans)
-		}
+		ans = extractMultiHelper(records, s.Start, s.End, s.RemoveGaps)
 		fasta.Write(s.OutFile, ans)
 	} else {
 		bedChan := bed.GoReadToChan(s.Bed)
 		for b := range bedChan {
-			ans = extractMultiHelper(records, b.ChromStart, b.ChromEnd)
-			if s.RemoveGaps {
-				ans = fasta.RemoveGaps(ans)
-			}
+			ans = extractMultiHelper(records, b.ChromStart, b.ChromEnd, s.RemoveGaps)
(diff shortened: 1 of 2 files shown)


rimangan (approved):

cmd/multiFaExtract/multiFaExtract.go
```diff
@@ -15,30 +15,31 @@ func multiFaExtract(s Settings) {
...
-func extractMultiHelper(records []fasta.Fasta, start int, end int) []fasta.Fasta {
+func extractMultiHelper(records []fasta.Fasta, start int, end int, removeGaps bool, revComp bool) []fasta.Fasta {
 	var ans = make([]fasta.Fasta, len(records))
+	if !(start < end) {
+		log.Fatalf("Invalid arguments, start must be lower than end. start=%d end=%d\n", start, end)
+	}
 	for i := range records {
 		ans[i] = fasta.Extract(records[i], fasta.RefPosToAlnPos(records[0], start), fasta.RefPosToAlnPos(records[0], end), records[i].Name)
+		if revComp {
+			ans[i] = fasta.Copy(ans[i])
```
RavenYL, line 37: fasta.Copy creates a copy of the fasta struct and the associated slice of dna so that modifying one will not also modify the other, but by assigning the fasta.Copy product of ans[i] to ans[i] itself, is fasta.Copy not ultimately doing anything here?

RavenYL: This line takes ans[i], makes new struct, as well as new slice of DNA (end up with 2 structs, each pointing at their own slice of DNA), aka separate in memory, then assign ans[i] to it, so now ans[i] is its own struct not pointing to the genome

RavenYL (approved):

RavenYL (changes requested):
If following strand could be written as a option that would be great!

cmd/multiFaExtract/multiFaExtract.go
```diff
@@ -15,30 +15,35 @@ func multiFaExtract(s Settings) {
...
 		fasta.Write(s.OutFile, ans)
 	} else {
 		bedChan := bed.GoReadToChan(s.Bed)
 		for b := range bedChan {
-			ans = extractMultiHelper(records, b.ChromStart, b.ChromEnd)
-			if s.RemoveGaps {
-				ans = fasta.RemoveGaps(ans)
+			ans = extractMultiHelper(records, b.ChromStart, b.ChromEnd, s.RemoveGaps, b.Strand == bed.Negative)
+			if b.FieldsInitialized >= 4 {
+				fasta.Write(fmt.Sprintf("%s.fa", b.Name), ans)
```
cshoben, line 25: Can you help me understand why we would name the output file differently if the fields initialized are less than or equal to 4? I'm assuming it would be to not lose information, but then we do not seem to include the chrom start end anymore?

cshoben (commented):
Not sure if we are waiting to review this one, but happy to approve when it's time.

PR #670, opened 2022-04-21
this is first draft of wigStats

sweavs111:
wigStats provides coverage statistics for input files. do not merge, flag is pending testing


wig/wig.go
```diff
@@ -139,3 +139,15 @@ func WriteToFileHandle(file io.Writer, rec Wig) {
 
 	}
 }
+
+// ChromToSlice returns the values from a wig entry matching a user-specified chromosome name
+func ChromToSlice(w []Wig, chrom string) []float64 {
```
RavenYL, line 144: Great to have this function!

cmd/wigStats/wigStats.go
```diff
@@ -0,0 +1,77 @@
...
+		exception.PanicOnErr(err)
+	}
+	err = out.Close()
+	exception.PanicOnErr(err)
+
+}
+
+func usage() {
+	fmt.Print(
+		"wigStats - provide coverage statistics for WIG format visualization files.\n" +
```
RavenYL, line 52: If the requirements of the program can be specified in the usage that would be great, e.g. wig values must be non-negative integers (because stats are calculated by calling index numbers)

RavenYL (commented):
Just waiting for specifications in the usage and a test file!

wig/wig.go
```diff
@@ -139,3 +139,15 @@ func WriteToFileHandle(file io.Writer, rec Wig) {
...
+	var output []float64
+	for _, v := range w {
+		if v.Chrom == chrom {
+			output = make([]float64, len(v.Values))
+			output = v.Values
+		}
+	}
+	return output
+}
```
cshoben, line 153: It'd be good to add testing for this function in the wig_test.go file.

cmd/wigStats/wigStats.go
cshoben: I would also add that this can only handle fixed step Wig files. Also, I'm unable to tell, but can this handle wig files with a Step and/or Span other than "1"? If not, worth saying that here. And I would throw an error somewhere in the code if it is unable to handle this. 

On that note, I don't know if the wig is read in so that we can ask for a value at any coordinate if the Step and/or Span != 1? I see the wig.go file reads in Step and Span !=1, but I think it just stores the string, not that it creates a []int where each chr coordinate has a value. I think you would have to do this math in this code. Otherwise, a non-1 step/span would ruin the logic used on line 27 I believe. Technically, wig.Read can read in a Step and/or Span != 1, so I would again throw an error in this code and add this limitation to the usage statement. bedValueWig.go also has this limitation.

cmd/wigStats/testdata/test.wig
```diff
@@ -0,0 +1,11 @@
+fixedStep chrom=chr1 start=1 step=1
+1
+-10
```
cshoben, line 3: I saw in the usage that the wig needs to have non-negative integers. It looks like this test wig would break that rule?

wig/wig.go
```diff
@@ -139,3 +139,15 @@ func WriteToFileHandle(file io.Writer, rec Wig) {
...
 	}
 }
+
+// ChromToSlice returns the values from a wig entry matching a user-specified chromosome name
+func ChromToSlice(w []Wig, chrom string) []float64 {
+	var output []float64
+	for _, v := range w {
+		if v.Chrom == chrom {
+			output = make([]float64, len(v.Values))
+			output = v.Values
```
craiglowe, line 149: Do we need each wig entry to be an entire chromosome?  I think this is the case for most of the wigs we make with gonomics, but the actual wig spec is much more flexible. Should we have a check in this loop to make sure we don't get more than one wig of the same chromosome and that it starts at zero with a step of 1?

craiglowe (commented):
One quick comment

cmd/wigStats/testdata/test.wig
RavenYL: Seth said the negative value is the missing data value, and has updated Usage to reflect this

cmd/wigStats/wigStats.go
```diff
@@ -0,0 +1,77 @@
...
+			"Usage:\n" +
+			"wigStats in.wig noGap.bed output.tsv\n" +
```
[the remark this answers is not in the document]
RavenYL: It's a bed file with no gaps. We have noGap versions of assemblies like hg38 in our DCC shared directories

cmd/wigStats/wigStats.go
```diff
@@ -0,0 +1,80 @@
...
+	var err error
+	var i, j int
+	statValues = make([]int, 1000)
+	for i = range records {
+		wigValues = wig.ChromToSlice(w, records[i].Chrom)
+		for j = records[i].ChromStart; j < records[i].ChromEnd; j++ {
+			if wigValues[j] == missingDataValue {
+				continue
+			}
+			if int(wigValues[j]) > len(statValues) {
```
RavenYL, line 31: Did we decide if this line needs to be changed to int(wigValues[j]) >= len(statValues)?

craiglowe: Yes, I think this is the last change that we need before merging.

craiglowe (commented):
I think we are waiting on a fix that is needed in an "if" statement.  Chelsea also pointed out that a function could use some testing

cmd/bedMaxWig/bedMaxWig.go
```diff
@@ -22,7 +22,7 @@ func bedMaxWig(infile string, database string, chromsizeFile string, outfile str
 	var i int
 
 	for i = 0; i < len(sizes); i++ {
-		chromSlice = WigChromToSlice(wigData, sizes[i].Size, sizes[i].Name)
+		chromSlice = wig.ChromToSlice(wigData, sizes[i].Name)
```
cshoben, line 25: I think this file was changed to bedValueWig.go back in the spring or early summer. I think there will be a conflict if we try to merge? My guess is this code was written without having done a `git pull` for the most recent version of main. Might need to also check that `go test ...` is also happy.

wig/wig.go
```diff
@@ -139,3 +139,23 @@ func WriteToFileHandle(file io.Writer, rec Wig) {
...
 	}
 }
+
+// ChromToSlice returns the values from a wig entry matching a user-specified chromosome name
+func ChromToSlice(w []Wig, chrom string) []float64 {
+	var output []float64
+	for _, v := range w {
+		if v.Chrom == chrom {
+			if v.Step == 1 {
+				if v.Start == 0 {
```
cshoben, line 149: Doesn't the wig start at 1 in the test example? I don't see the wig.Read() turning the wig.Start into a 0 if it is a 1, but I could be missing it?>

cfauci: Test fails on the branch with error code "start coordinate invalid."

cfauci (approved):

PR #671, opened 2022-04-26
fix gonomics cmd error after intervalOverlap rename. 

dasnellings:
the gonomics command was panicing because it expected to find intervalOvlerap.go in cmd/intervalOverlap, but the files were still named sequelOverlap. This renames the files and fixes the error.


RavenYL (approved):
Thanks for catching this Dan!

rimangan (approved):

PR #672, opened 2022-04-26
in silico PCR

dasnellings:
Wrote a new command `simulatePcr` that inputs a reference sequence and an arbitrary number of primer sequences and outputs a bed file of predicted products.


fasta/fasta.go
```diff
@@ -46,6 +47,36 @@ func Read(filename string) []Fasta {
 	return answer
 }
 
+// ReadToString reads a fasta file to a map of sequence strings keyed by the record name.
+func ReadToString(filename string) map[string]string {
```
RavenYL, line 51: I didn't know we were missing this function before!

RavenYL (approved):
Looks good to me, and it's a great functionality to have!

rimangan (approved):

PR #675, opened 2022-05-04
Pad bed regions

dasnellings:


cmd/bedFormat/bedFormat.go
```diff
@@ -23,6 +23,13 @@ func bedFormat(s Settings) {
 	}
 
 	for v := range ch {
+		if s.Pad > 0 {
+			v.ChromStart -= s.Pad
+			v.ChromEnd += s.Pad
+			if v.ChromStart < 0 {
+				v.ChromStart = 0
+			}
```
rimangan, line 31: no check for chromEnd > len(chrom). Perhaps pad will require a chromsizes file?

craiglowe: Yes, I think we probably want a check to make sure we don't go off the end.

craiglowe (commented):
I think requiring a chromSize file for this option sounds like the way to go so that we don't end up with "illegal" bed records.
We could also think about making the name of this program more general?  I am not sure if I would think that bedFormat would expand the records.  Maybe something like "bedOperations?"  bedtools is already taken.

PR #678, opened 2022-05-10
Lift modifications

dasnellings:
Minor modifications to liftCoordinates. Switched to using fasta.Seeker to reduce runtime and added a new option to swap some fields in INFO that are common for microarray data.


cmd/liftCoordinates/liftCoordinates_test.go
```diff
@@ -23,10 +23,10 @@ var LiftTests = []struct {
 
 func TestLift(t *testing.T) {
 	for _, v := range LiftTests {
-		liftCoordinates(v.chainFile, v.inputFile, "tmp.vcf", v.faFile, "tmp.unmapped", 0.95, 0)
+		liftCoordinates(v.chainFile, v.inputFile, "tmp.vcf", v.faFile, "tmp.unmapped", 0.95, false, 0)
```
RavenYL, line 26: If there could be a test case for swapAlleleAB in the future, that would be great!

RavenYL (dismissed):

cmd/liftCoordinates/liftCoordinates_test.go
cshoben: I think it'd be good to add test coverage for this option in this PR before merging it in. An example, where we swap A and B.

cshoben (changes requested):

dasnellings:
I've added a few test cases. I imagine you would want to swap A and B if you have the annotation, but I chose to add it as a flag so we are not mandating computation that will likely not be relevant for the vast majority of use cases. If there is not ALLELE_A and ALLELE_B we cannot do a swap since there is nothing to swap with. In a biallelic case we could make 0 -> 1 and 1 -> 0, but this breaks down with multiallelic sites. This would probably only happen with a malformed VCF, though who knows what wacky things people come up with. The current behavior is to not modify the record if either A or B is missing and make a log warning print if 1 of the two are present.

craiglowe (approved):
Great!  Thank you for adding a test!

cmd/liftCoordinates/testdata/expected_swapAB.vcf
```diff
@@ -0,0 +1,3 @@
+chr1New	5	rs100	T	A	100	PASS	NoSwap;ALLELE_B=1;ALLELE_A=0	GT	0|0	1|0	1|1
+chr1New	7	rs101	C	T	100	PASS	InvertAndSwap;ALLELE_A=0;ALLELE_B=1	GT	1|0	1|1	0|0
+chr1New	75	rs102	T	A	100	PASS	InvertAndPosChange	GT	1|1	1|1	0|0
```
cshoben, line 3: should this say "chr1New 65" ?

cshoben (approved):
Just one last comment, but I dont think it holds back an approval. Feel free to merge if my question is a no.

rimangan (approved):

PR #679, opened 2022-05-16
Raven align anchor4

RavenYL:
My goal is to fix outliers, aka the places where the alignment follows the wrong path and deviates greatly from the diagonal (see illustration). In order to achieve this goal, I made the following changes.

Edited the gapBedPass helper function and the matchToGap function:
- The gaps that we doNotCalculate are no longer used to update alignment position trackers in the alignment, so that consecutive outliers are not accidentally accepted and calculated (to do this, I realized I need to use the alignment position trackers to check whether gapBedPass, but use the original bed positions to write to bed. Also, I need to use variables to find previous chromosome and position, rather than relying on indexing i-1, since outliers may cause us to have to skip multiple index numbers)
- Rescue largeGapSize if it comes back to within 5% of the diagonal (but check this before checking largeGapSizeProduct)
- Realized that "the most recently updated alignment position" (pos_species) needs to be considered separately from any bed entry's ChromStart and ChromEnd, so now gapBedPass needs more input variables
- Rearranged the order of if/else checks

Added a matchMafPass helper function:
- In addition to checking if entries are from the same chromosome (special case for chimp chr2a and chr2b matching to human chr2), reject maf entries that are off the diagonal by over 5%*SrcSize (SrcSize is chromosome length)

Updated test files.

Before:
panTro6_vs_hg38_1

After:
plotall


cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -25,40 +25,78 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
...
+	} else if (float64(species1_ChromStart) <= float64(species2_ChromStart)-0.05*float64(species2_SrcSize)) || (float64(species1_ChromStart) >= float64(species2_ChromStart)+0.05*float64(species2_SrcSize)) {
+		pass = false
+	}
+
+	return pass
+}
+
 // helper function: check if gap bed entry passes checks
-func gapBedPass(species1_ChromStart int, species1_ChromEnd int, species2_ChromStart int, species2_ChromEnd int, gapSizeProductLimit int) (bool, string, string) {
+// need both pos_species (most recently updated alignment position)
```
cshoben, line 49: What does "pos" mean in "pos_species"?

cshoben (commented):
Only at the beginning of reviewing.... but wanted to ask, are there empty files included in this PR that need to be removed? Or are these being kept around on purpose?

RavenYL:
> Only at the beginning of reviewing.... but wanted to ask, are there empty files included in this PR that need to be removed? Or are these being kept around on purpose?

I'm keeping them around on purpose, because I compare the test output with the expected output in each test case, and those are expected output files that happen to be empty (e.g. we should find no bed regions that don't pass the tests, therefore the expected doNotCalculate file is empty)

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
RavenYL: position, as in the numeric position on the chromosome

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -25,40 +25,78 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
...
+// and species1_ChromStart (the start of the current bed entry)
+func gapBedPass(pos_species1 int, species1_ChromStart int, species1_ChromEnd int, pos_species2 int, species2_ChromStart int, species2_ChromEnd int, gapSizeProductLimit int) (bool, string, string) {
 	pass := true
 	species1_Name := "species1_gap"
 	species2_Name := "species2_gap"
+	// calculate gap size of current bed entry, based on species1_ChromStart
 	species1_gapSize := species1_ChromEnd - species1_ChromStart
```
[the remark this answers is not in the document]
RavenYL: You're right that I am calculating the total bp in the bed entry. In this case, this is the "gap size". This is is part of the gapBedPass function. I pass a bed region that specifies a "gap" into this function. If the gap passes tests, then I write it to the gap output file. Here, gap stands for the regions that are in between "matches" (aka the alignments from the original maf files, which I got from the UCSC Genome Browser, aligning human and chimp chromosomes). I write a bed representing the gap in one of the main functions below (matchToGap), then pass it to this function (gapBedPass) to check if I should actually align it, or if I won't align it (gap_doNotCalculate).

I think you might be thinking of gaps in alignments, aka when one sequence is longer than another. But I'm using "gap" to refer to a different thing, before I even get to aligning sequences

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -25,40 +25,78 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
 	exception.PanicOnErr(err)
 }
 
+// helper function: check if maf entry passes checks
+func matchMafPass(assembly_species1 string, assembly_species2 string, chrom_species1 string, chrom_species2 string, species1_SrcSize int, species2_SrcSize int, species1_ChromStart int, species1_ChromEnd int, species2_ChromStart int, species2_ChromEnd int) bool {
+	pass := true
+
+	// chrom should be the same between species1 and species2 (special case: panTro6 chr2A, chr2B both count as a match to hg38 chr2)
+	pT6_hg38_chr2 := ((assembly_species1 == "hg38" && assembly_species2 == "panTro6" && chrom_species1 == "chr2" && ((chrom_species2 == "chr2A") || (chrom_species2 == "chr2B"))) || (assembly_species1 == "panTro6" && assembly_species2 == "hg38" && ((chrom_species1 == "chr2A") || (chrom_species1 == "chr2B")) && chrom_species2 == "chr2")) // the special case written as a bool variable
```
cshoben, line 33: Looks like you're hard coding in the assemblies. Have you thought about this being user-specified to help this be more generalized?

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -25,40 +25,78 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
...
+	}
+
+	return pass
+}
+
 // helper function: check if gap bed entry passes checks
-func gapBedPass(species1_ChromStart int, species1_ChromEnd int, species2_ChromStart int, species2_ChromEnd int, gapSizeProductLimit int) (bool, string, string) {
+// need both pos_species (most recently updated alignment position)
+// and species1_ChromStart (the start of the current bed entry)
+func gapBedPass(pos_species1 int, species1_ChromStart int, species1_ChromEnd int, pos_species2 int, species2_ChromStart int, species2_ChromEnd int, gapSizeProductLimit int) (bool, string, string) {
```
cshoben, line 51: The old code in this file labels each function as "Step 1..." , "Step 2..." , etc. I liked that it had a flow, but I notice that these new functions are not numbered.

cshoben (dismissed):

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
RavenYL: That's a good idea! But this is the best way I can think of now to address chimp chromosome 2A and 2B matching human chromosome 2, and I believe I can keep this hard-coded because no matter what species the user is trying to align, if they have hg38 and panTro6, they will need this adjustment for the chr2

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
RavenYL: Thanks! That's because these functions are helper functions I call in some of the steps, but they are not steps themselves. For example, gapBedPass is called in Step2 matchToGap

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
craiglowe: I don't think we need it for now, but we could try to think of something more general, like we could consider "chrX" to match "chrXY", as long as X was a number (i.e. one or more digit) and Y was a letter.

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -25,40 +25,78 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
...
+	// define limits and necessary calculations to evaluate limit
+	// gap size product limit should be evaluted based on gapSize
 	gapSizeProduct := species1_gapSize * species2_gapSize
-
-	if species1_gapSize != 0 {
-		gapSizeMultiple = float64(species2_gapSize / species1_gapSize)
+	// gap size multiple should be evaluated based on gapSizeBig
+	gapSizeBigMultiple := 0.0
+	if species1_gapSizeBig != 0 {
+		gapSizeBigMultiple = float64(species2_gapSizeBig / species1_gapSizeBig)
```
craiglowe, line 67: Do we also need to test for species1_gapSizeBig / species2_gapSizeBig?

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -25,40 +25,78 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
...
 	}
+	gapSizeBigMultipleLimit := 100.00 // gapSizeBigMultipleLimit is currently hardcoded tentatively, 100
```
craiglowe, line 69: Is this variable saying that when we have two sequences aligned to each other and one of them is 100 times longer than the other, we will consider the longer one to be an insertion?

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -68,7 +106,7 @@ func mafToMatch(in_maf string, species1 string, species2 string) {
 	mafRecords := maf.Read(in_maf) // read input maf
 
 	// open output files to write line-by-line and create variable for error
-	out_maf_filename := strings.Replace(in_maf, ".maf", ".filtered.maf", 1)
+	out_maf_filename := strings.Replace(in_maf, ".maf", ".filtered.maf", 1) // this means the in_maf path name should not contain .maf, otherwise the string renaming will be triggered in the pathname instead of in the filename
```
craiglowe, line 109: Can we have the user supply this output name?
Instead of using string.Replace, could we check for .maf with strings.HasSuffix (error if not) and then trim the name back three characters and add on filtered.maf?  Then we wouldn't need to worry about what else might be in the name/path

craiglowe (commented):
Wonderful!  I had a couple small comments/questions

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
RavenYL: No, see below

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
RavenYL: No, it is saying that if the currently aligned sequence in species2 is really far away from the previously aligned sequence in species2, 100 times further than the currently aligned sequence in species1 is from the previously aligned sequence in species1, then the current alignment should be discarded and not "trusted" to be a true alignment, because the currently aligned sequence in species2 is probably a random region further downstream on the chromosome, which breaks synteny. This is only a concern for species2 because species1 is always progressing linearly (has synteny) and a big gap in species1 might just mean a big insertion in species1.

RavenYL: IMG_3653

craiglowe (approved):
Great work!

rimangan (approved):

PR #680, opened 2022-05-23
build: dockerfile with small description & small refactor to provide a useful error msg and function re-naming

edotau:
As containerize pods and applications continue to sweep though the software industry the past few years, I think it would be good to keep a docker build file (somewhat similar to a Makefile) to make gonomics more accessible to people who aren't too familiar with setting up golang. Here are some brief instructions on how to create a containerized docker of gonomics:

## Gonomics build

1. Clone the repo if you need to and location the root directory of the gonomics repository:
```
git clone git@github.com:vertgenlab/gonomics.git && cd gonomics
```
2. Build docker container by running the command the the root directory of gonomics:
```
docker build . -t gonomics
```
3. Upon a successful build, every gonomics binary executable should be installed the $GOBIN directory of the docker. To check run this command:
```
docker run gonomics ls /go/bin
```
4. You can also use the golang compiler from inside the container:
```
docker run gonomics go
```
5. (optional) To troubleshoot the docker container, you can open a bash shell which can function as a full linux box:
```
docker run -it gonomics bash
```
Screen Shot 2022-05-23 at 4 12 32 AM

## Fun fact: docker is actually written in go:
Screen Shot 2022-05-24 at 3 59 40 AM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,28 @@
 FROM golang:1.18-buster as builder
 
-# To build docker container image: run the following command:
-#
-# - docker build . -t gonomics
-# 
-# Use gonomics programs with:
-#
-# - docker run gonomics gsw -h
-#
-# All binary executables should exist in /bin if built sucessfully, to check, run:
-# - docker run -it gonomics ls /go/bin
-# 
-
 ENV TZ=America/New_York
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
-
-# Set up working directory and add all golang files to container
-WORKDIR /src/github.com/vertgenlab/gonomics
-ADD . ./
 
 # Update linux image which contains golang-1.18
 RUN set -x && apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
   ca-certificates && \
   rm -rf /var/lib/apt/lists/*
+
+# To build docker container image: run the following command:
+#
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM golang:1.18-buster as builder
+
+ENV TZ=America/New_York
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# Update linux image which contains golang-1.18
```
cshoben, line 6: Does this mean we would need to be on a linux machine to launch/use this docker?

edotau: It's actually the exact opposite.We can essentially create a build from any base image that exists in the registry [hub.docker.com/](https://hub.docker.com/search?q=) I basically run a ubuntu image as my mac and do essentially most of my day to day in linux [ubuntu](https://hub.docker.com/_/ubuntu)

dasnellings (approved):
Very cool! I think this will make gonomics a lot more accessible to people that are unfamiliar with Go. I did not realize that docker was written in Go!

craiglowe (approved):
It will be great to have this as another way for people to use gonomics!

edotau:
I actually think you are missing one thing. You need to add a Makefile to satisfy the make builds crew. Something like:

Makefile:
```
install:
  go install ./...
```

then people can run:
`make` or `make install` 
😎

PR #682, opened 2022-06-13
Bam nil cigar method fix

dasnellings:
Bam stores unmapped read cigars as nil which breaks sam methods that check if a read is unmapped by cig[0] == '*'

This PR adds a check to see if the cigar is nil, then processed as unmapped.

### DIFF
--- a/sam/methods.go
+++ b/sam/methods.go
@@ -15,7 +15,7 @@ func (s Sam) GetChromStart() int {
 
 func (s Sam) GetChromEnd() int {
 	var runLength int = 0
-	if s.Cigar[0].Op == '*' {
+	if s.Cigar == nil || s.Cigar[0].Op == '*' {
 		return s.GetChromStart()
 	}
 	for i := 0; i < len(s.Cigar); i++ {


dasnellings:
Note: I am getting some non-determinism in the chain package. The first time I ran tests it failed, but I reran with no changes and it passed. I don't see how this PR would impact chain, so I think this is a separate problem. someone may want to look into it at some point.

sam/methods.go
```diff
@@ -15,7 +15,7 @@ func (s Sam) GetChromStart() int {
 
 func (s Sam) GetChromEnd() int {
 	var runLength int = 0
-	if s.Cigar[0].Op == '*' {
+	if s.Cigar == nil || s.Cigar[0].Op == '*' {
```
craiglowe, line 18: Could this also be addressed by standardizing the value of s.Cigar when the read is unmapped?  Maybe it should always be a star, or always be nil (less memory allocation)?

craiglowe (commented):
Quick comment if we should standardize the value of c.Cigar instead of handling both possible values.

sam/methods.go
dasnellings: I agree that would be ideal, though I don't have the bandwidth for it at the moment. If someone else can handle it, then I will close this PR.

cfauci (approved):

PR #687, opened 2022-06-21
bam parsing corner case fix

dasnellings:
This PR handles a corner case where the length of the sequence in a bam file is zero.


craiglowe (approved):
This looks good to me.

PR #688, opened 2022-06-22
bedpe package

rimangan:
Package bedpe handles the "bed paired end" file format, in line with the documentation described here:
https://bedtools.readthedocs.io/en/latest/content/general-usage.html.
bedpe files contain "pairs" of genomic regions, which are used in the HiCAR paper to visualize contact loops, as in the  bottom tracks in the screenshot below.
image
This PR gives us core infrastructure for reading and writing these files. The eventual plan here is to develop tools to identify gene targets of genomic regions in bed format, which I plan to pursue in future PRs.


bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,185 @@
+// Package bedPe provides functions for handling Bed Paired End format files, used for
+// handling pairs of genomic regions.
+
+package bedPe
```
craiglowe, line 4: might be better to make this all lowercase:
https://go.dev/blog/package-names

bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/vertgenlab/gonomics/fileio"
+	"io"
+	"log"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+type BedPe struct {
+	Chrom1            string
```
craiglowe, line 19: What do you think about ChromOne or ChromA.  I often have trouble reading code that does math, but also has numerals in the names.

bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,185 @@
...
+	ChromEnd1         int
+	Chrom2            string
+	ChromStart2       int
+	ChromEnd2         int
+	Name              string
+	Score             int
+	Strand1           bed.Strand
+	Strand2           bed.Strand
+	FieldsInitialized int      //number of fields that are initialized, used for smart writing.
+	Annotation        []string //long form for extra fields
```
craiglowe, line 30: Do we need this?  I think we have this in bed as a bit of a hack to make things work, but it might be good to not propagate it if we can.

bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// String converts a BedPe struct to a string so it will be automatically formatted when printing with the fmt package.
+func (b BedPe) String() string {
```
craiglowe, line 34: Should this go in a "methods.go" file?

bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		return out
+	default:
+		log.Fatalf("Error: expecting a request to print at least 6 bedPe fields, but got: %d\n", fields)
+	}
+	return ""
+}
+
+// WriteBed writes an input BedPe struct to an io.Writer.
+func WriteBedPe(file io.Writer, input BedPe) {
```
craiglowe, line 64: I am having trouble telling the difference between WriteBedPe and WriteToFileHandle.  If these are the same, I would keep the second name.

bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,185 @@
...
+	exception.PanicOnErr(err)
+	return answer
+}
+
+//processBedPeLine is a helper function of Read that returns a BedPe struct from an input line of a file.
+func processBedPeLine(line string) BedPe {
+	var err error
+	var start1Num, end1Num, start2Num, end2Num int
+	words := strings.Split(line, "\t")
+	start1Num, err = strconv.Atoi(words[1])
```
craiglowe, line 112: Nothing wrong with this, but I think we have functions that do this conversion and check.  Maybe in "numbers" or "fileio"?

craiglowe (commented):
Looks great!  I had a couple small comments.

bed/bedPe/bedPe.go
rimangan: it's messy, but this is the bedpe standard, additional tab delimited fields at the user's discretion, just like bed.

bed/bedPe/bedPe.go
rimangan: ChromOne in next commit, I was following the bedpe documentation names but i agree they are confusing

dasnellings (dismissed):
This looks good. My only big picture though would be to consider making bedPe an extension of the existing bed data structure. Perhaps something like 
```
type bedPe struct {
A bed
B bed
}
```

It would still need a custom reader/writer, but it would be able to use all the rest of our bed infrastructure. As-is someone would need to copy all the basic bed functions (like IsOverlapping) and reimplement for bedPe. 

Nothing to hold back approval, but it might be worth considering alternative structs.

bed/bedPe/bedPe.go
dasnellings: I would be in favor of replacing the []string in bed and bedPe with an unparsed tab-delimited string. A string here would have a few advantages
1) less heap memory allocation
2) you can do whole-struct comparisons (bedA == bedB)
3) you can use beds as keys in maps (something that I have recently been doing by converting beds to a custom struct that removes the annotation field)

craiglowe: I like Dan's idea here.  I think we should also think about applying it to "bed" (although that would be a separate PR) sometime in the future.  Do we have any code/use-case where keeping it unparsed would be really bad?

craiglowe:
I like the idea of having bedpe be a struct that holds two beds.

bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,173 @@
+// Package bedpe provides functions for handling Bed Paired End format files, used for
+// handling pairs of genomic regions.
+
+package bedpe
```
craiglowe, line 4: Can we also use "git mv" to change the "P" to lowercase in the directory name and file name?

craiglowe (commented):
This looks good to me other than a directory and file renaming.

bed/bedPe/bedPe.go
RavenYL: Are you going to change the bedpe struct and the Annotation []string in this commit? If not, then I am ready to approve when the names of folders and functions are all fixed.

bed/bedPe/bedPe.go
craiglowe: I think we need a couple commands to do this name change:
cd bed
mkdir bedpe
git mv bedPe/bedPe.go bedpe/bedpe.go

bed/bedPe/bedPe.go
```diff
@@ -0,0 +1,169 @@
...
+			current.B.Annotation = append(current.B.Annotation, words[i])
+		}
+	}
+
+	return current
+}
+
+// NextBedPe returns a BedPe struct from an input fileio.EasyReader. Returns a bool that is true when the reader is done.
+func NextBedPe(reader *fileio.EasyReader) (BedPe, bool) {
+	line, done := fileio.EasyNextLine(reader)
```
craiglowe, line 138: This will not allow comment lines in the file, but changing to "EasyNextRealLine" will allow comment lines

bed/bedPe/bedPe_test.go
```diff
@@ -0,0 +1,84 @@
+package bedpe
```
craiglowe, line 1: This file looks like it will also need a rename/mv

craiglowe (commented):
I had a comment about the Read function not using the NextBedPe function.  I also think we need a few files to be moved/renamed and then we are all set!

craiglowe (approved):
Wonderful!

PR #689, opened 2022-06-22
chimpBiasedMleRecon

rimangan:
primateRecon is a program that takes a 5-way alignment of great apes and reconstructs ancestral sequences. Currently, when we are uncertain about the identity of the base at the Human-Chimpanzee Ancestor, we can either assign it to the human by default (-mleHumanBiased) or to N by default (-mleUnbiased). Both options are useful in distinct cases. 
This PR adds a third option, (-mleChimpBiased), which as its name suggests assigns the HCA to the chimpanzee allele by default when we are uncertain about the true identity of the HCA base.
This option replaces the previous program "chimpAncestorRecon", which has been deleted in this request.


cmd/primateRecon/primateRecon.go
```diff
@@ -130,7 +160,7 @@ func primateReconMle(inFastaFilename string, inTreeFilename string, humanBias bo
 		log.Fatalf("Didn't find hca in the tree\n")
 	}
 
-	if !humanBias {
+	if !(humanBias || chimpBias) {
 		humanAltNode = expandedTree.FindNodeName(tree, "hg38alt")
```
craiglowe, line 164: Probably not needed, but we could think of doing something similar with variation we observe in chimps

craiglowe (approved):
Wonderful!

cmd/primateRecon/primateRecon.go
RavenYL: Is this saying that if there is no humanBias or chimpBias, then the uncertain base in the HCA will become the hg38Alt? Is it biologically true that hg38Alt is not as good at predicting HCA as human or chimp Ref alleles?

craiglowe: This is adding an additional node to the tree that is a "second human genome."  In the case of an A/G SNP in humans, we can make two human genomes (where one is A and one is G) so that we don't arbitrarily bias towards one of the alleles for a segregating site.

craiglowe (commented):
I included a response to a comment.

cmd/primateRecon/primateRecon.go
RavenYL: OK I see!

RavenYL (approved):

PR #690, opened 2022-06-25
bedPad

rimangan:
Picking up where Dan left off on this, bedFormat in this pr supports padding entries by extending chromStart and chromEnd. Requires a chromSize file to ensure the padded bed output doesn't run off the end of the chromosome.
A smarter, but more involved solution would be to truncate pads with a noGap.bed file, any thoughts on this?


dasnellings (approved):
Looks good. I have no objection to a noGap.bed, but I would not like to have it as a requirement.

cmd/bedFormat/bedFormat.go
```diff
@@ -6,23 +6,40 @@ import (
...
+	if s.ChromSizeFile == "" && s.PadLength > 0 {
+		log.Fatalf("Must specify a chromFile to use the padLength option.")
+	}
+	if s.ChromSizeFile != "" && s.PadLength > 0 {
+		sizes = chromInfo.ReadToMap(s.ChromSizeFile)
+	}
 
 	for v := range ch {
+		if s.PadLength > 0 {
+			if _, inMap = sizes[v.Chrom]; !inMap {
```
craiglowe, line 37: Not a big deal, but we are throwing away the value on this line with the underscore, but then a couple lines later we are using the value to set v.ChromEnd.  We could have as single lookup by saving the value on this line.

craiglowe (approved):
Wonderful!

craiglowe:
I can imagine restricting to a noGap file being important for an analysis, but I think just using the chrom sizes file will fit most use cases.

PR #691, opened 2022-06-28
bam qual fix

dasnellings:
Fixes panic on case where length of seq/qual is 0


craiglowe (approved):
This looks like a good fix to me.  I am guessing these are bam files with only sequence and no base quality that you are running into?

dasnellings:
Actually I have some bams that have no sequence info nor any quality data. I was doing some (admittedly slapdash) fastq -> bam processing and hard trimming barcodes from the sequences and some barcode-barcode ligated reads were entirely trimmed. I should really have a check to remove those reads, but I figure gonomics probably shouldn't panic if that does happen.

rimangan (approved):

PR #693, opened 2022-07-05
Parse bam extra field to text

dasnellings:
Parses binary extra field from bam readin to text version for editing.


craiglowe (approved):
This looks good to me.

PR #694, opened 2022-07-05
intervalContacts

rimangan:
intervalContacts takes a bedpe file (specifying a set of contacting genomic regions, as in HI-C or HICAR results) and an interval format file (specifying variants or genomic regions of interest to query) and returns a bed format file of the contact sites corresponding to all intersections between the queried interval file and the bedpe file. 

Useful for identifying putative target genes informed by chromatin conformation for a set of genomic regions.


cmd/intervalContacts/intervalContacts.go
```diff
@@ -0,0 +1,67 @@
...
+			bed.WriteBed(out, i.A) // and vice versa
+		}
+	}
+
+	err = out.Close()
+	exception.PanicOnErr(err)
+}
+
+func usage() {
+	fmt.Print("intervalContacts - Returns all regions that contact input genomic regions. " +
```
cshoben, line 45: I think it would be useful to list the types of files gonomics can support with our interval package. I see based on test files we can handle .bed and .vcf. Are there other interval file types we can handle too?

cmd/intervalContacts/intervalContacts.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for i := range inChan {
+		inIntervals = append(inIntervals, i)
+	}
+	inTree := interval.BuildTree(inIntervals)
+
+	for _, i := range contactRecords {
+		currOverlaps = interval.Query(inTree, i.A, "any")
+		if len(currOverlaps) > 0 {
+			bed.WriteBed(out, i.B) // if anything in the input set overlaps i.A, write i.B to the output.
```
cshoben, line 32: Why do we write/output bed B when we have an overlap with bed A?

cshoben (approved):

cmd/intervalContacts/intervalContacts.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		currOverlaps = interval.Query(inTree, i.B, "any")
+		if len(currOverlaps) > 0 {
+			bed.WriteBed(out, i.A) // and vice versa
```
RavenYL, line 36: Can I ask why you chose to write the interacting (other) bed region in the bedpe as the output, and not include the overlapping bed region in the bedpe or the entry in the interval inFile in the output? Is this best format for downstream applications? I guess you can grep for lines in the output in the bedpe and interval inFile with the line number to find all 3 regions (bedpe overlapping, bedpe interacting, inFile interval entry). I am curious because I might want to use this program to find interacting regions for UHIs as well.

cmd/intervalContacts/intervalContacts.go
```diff
@@ -0,0 +1,67 @@
...
+	contactRecords := bedpe.Read(bedpeFile)
+	inChan := interval.GoReadToChan(inFile)
+
+	for i := range inChan {
+		inIntervals = append(inIntervals, i)
+	}
+	inTree := interval.BuildTree(inIntervals)
+
+	for _, i := range contactRecords {
+		currOverlaps = interval.Query(inTree, i.A, "any")
```
RavenYL, line 30: I was reading interval.Query but I couldn't figure out if all corner cases would be accounted for here. For example, bed is [closed, open) so the last base is not part of the entry and inFile intervals that only overlap the end of a bed should not count. This particular case might not matter so much though since it's only 1 base. Do you think interval.Query has been giving you results that look real and reliable?

RavenYL: In relation to this, I wonder if it will be helpful to include another field in the output file indicating "how much the inFile interval overlaps with the bedpe entry", whether it is 1 base out of 100, or 50 bases out of 100, etc.

cmd/intervalContacts/intervalContacts.go
craiglowe, line 45: Maybe a little more info here?  Something like, "if an input interval overlaps a bedpe region, the pair to that overlapping bedpe region will be sent to the output.

RavenYL (approved):
I am approving but I have some questions

craiglowe (approved):
Great!

cmd/intervalContacts/intervalContacts.go
RavenYL: I forget if you were there when Riley discussed his last pull request on bedpe, but I learned from there that since the bedpe files report interacting regions in the genome (e.g. Hi-C), we want to use the bedpe files to see what the regions the HAQERs are interacting with, meaning if a HAQER overlaps with A, then we are interested in what B is because B is the target of A aka the target of the HAQER which might be an enhancer

PR #695, opened 2022-07-06
gorilla-biased HGA reconstruction with primateRecon

rimangan:
Adds a new option to primateRecon to perform a gorilla-assembly biased reconstruction of the human-gorilla ancestor (HGA) using the same maximum likelihood framework as the HCA reconstruction. Works on the same human-chimp-bonobo-gorilla-orangutan alignment.


craiglowe (approved):
Wonderful!

cmd/primateRecon/primateRecon.go
```diff
@@ -111,30 +83,47 @@ func hcaIsPresent(human, bonobo, chimp, gorilla, organutan dna.Base) bool {
 	return false
 }
 
+func hgaIsPresent(human, bonobo, chimp, gorilla, orangutan dna.Base) bool {
+	if baseIsPresent(gorilla) && (baseIsPresent(human) || baseIsPresent(chimp) || baseIsPresent(bonobo)) {
```
RavenYL, line 87: The if conditions in hgaIsPresent are different than the way the if conditions are formulated in hcaIsPresent, but if I am understanding correctly this corresponds to the position of gorilla in the human-chimp-gorilla-bonobo-orangutan phylogeny tree and so is normal?

cmd/primateRecon/primateRecon.go
```diff
@@ -111,30 +83,47 @@ func hcaIsPresent(human, bonobo, chimp, gorilla, organutan dna.Base) bool {
...
+func reconHgaBase(root, gorillaNode, nodeToRecon *expandedTree.ETree, position int, probThreshold float64, nonBiasProbThreshold float64) {
+	var likelihoods []float64
+	var nextBase dna.Base
+	reconstruct.SetState(root, position)
+	likelihoods = reconstruct.FixFc(root, nodeToRecon)
+	nextBase = likelihoodsToBaseBias(likelihoods, gorillaNode.Fasta.Seq[position], probThreshold, nonBiasProbThreshold)
+	nodeToRecon.Fasta.Seq = append(nodeToRecon.Fasta.Seq, nextBase)
+}
+
+func primateReconHcaMle(inFastaFilename string, inTreeFilename string, humanBias bool, chimpBias bool, probThreshold float64, nonHumanProbThreshold float64, outputFastaFilename string) {
```
RavenYL, line 120: I noticed that primateReconHcaMle and primateReconHgaMle have different numbers of input variables, since the Hga version does not have humanBias bool or chimpBias bool. I wonder if having the same number of input variables will be easier to write in the program, in case you expand to other types of biased common ancestor reconstruction? This way, maybe you won't have to write as many checks in the main function to determine if the input variables are legal and point them to the correct primateReconMle function. This is just an idea since your current style already works

cmd/primateRecon/primateRecon.go
rimangan: yes that's the intuition

cmd/primateRecon/primateRecon.go
rimangan: yes we would want that if we ever decided to make a human-biased HGA estimate. For now, I only wrote the options I needed for the tasks at hand

PR #696, opened 2022-07-07
intervalSubsetMatrix

rimangan:
This one is sort of niche, i'll have a readme coming soon. 

Essentially, I have about 200 bed files corresponding to open chromatin sites from different biological tissues. If I run cat *.bed.gz | bedMerge, I find that the union set of elements (all elements that occur in at least one bed file) contains 1.2 million open chromatin sites.
Here, I want to make a binary matrix of dimensions RxC, where the rows R are the 1.2 million chromatin sites and the columns C are the 200 bed files, with values of 1 if that element is open in the queried biological tissue, and 0 otherwise.


cmd/intervalSubsetMatrix/intervalSubsetMatrix.go
```diff
@@ -0,0 +1,97 @@
...
+	var currOverlaps []interval.Interval
+	var currIntervalChan <-chan interval.Interval
+	var unionIntervals = make([]interval.Interval, 0)
+	var currLineString string
+
+	unionRecChan := interval.GoReadToChan(unionFile)
+	files := fileio.Read(fileListFile)
+	out := fileio.EasyCreate(outFile)
+
+	for curr := range unionRecChan {
```
craiglowe, line 26: Should we make a simple "Read" function for intervals at some point in the future?

cmd/intervalSubsetMatrix/intervalSubsetMatrix.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, u := range unionIntervals {
+		mat[interval.CoordsToString(u)] = make([]int, len(files)) //initialize all rows to zeros based on the number of files. Map is indexed by coords of regions.
+	}
+
+	for i = range files {
+		currIntervalChan = interval.GoReadToChan(files[i])
+		for j = range currIntervalChan {
+			currOverlaps = interval.Query(unionTree, j, "any")
+			if len(currOverlaps) > 0 {
+				mat[interval.CoordsToString(currOverlaps[0])][i] = 1
```
craiglowe, line 41: Do you need to loop through currOverlaps and mark them all as one?  I guess not because unionTree elements are defined to be a merge of all elements in "files."  Maybe there is a more general solution here?

cmd/intervalSubsetMatrix/intervalSubsetMatrix.go
```diff
@@ -0,0 +1,97 @@
...
+		headerString = fmt.Sprintf("%s\t%s", headerString, files[k])
+	}
+	_, err = fmt.Fprintln(out, headerString)
+	exception.PanicOnErr(err)
+
+	keys := make([]string, 0, len(mat))
+	for k := range mat {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys) //jusst alphanumeric sort, might change to coordinate sort, though order doesn't matter.
```
craiglowe, line 56: jusst -> just
Coordinate-based sort might be nice.

craiglowe (approved):
Great!

RavenYL (dismissed):

PR #697, opened 2022-07-09
Bed merge bug

rimangan:
Small bug in bedMerge (highMem version), where a duplicate of the first record in the file appears in the output if you run on an initially unsorted file. The intended behavior is that we first presort the input set of bed entries and then set the first value to the currentMax. In main, we set the records[0] to currentMax, and then sort, so our initial currentMax is not the first entry in the bed file, causing a duplicate of the first entry to appear when running the high mem algorithm on an unsorted input file.


cmd/bedMerge/bedMerge.go
```diff
@@ -52,9 +52,9 @@ func bedMergeLowMem(infile string, outfile string, mergeAdjacent bool) {
 func bedMergeHighMem(infile string, outfile string, mergeAdjacent bool) {
 	var records []bed.Bed = bed.Read(infile)
 	var outList []bed.Bed
-	var currentMax bed.Bed = records[0]
 
 	bed.SortByCoord(records)
+	var currentMax bed.Bed = records[0]
```
RavenYL, line 57: I approve this change but it seems like all the other changes are from the bedSubsetMatrix pull request. Should I not approve this (bedMergeBug) pull request now and wait until after bedSubsetMatrix has been merged?

cfauci (approved):

RavenYL (approved):

craiglowe (approved):
Great catch!

PR #699, opened 2022-07-15
bedNoGapFix

rimangan:
faFormat produces nogap.bed files from a given fasta format sequence file. For some downstream applications, it is useful to have a unique "Name" for each bed entry, so I've set the Name field for nogap.bed files to be Chrom_Start_End. This can always be removed with "cut" to produce the old 3 field beds.

### DIFF
--- a/cmd/faFormat/testdata/expected.NoGap.bed
+++ b/cmd/faFormat/testdata/expected.NoGap.bed
@@ -1,6 +1,6 @@
-InputOutput	0	4
-TrimName Test	0	4
-ToUpperTest	0	4
-NoGapTest	0	3
-NoGapBedTest	0	3
+InputOutput	0	4	InputOutput_0_4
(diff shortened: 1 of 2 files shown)


cmd/faFormat/testdata/expected.NoGap.bed
```diff
@@ -3,4 +3,4 @@ TrimName Test	0	4
 ToUpperTest	0	4
 NoGapTest	0	3
 NoGapBedTest	0	3
```
RavenYL, line 5: Only "NoGapBedTest 6 10" has the unique Name field, but not "NoGapBedTest 0 3", "NoGapTest 0 3" etc. Is this expected behavior?

RavenYL (approved):

cfauci (approved):

craiglowe (approved):
Wonderful!

PR #702, opened 2022-07-26
Update README.md installation instructions

RavenYL:
- From Dan's comment "Just a heads up, the install directions for gonomics are out of date. go get no longer works. I think we should advise people to clone the repo and run go install ./…. I may work on updating it later in the week, but if others are motivated then have at it."
- It looks like if we clone the repo and run go install, then the "user" and "contributor" instructions become the same. So I merged them.
- I also thought we need to run go mod tidy to make sure our dependencies are updated? I can delete that line if it's not necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,6 @@ This should print to screen a line for each test with "ok" printed in the left m
 </ul>
 
     go install ./...
-    go mod tidy
 <ul>
 This will tidy up the necessary modules
 </ul>
--- a/README.md
+++ b/README.md
@@ -144,27 +144,7 @@ Goal is to  get gonomics into the “src” directory of go. This path needs to 
 
 <br>
 
-*Workspace set up for basic gonomics installation (most users):*
-<ul>
-<ul>
-
-    cd $GOPATH/src
-    go get
-</ul>
-this should install gonomics and all its dependencies.
-<ul>
-
-    cd /github.com/vertgenlab/gonomics/ 
-    go test ./…
-</ul>
-This should print to screen a line for each test with ok printed in the left margin when something passes
-</ul>
-</ul>
-<br>
-<br>
-<br>
-
-*Gonomics with version control (for contributors of gonomics):*
+*Gonomics installation with version control (for both users and contributors of gonomics):*
 <ul>
 
     cd $GOPATH/src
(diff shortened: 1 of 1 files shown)


craiglowe (commented):
Thank you!  Good to get this documentation updated!  I don't think we need the "go mod tidy", but I am not absolutely certain.  Anyone know for sure?  I could do a clean setup on my computer to test.

README.md
```diff
@@ -144,27 +144,7 @@ Goal is to  get gonomics into the “src” directory of go. This path needs to
...
-</ul>
-<br>
-<br>
-<br>
-
-*Gonomics with version control (for contributors of gonomics):*
+*Gonomics installation with version control (for both users and contributors of gonomics):*
 <ul>
 
     cd $GOPATH/src
```
dasnellings, line 150: Go also modified the way that GOPATH works. I believe it is no longer required that code be inside GOPATH so I think we could delete the entire section about setting GOBIN, GOROOT, and GOPATH and tell the user to just run `git clone` wherever they like. Someone should double check that though before changing. May be something for a future PR.

README.md
```diff
@@ -182,6 +162,7 @@ This should print to screen a line for each test with "ok" printed in the left m
 </ul>
 
     go install ./...
+    go mod tidy
```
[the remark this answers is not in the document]
dasnellings: Yeah, this is only necessary for developers.

README.md
RavenYL: I think we can do this, because when I was walking through installing go and gonomics with Evon on 2022-04-14, we didn't do the GOBIN, GOROOT, GOPATH stuff. We just downloaded go from the website, and I had Evon mkdir GOPATH/src/github.com/vertgenlab/gonomics and do git clone. I will make some changes and others can comment on if they are approprite or not

cfauci (dismissed):

README.md
```diff
@@ -45,131 +45,15 @@ Authors:
...
-
-*Gonomics with version control (for contributors of gonomics):*
-<ul>
-
-    cd $GOPATH/src
-    mkdir -p /github.com/vertgenlab/
-    cd github.com/vertgenlab
+    
+    cd /pathWhereYouWantGonomicsCloned/
+    mkdir -p /src/github.com/vertgenlab
```
craiglowe, line 55: I don't think we want the first "/" before src

craiglowe (dismissed):
Looks great!  One quick edit request.

cfauci (approved):

craiglowe (approved):
Great!

PR #704, opened 2022-08-04
faFormat Rev

rimangan:
While faFormat has a "revComp" option, it did not have a simple "rev" option. Added here with testing.

### DIFF
--- a/dna/modify.go
+++ b/dna/modify.go
@@ -106,6 +106,13 @@ func ComplementSingleBase(b Base) Base {
 	}
 }
 
+//Reverse an input sequence of bases. Does not maintain 5' -> 3' orientation.
+func Reverse(bases []Base) {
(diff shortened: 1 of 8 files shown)


dna/modify.go
```diff
@@ -106,6 +106,13 @@ func ComplementSingleBase(b Base) Base {
 	}
 }
 
+//Reverse an input sequence of bases. Does not maintain 5' -> 3' orientation.
+func Reverse(bases []Base) {
+	for i, j := 0, len(bases)-1; i <= j; i, j = i+1, j-1 {
```
cshoben, line 111: Could you please help me understand this logic in person?

cshoben (approved):
Looks great! I had a question about the logic with the reverse code in the DNA package.

RavenYL (approved):
Approving but not merging because I saw some lingering questions in the PR

PR #707, opened 2022-08-07
small changes, want to chat before writing testing for this

cshoben:
I noticed this didn't have testing and I felt like it could be a quick thing for me to do. However, after looking at this code more I'm wondering if we should remove this for now? 

I was unable to find anything online to support the formula being used. I tried to find an online formula but I was unsuccessful. I tested this code using a protocol I found online : See Protocol on **Ampure Bead Double DNA Size Selection** https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3849550/#:~:text=A%20double%20(upper%20and%20lower,the%20larger%20fragments%20are%20discarded
where the goal was a range of 400-600bp. The output from this code was close to the protocol. However, if I run another example I found with double size selection using Spriselect (https://research.fredhutch.org/content/dam/stripe/hahn/methods/mol_biol/SPRIselect%20User%20Guide.pdf) I am getting negative volumes of beads to add which tells me that this formula isn't fully hashed out. I have also now read about ratios being sensitive to the buffers used and this cmd does not specify which protocol it was originally written for.


RavenYL:
I remember learning from Christi that this command was maybe written by Eric some time ago and may not be used a lot anymore. This is the resource Seth and I have been using to determine AMPure bead volume: https://www.beckman.com/reagents/genomic/cleanup-and-size-selection/pcr/performance

cfauci:
Yes, this was written by Eric a long time ago and I've used it in a pinch when I don't want to look up the protocol, but it is generally not useful because the math absolutely doesn't work for any number I've put in. I suspect that Eric made it work for small lengths, where you wanted to deplete for above a certain size and below a certain size, rather than get rid of anything below X base pairs

RavenYL (commented):
I feel like it might be more efficient to remove this cmd and use existing protocols, rather than trying to figure out a math model to fix/improve the cmd. But that's my personal opinion and we should discuss this at code review

rimangan:
I also vote to remove this cmd

RavenYL (approved):

rimangan (approved):

cfauci (approved):

PR #708, opened 2022-08-08
divergenceSpectrum

rimangan:
This program takes in a bed file specifying a set of genomic regions and a vcf specifying a set of variants (in this case, intended to represent divergent sites, which can be made from a pairwise alignment with multiFaToVcf).

For each bed region, we report the counts for six types of mutations and store those counts in the annotations fields.
(0: A->G/T->C) (1: G->A/C->T) (2: A->T/T->A) (3: G->C/C->G) (4: A->C/T->G) (5: C->A/G->T).

Restricted to biallelic SNPs, will fatal otherwise. In other words, we expect the input vcf to be pre-filtered.


cmd/divergenceSpectrum/divergenceSpectrum.go
```diff
@@ -0,0 +1,128 @@
...
+			i.Annotation[k] = fmt.Sprint(currSpectrum[k])
+		}
+		i.FieldsInitialized = 13
+		bed.WriteBed(out, i)
+	}
+	err = out.Close()
+	exception.PanicOnErr(err)
+}
+
+// mutationType returns an int corresponding to six categories of mutations. (0: A->G/T->C) (1: G->A/C->T) (2: A->T/T->A) (3: G->C/C->G) (4: A->C/T->G) (5: C->A/G->T)
```
RavenYL, line 48: Would it be helpful to include info on what the numbers 0-5 mean in the usage message?

RavenYL (approved):

cmd/divergenceSpectrum/divergenceSpectrum.go
cfauci: I agree, the usage statement should have this list in it

cfauci (changes requested):
add details for which mutations correspond to 0=5 in usage

dasnellings:
Not sure if it is relevant for your use case, but in a future version you might consider recording the mutations by their trinucleotide context rather than just the mutated site itself (see https://cancer.sanger.ac.uk/signatures/sbs/). This is commonly done in the somatic mutation field and it is common for sources of mutations to have highly specific contexts. It also helps control for noise from some of the more common mutation sources (read CpG transitions) so you can see more subtle differences in other contexts. 
Screen Shot 2022-08-17 at 4 28 43 PM

cfauci (approved):

PR #710, opened 2022-08-13
vcfFilter minDaf/MaxDaf

rimangan:
New options on vcfFilter enable the user to filter a set of vcf variants by minimum or maximum derived allele frequency.
Critically, ancestral alleles must be annotated for these options to work (perhaps we could do a minorAlleleFrequency based filter if that is desirable in the future, but not for this use case). 
Also, this will fail if the program detects non-segregating sites in the file (where all the sample haplotypes have the same allele).
Passes tests.

### DIFF
--- a/cmd/vcfFilter/vcfFilter.go
+++ b/cmd/vcfFilter/vcfFilter.go
@@ -215,6 +215,9 @@ func getTests(c criteria, header vcf.Header) testingFuncs {
 	}
 
 	if c.minDaf != 0 {
+		if c.minDaf < 0 || c.minDaf > 1 {
+			log.Fatalf("minDaf must be between 0 and 1.")
+		}
 		answer = append(answer,
 			func(v vcf.Vcf) bool {
 				return popgen.VcfSampleDerivedAlleleFrequency(v) > c.minDaf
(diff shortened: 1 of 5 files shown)


cmd/vcfFilter/vcfFilter.go
```diff
@@ -212,6 +214,20 @@ func getTests(c criteria, header vcf.Header) testingFuncs {
...
+			})
+	}
+
+	if c.maxDaf != 1 {
+		answer = append(answer,
+			func(v vcf.Vcf) bool {
+				return popgen.VcfSampleDerivedAlleleFrequency(v) < c.maxDaf
+			})
+	}
+
```
RavenYL, line 230: Is there a check for maxDaf > minDaf, and have an error if false? Would that be necessary, or would you allow maxDaf < minDaf in some cases?

rimangan: fixed in next commit

RavenYL (approved):

cshoben (commented):
I'm thinking there should be a check to ensure that the ancestral allele is defined. I see it is listed in the INFO field. I think right now the code only checks that sites are segregating?

cfauci (approved):

craiglowe (approved):
Great!

cshoben (approved):
Thanks for the updates!

PR #713, opened 2022-08-22
faUniq

shaesimpson:
Ready for review


cmd/faUniq/faUniq_test.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+	"github.com/vertgenlab/gonomics/exception"
+	"github.com/vertgenlab/gonomics/fileio"
+	"os"
+	"testing"
+)
+
+inputFile := test_in.fa
```
rimangan, line 10: You'lll need to envelope test_in.fa in quotation marks so golang can interpret this as a string.
Also, these lines should be placed inside the TestFaUniq function in order to be visible to the faUniq function call

rimangan (approved):

cmd/faUniq/faUniq.go
```diff
@@ -0,0 +1,63 @@
...
+	var seqExists bool
+	var val int
+	for i := 1; i < len(allSeqs); i++ {
+		currSeq = dna.BasesToString(allSeqs[i].Seq)
+		val, seqExists = seqsSeen[currSeq]
+		if !seqExists {
+			seqsSeen[currSeq] = len(uniqueSeqs)
+			uniqueSeqs = append(uniqueSeqs, allSeqs[i])
+		} else {
+			uniqueSeqs[val].Name = uniqueSeqs[val].Name + "; " + allSeqs[i].Name
```
RavenYL, line 30: I like that in the output fa seq name you include all the names of the same seq in the input fa!

RavenYL (approved):

cfauci (approved):

PR #715, opened 2022-09-02
Create bins

cfauci:
created fasta/bin.go and binFasta option to faFilter. These will allow a user to specify a number of bins for the input to be broken up into and name the outputs as <given outputFileName.#.fa> so that the names are easily sorted or handled downstream. Added testing for both the package and command, all tests are passing.


fasta/bin.go
```diff
@@ -0,0 +1,142 @@
...
+	//TODO: chance that there's a situation where the final bin with leftOvers is empty or very small, what I have with baseCap,
+	//	or chance that the last bin could be almost double the size of everything else (totalBases/BinNum and remainder goes in last bin)
+	//	could do a check to see if remainder is super large and then handle as necessary
+	equalBinNum = binNum - 1
+	remainder = totalBases % binNum
+	if remainder == 0 {
+		baseCap = totalBases / binNum
+	} else {
+		splitBases = totalBases - remainder //number of bases to split between equalBins
+		baseCap = splitBases / equalBinNum
```
RavenYL, line 31: Can you help me understand how the bins are "relatively equal"? Judging from baseCap, does this mean each bin has about the same number of bases (as opposed to about the same number of fasta sequences)? Why is "relatively equal" important?

cmd/faFilter/faFilter_test.go
```diff
@@ -11,20 +12,27 @@ var FaFilterTests = []struct {
...
 	minSize      int
+	binFasta     int
+	keepWhole    bool
 }{
-	{"testdata/minSizeTest.fa", "testdata/minSizeOutput.fa", "testdata/minSizeExpected.fa", "", "", false, 0, -1, 10},
+	{"testdata/minSizeTest.fa", "testdata/minSizeOutput.fa", "testdata/minSizeExpected.fa", "testdata/binExpected.fa", "", "", false, 0, -1, 10, 2, true},
+}
+var binOutputs = []string{
+	"testdata/minSizeOutput.0.fa",
+	"testdata/minSizeOutput.1.fa",
```
RavenYL, line 29: Am I understanding correctly that the output after using the binFasta option should be multiple fasta files, but those multiple output fasta files are just not in cmd/faFilter/testdata because you compare them to the existing 1 expected output fasta file (binExpected.fa) and remove them?

cmd/faFilter/faFilter.go
```diff
@@ -63,6 +77,8 @@ func main() {
 	var name *string = flag.String("name", "", "Specifies the fasta record name.")
 	var notName *string = flag.String("notName", "", "Returns all fasta records except for this input.")
 	var minSize *int = flag.Int("minSize", 0, "Retains all fasta records with a sequence of at least that size")
+	var binFasta *int = flag.Int("binFasta", 0, "Makes specified number of bins from given fasta where all output files will contain similar amounts of sequence.")
+	var keepWhole *bool = flag.Bool("keepWhole", false, "Keep the output of faFilter before and after binning.")
```
RavenYL, line 81: Can you help me understand what keepWhole does? Does it interact with the binFasta option?

RavenYL (approved):

cmd/faFilter/faFilter.go
cshoben: It looks like "keepWhole" flag being true is the same as "binFasta" = 0. I think this flag could be dropped. Unless there intended to be a role for keepWhole when binFasta != 0?

cshoben (commented):
partially reviewed. Submitting current comments.

cfauci:
This is going to get an overhaul, but I wanted to address comments anyway.

cmd/faFilter/faFilter.go
cfauci: thats exactly what it's for chelsea, I wanted to give an option to keep whatever filtered version of the genome was made before being broken into the bins, it's something I needed to keep the testing the way it was written, but also something I saw a use for down the line.

cmd/faFilter/faFilter_test.go
cfauci: yes, they are compared to make sure they are being written properly and then removed since they are produced by the test run, we try to remove any files that are produced in this way so that we aren't creating a ton of extra stuff any time we test everything

fasta/bin.go
cfauci: the last bin won't be equal to the others, and depending on the math could be much smaller or nearly the same size. The bins other than the last one will all have exactly equal number of bases.

PR #717, opened 2022-09-09
Local alignment with AffineGap

dasnellings:
This PR adds AffineGapLocal to the align package. It functions identically to AffineGap_highMem except it does not penalize gaps at the start or end of the alignments. This is useful for when you are trying to align a subset of a sequence to a larger sequence (e.g. a sequencing read to a ref genome).


RavenYL (approved):
Dan and I have already discussed this but just to put it in writing: AffineGapLocal is a version of AffineGap_highMem because it does not need AlignGap (lowMem). AffinaGapLocal is for aligning short read sequences to ref genomes; AffineGap (lowMem) is for aligning 2 large chromosomes and at that size is when lowMem helps by breaking up the 2 sequences into checkerboards of 10kb fragments

craiglowe (approved):
Wonderful!

PR #719, opened 2022-09-09
AffineGap struct recycling

dasnellings:
This PR builds on my previous PR adding a function to start an AffineGap "engine" that can accept new inputs to be aligned without needing to reallocate all the data structures. This saves a LOT of memory turnover when executing a large number of small alignments.


align/affineGap_highMem.go
```diff
@@ -25,6 +25,34 @@ func initAffineScoringAndTrace(firstSeqLen int, secondSeqLen int) ([][]int64, []
...
+		mRowCurrent[k] = setSliceSize(mRowCurrent[k], secondSeqLen+1)
+		mRowPrevious[k] = setSliceSize(mRowPrevious[k], secondSeqLen+1)
+		for i := range trace[0] {
+			trace[k][i] = setSliceSize(trace[k][i], secondSeqLen+1)
+		}
+	}
+	return mRowCurrent, mRowPrevious, mColumn, trace
+}
+
+func setSliceSize[t any](s []t, size int) []t {
```
craiglowe, line 44: This is neat and might come in handy in other places.  I am not sure of a more general package where we could put it (and people would find it), but maybe someone else has an idea?

align/affineGap_highMem.go
```diff
@@ -66,23 +94,113 @@ func expandCigarRunLength(route []Cigar, chunkSize int64) {
...
+}
+
+type TargetQueryPair struct {
+	Target []dna.Base
+	Query  []dna.Base
+	Score  int64
+	Cigar  []Cigar
+}
+
+func GoAffineGapLocalEngine(scores [][]int64, gapOpen int64, gapExtend int64) (inputs chan<- TargetQueryPair, outputs <-chan TargetQueryPair) {
```
craiglowe, line 113: Not needed, but it would be fun to have a benchmark for this

craiglowe (commented):
This all looks good to me, but it would help me if we could merge in the already approved changes to this PR.  I also think it would be good to have a test of GoAffineGapLocalEngine or the function it calls (which I don't think is tested right now).

cfauci (approved):

rimangan (approved):

PR #720, opened 2022-09-09
bam seeking

dasnellings:
This PR adds support for random access of BAM files using index files.


craiglowe (commented):
I am not sure about the bai file specs, but I did my best to review this and I don't notice any bugs.  Would it be possible to add more than one test, because I (and others?) am having trouble giving a thorough review and maybe a couple more tests would be a good alternative to show correctness even in some corner cases?

dasnellings:
I added a few more tests. I had trouble coming up with many corner cases. Let me know if anyone has ideas for more corner cases to test.

cshoben (approved):

cfauci (approved):

PR #721, opened 2022-09-14
first draft of hiCarOverlap. Which sorts hiCarContact data based on a…

sweavs111:
…n input gene list

### DIFF
--- a/cmd/hiCarOverlap/hiCarOverlap.go
+++ b/cmd/hiCarOverlap/hiCarOverlap.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/exception"
+	"github.com/vertgenlab/gonomics/fileio"
+	"log"
+	"strings"
+)
+
+func hiCarOverlap(geneList string, hiCarData string, outFile string) {
+	var hiCarPairs []string = fileio.Read(hiCarData)
+	var genes []string = fileio.Read(geneList)
+	var err error
+	out := fileio.EasyCreate(outFile)
+
+	hiCarGeneMap := make(map[string]int)
+
+	for i, line := range genes {
+		hiCarGeneMap[line] = i
+	}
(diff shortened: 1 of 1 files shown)


cmd/hiCarOverlap/hiCarOverlap.go
```diff
@@ -0,0 +1,60 @@
...
+		hiCarGeneMap[line] = i
+	}
+	for _, hiCarCont := range hiCarPairs {
+		items := strings.Split(hiCarCont, "\t")
+		_, ok := hiCarGeneMap[items[9]]
+		_, ok2 := hiCarGeneMap[items[10]]
+		if ok || ok2 {
+			_, err = fmt.Fprintf(out, hiCarCont+"\n")
+			exception.PanicOnErr(err)
+		}
```
cfauci, line 30: i recommend an else fatalf here if !ok or !ok2 and it should report what each of them is so you dont have to track back which element is missing without extra context.

cmd/hiCarOverlap/hiCarOverlap.go
```diff
@@ -0,0 +1,60 @@
+package main
+
```
cfauci, line 2: should we discuss adding hiCar as a struct and giving it a package?

cmd/hiCarOverlap/hiCarOverlap.go
```diff
@@ -0,0 +1,60 @@
...
+		log.Fatalf("Error: expecting %d arguments, but got %d\n",
+			expectedNumArgs, len(flag.Args()))
+	}
+
+	geneList := flag.Arg(0)
+	hiCarData := flag.Arg(1)
+	outFile := flag.Arg(2)
+
+	hiCarOverlap(geneList, hiCarData, outFile)
+}
```
cfauci, line 60: just a note to make sure you add testing.

cfauci (commented):
looks good, but i'd definitely like to hear you describe it. I think this could also benefit from some documentation, since most of us haven't been working with HiCar data thus far.

PR #722, opened 2022-09-16
Fixed bug in SeekBamRegion

dasnellings:
I had been indexing by `i` when I should have indexed by `bins[i]` in old lines 29/30. It made seek work a bit harder to find reads as it would start further back than it needed to.


craiglowe (approved):
Good catch!

cshoben (approved):

PR #723, opened 2022-09-20
SeekBamRegion multiple reads returned

dasnellings:
Chelsea caught a bug where SeekBamRegion would return the same read multiple times. This PR makes a temporary fix that should ensure correctness, but future modifications will will be needed.


craiglowe (approved):
Thank you for the fix!

RavenYL (approved):

PR #724, opened 2022-09-27
Binning

cfauci:
Binning has functionality for a few things:
1: it can break up a fasta with multiple records into bins of relatively equal sizes
2: it can break up a fasta with multiple records into fasta files of a minimum number of bases
3: it can break up a single fasta record into bins of relatively equal sizes (tho this part of the package does not currently have a command, it is tested and functional.)


bin/bin.go
```diff
@@ -0,0 +1,250 @@
+package bin
```
rimangan, line 1: "bin" is a bit of a reserved word in cs. I worry that users looking at a bin package may expect it to handle binary information. If everything in here handles fastas, can this go in that package? 
I worry someone looking for a tool to bin fastas might want a more descriptive name like faBin and it will be easier to find that way

cmd/bin/bin.go
```diff
@@ -0,0 +1,64 @@
+package main
```
rimangan, line 1: same as above, i would consider "faBin" or "binGene" as the filename if possible. Otherwise we'll have this weird construction of ~/go/bin/bin

bin/bin.go
cfauci: that's a great point, I can go with something like faBin for sure for the command and change the package name. I was originally going to put it in the fasta package, but when Craig and I spoke we waffled a bit on that point, happy to put it there if that's the preferred home for it

cmd/bin/bin.go
```diff
@@ -0,0 +1,64 @@
...
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/bin"
+	"github.com/vertgenlab/gonomics/fasta"
+	"log"
+)
+
+func binGenome(genome string, path string, binNum int, minSize int) {
+	records := fasta.Read(genome)
+	var bins map[int][]fasta.Fasta
```
RavenYL, line 13: I'm trying to think through if the map data type works perfectly for bins. If I remember correctly, maps are not ordered, but that shouldn't matter here, right?

cfauci: correct, it shouldn't matter here, I'm actually numbering them as a place holder for a key anyway, so they kind of correspond to the order they are created, but in general it doesn't matter at all for this kind of function where I'm depending on the name of a contig to be access from the value anyway.

PR #725, opened 2022-09-28
added newline after usage in intervalContacts.go

sweavs111:


cmd/bedPeOverlap/bedPeOverlap.go
```diff
@@ -0,0 +1,80 @@
...
+	var err error
+
+	out := fileio.EasyCreate(contactOutFile)
+
+	contactRecords := bedpe.Read(bedpeSelectFile)
+
+	inBedPe := bedpe.Read(bedpeInFile)
+
+	for _, i := range inBedPe {
+		left, right := bedpe.SplitBedPe(i)
```
rimangan, line 28: i would declare left and right before the for loop. A line like 'var left, right bedpe.BedPeHalf' and then change the colon equals to just an equals. This will be a slight runtime optimization, as you won't need to reallocate memory at each iteration of the loop

bed/bedpe/bedpe.go
```diff
@@ -20,6 +20,13 @@ type BedPe struct {
 	B bed.Bed
 }
 
+type BedPeHalf struct {
+	Chrom      string
+	ChromStart int
+	ChromEnd   int
+	Home       *BedPe
```
RavenYL, line 27: Can you help me understand why Home is a *BedPe as opposed to a BedPe? Is it so that the pointer can point to the correct value regardless of variable name?

cmd/bedPeOverlap/bedPeOverlap.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+
+	err = out.Close()
+	exception.PanicOnErr(err)
+
+}
+
+func usage() {
+	fmt.Print("bedpeOverlap - Returns all bedpe entries that fully overlap another bedpe entry. " +
```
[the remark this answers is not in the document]
RavenYL: "Returns all bedpe entries that describe the same contact as another bedpe entry (the 2 bedpe entries describe contacts that overlap)"?

bed/bedpe/bedpe.go
sweavs111: The pointer is so that each half of the bedPe can be assigned back to their parent bedPe after overlaps

RavenYL (approved):

rimangan (approved):

PR #727, opened 2022-10-04
Binning

cfauci:
Passing all tests and working on real-world genomes. This command can:
1. bin fastas into a specified number of bins, filling the smallest bin with a fasta record
2. bin fastas to meet a size minimum in each output file
3. allows an option so that when using standard in the output file will contain the name of the assembly


RavenYL (approved):

fasta/faBin.go
```diff
@@ -0,0 +1,250 @@
...
+import (
+	"github.com/vertgenlab/gonomics/dna"
+	"log"
+)
+
+//BinGenomeNoBreaks takes in an entire genome which is sorted largest to smallest contig and breaks up the fasta so
+//that smaller contigs get combined into a single fasta, while large contigs become a single fasta on their own. The user
+//must specify the number of bins for the genome to be broken into, the genome must have more contigs than bins in order
+//to combine any contigs and equal number bins to contigs if each contig gets its own record. The bins will all be
+//filled with the first contig encountered when it's empty, and then the smallest of those bins will be filled when the
```
cshoben, line 12: Suggestion for language change... 
instead of "smallest bin" calling it "the bin with the smallest contig"
Can talk more at code review.

fasta/faBin.go
```diff
@@ -0,0 +1,250 @@
...
+
+	for i := len(bins); i < len(genome); i++ {
+		b := findSmallestBin(bins)
+		bins[b] = append(bins[b], genome[i])
+	}
+
+	return bins
+}
+
+//findSmallestBin finds the bin with the least sequence and returns which bin it is so that it can be filled with the next contig.
```
cshoben, line 49: More language...
Instead of "least sequence" could this be called the "smallest contig"?
Can also talk more about this in code review.

fasta/faBin.go
```diff
@@ -0,0 +1,250 @@
...
+func findSmallestBin(bins map[int][]Fasta) int {
+	var smallest int
+	var sizeSmallest int
+	for i := 0; i < len(bins); i++ {
+		rec, exists := bins[i]
+		if !exists {
+			log.Panic("Map was not filled properly in BinGenomeNoBreaks.")
+		}
+		size := calcNumBasesInBin(rec)
+		if size < sizeSmallest {
```
cshoben, line 59: What if the last two bins have the same size contig? (corner case of course...)
I think it would still be added to the "last" bin but want to double check that corner case doesn't break the code.

fasta/faBin_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/vertgenlab/gonomics/dna"
+	"log"
+	"testing"
+)
+
+var (
+	genome1     = "testdata/testBreakBinning.fa"
+	expBases1   = dna.StringToBases("ACGTA")
+	expBases2   = dna.StringToBases("CGT")
+	expBases3   = dna.StringToBases("AC")
```
cshoben, line 13: Can you help me understand why the third bin would have 'AC' as the bases and not ''ACGTA'?
Actually, I expected different functionality for all the bins. Maybe we can talk more in code review about this as well.

cshoben (commented):
Looks great overall! Very clean and easy to follow. Only part way through reviewing, happy to talk more.

fasta/faBin.go
cfauci: This is a good suggestion, but it can contain more than just a single contig it will take as many as it needs to so its not the smallest contig necessarily

fasta/faBin.go
cfauci: so long as everything is binned, it doesnt' matter if the last bin is larger than some other bin

rimangan (approved):

PR #729, opened 2022-10-14
bedPeOverlap bedSelect

rimangan:
Adds an option to bedPeOverlap to use a bed as a select file. 
Here, bedPe entries in the input file are retained in the output if at least one end of the bedPe entry overlaps one of the bed entries in the select file.
Passes tests.


rimangan:
image
A bit of data to accompany this PR. Here is an example where I had an overwhelming amount of bedpe contacts and I wanted to filter then to just the bedpe contacts that overlap regions of interst (in this case, the HAQER peak in the center of the image).

cmd/bedPeOverlap/bedPeOverlap.go
```diff
@@ -51,18 +90,20 @@ func bedpeOverlap(bedpeSelectFile string, bedpeInFile string, contactOutFile str
...
+		"Default behavior expects a bedpe select file and returns entries where both ends of a bedpe entry from the input file" +
+		"overlap both ends of a bedpe entry from the select file.\n" +
 		"Usage:\n" +
-		"	bedpeOverlap [options] selectFile.bedpe inFile.bedPe out.bedpe\n\n")
+		"bedpeOverlap [options] selectFile inputFile.bedPe out.bedpe\n\n")
 	flag.PrintDefaults()
 }
 
 func main() {
+	var bedSelect *bool = flag.Bool("bedSelect", false, "Set select file to be a BED file instead of a bedpe.")
```
RavenYL, line 105: Would it be appropriate to add that "Set select file to be a BED file instead of a bedpe. Returns entries where either end of a bedpe entry from the input file overlaps a bed entry from the select file"?

cfauci (approved):

craiglowe (approved):
Great work!

PR #730, opened 2022-10-18
Mixture model

dasnellings:
This PR adds a gaussian mixture model algorithm to the numbers package. It takes a 1d array of data and tries to fit the data to k number of normal distributions. It does with with an iterative approach called the expectation-maximization algorithm where we initialize a model with random normal distributions, determine how well the data can be explained by the current model (the expectation step), then adjust the model and see if it performs better. It continues to iterate until the model is no longer improving. At which points it gives you the means, weights, and standard deviations for the gaussians. Notably this does not always give the correct answer as it can get stuck in local maximums which do not recapitulate the true dataset. It is somewhat involved so I'd be happy to join a call for the next code review to discuss. I also found this page helpful to conceptualize the idea https://aishack.in/tutorials/expectation-maximization-gaussian-mixture-model-mixtures/

### DIFF
--- a/numbers/mixtureModel.go
+++ b/numbers/mixtureModel.go
@@ -1,12 +1,10 @@
 package numbers
 
 import (
+	"log"
 	"math"
+	"math/rand"
 )
-
-// logProbEpsilon is the arbitrary set point for convergence. When the model likelihood is
-// increasing by < logProbEpsilon, we say the model has converged, possibly to a local minimum.
-const logProbEpsilon = 1e-08
 
 // MixtureModel holds data, results, and working memory for running the EM algorithm.
 type MixtureModel struct {
@@ -23,6 +21,10 @@ type MixtureModel struct {
 	lamSigRatio    []float64   // holds the ratio of Weights to Stdev
 	logLamSigRatio []float64   // holds the natural log of lamSigRatio
 	work           []float64   // holds intermediate values for calculating LogLikelihood
+
+	// logProbEpsilon is the arbitrary set point for convergence. When the model likelihood is
+	// increasing by < logProbEpsilon, we say the model has converged, possibly to a local minimum.
+	logProbEpsilon float64
 }
 
 // RunMixtureModel1D uses the expectation-maximization (EM) algorithm to find a mixture of k gaussian distributions that fit the input data slice.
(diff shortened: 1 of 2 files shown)


RavenYL:
Would you be available during our code review on 10/31 (Mon) 12:30pm-1:30pm? This coming week 10/24 (Mon) we won't have code review because Craig and Riley are traveling to ASHG.

dasnellings:
Yes, that works for me.

numbers/mixtureModel.go
```diff
@@ -0,0 +1,256 @@
...
+		mm.logLamSigRatio = make([]float64, k)
+	}
+	if cap(mm.work) >= k {
+		mm.work = mm.work[0:k]
+	} else {
+		mm.work = make([]float64, k)
+	}
+
+	// TODO smarter initial guess for mean and variance (k-means/PCA)
+	for i := range mm.Means {
```
rimangan, line 121: Most k-means implementations (which is very similar to Gaussian mixture models) pick the location of random data points as the starting means.

rimangan (dismissed):
Great implementation! Could we specify that this is a 1d mixture model somewhere?

numbers/mixtureModel.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+// logProbEpsilon is the arbitrary set point for convergence. When the model likelihood is
+// increasing by < logProbEpsilon, we say the model has converged, possibly to a local minimum.
+const logProbEpsilon = 1e-08
+
+// MixtureModel holds data, results, and working memory for running the EM algorithm.
+type MixtureModel struct {
+	Data           []float64   // 1d data slice
+	K              int         // number of component distributions
```
RavenYL, line 15: Is k a manual input, aka the user decides how many component distributions there are?

numbers/mixtureModel.go
dasnellings: Good idea. Added

numbers/mixtureModel.go
dasnellings: That is correct. You need to have some prior information about how many components you are expecting. In my case, I am using it for genotyping and I assume that there are 2 copies of the autosomes or 1 of X/Y

rimangan (approved):
LGTM!

craiglowe (commented):
Awesome!  I made a few small comments as I went through.  Sorry for the delay!

numbers/mixtureModel.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	if cap(mm.work) >= k {
+		mm.work = mm.work[0:k]
+	} else {
+		mm.work = make([]float64, k)
+	}
+
+	// TODO smarter initial guess for mean and variance (k-means/PCA)
+	for i := range mm.Means {
+		mm.Means[i] = data[RandIntInRange(0, len(data)-1)]
```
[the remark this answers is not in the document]
dasnellings: That is a good point. I changed it to sample w/o replacement.

numbers/mixtureModel.go
```diff
@@ -0,0 +1,255 @@
...
+			}
+		}
+		/* At this stage, work contains the normal density at data[i]
+		   divided by the normal density with the largest st'dized resid
+		   Thus, dividing by rowsum gives the posteriors: */
+		for j = 0; j < mm.K; j++ {
+			mm.posteriors[j][i] = mm.work[j] / rowsum
+		}
+		/* Finally, adjust the loglikelihood correctly */
+		mm.LogLikelihood += math.Log(rowsum) - min + mm.logLamSigRatio[minj]
```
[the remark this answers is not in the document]
dasnellings: That would be my guess. This was copied from c code and I admit I don't fully understand the math behind it. Can you recommend the functions you would use?

rimangan (approved):

cfauci (approved):

PR #731, opened 2022-10-22
3dGREAT Part 1: Laying the foundation

rimangan:
This PR, authored by me, Christi, and Raven, introduces a few key functions we will need for 3dGREAT:

1. We move parts of bedMerge (specifically the HighMem implementation) from the bedMerge command to the bed package so we can use this technique inside of other programs.
2. GenesToPromoterBed parses TSS sites from a gtf file and returns a bed of promoter regions (where the user specifies the number of bases upstream and downstream of a TSS as the promoter range).
3. A special case of GenesToPromoterBed is GenesToTssBed, which makes a bed file of 1 base-pair bed entries corresponding to the location of all TSSs in the input GTF.


cfauci (approved):

craiglowe (commented):
Looks great!  I included a couple comments about the commented out code and also had one comment about empty lists

cfauci (approved):

craiglowe (approved):
Excellent!

PR #735, opened 2022-10-27
Raven alignAnchor gorgor5

RavenYL:
- I am updating the program after implementing it on hg38 vs gorGor5, in addition to the hg38 vs panTro6 I did previously 
- I made chrMap.txt a necessary input, to indicate how species1 and species2 chr names match up. This accounts for assemblies with different chr names (e.g. human chr1 corresponds to gorilla CYUI01014905v1), and special cases (e.g. human chr2 corresponds to chimpanzee chr2A and chr2B)
- I also made diagonal an option. By default, I only allow maf entries that are roughly along the genomic coordinates diagonal to pass, but now there is an option to allow even non-diagonal maf entries to pass my checks. This option works better for hg38 vs gorGor5


cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -24,21 +24,57 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
...
+func makeChrMap(chrMap_filename string) map[string][]string {
+
+	chrMap := make(map[string][]string) // map to hold species1 species2 chr name matches
+	var chrMap_stringSplit []string
+	var exists bool
+
+	chrMap_string := fileio.Read(chrMap_filename) // read file so each line is 1 string
+
+	for i := range chrMap_string {
+		chrMap_stringSplit = strings.Split(chrMap_string[i], "\t") // convert each line = 1 string into a slice of 2 strings (assumption: each line only has 2 columns separated by 1 tab). Note that Notepad generates proper tab, but not Atom
```
craiglowe, line 37: You may want to make sure that this slice a length of 2

cmd/globalAlignmentAnchor/globalAlignmentAnchor.go
```diff
@@ -24,21 +24,57 @@ func writeToFileHandle(file io.Writer, species1 bed.Bed, species2 bed.Bed, score
...
+		if !exists {
+			chrMap_species2 := make([]string, 1)       // slice to hold just 1 species2 chr name the first time its species1 chr name appears
+			chrMap_species2[0] = chrMap_stringSplit[1] // use species2 chr name to populate slice of length 1 in preparation for [string][]string key-value pair. This slice will be overwritten each time there is a new species2 chr name
+			//TODO: Can I define chrMap_species2 []string outside of for loop? The below line doesn't work. Because of this, if I define chrMap_species2 []string outside of for loop, I can't use "copy" to prevent the chrMap map from always pointing to the chrMap_species2 []string when chrMap_species2 gets updated
+			//copy(chrMap[chrMap_stringSplit[0]], chrMap_species2)
+			chrMap[chrMap_stringSplit[0]] = chrMap_species2
+		} else {
+			chrMap[chrMap_stringSplit[0]] = append(chrMap[chrMap_stringSplit[0]], chrMap_stringSplit[1]) // if species1 chr name already exists, append to the exisitng species2 chr name slice the new species2 chr name
```
craiglowe, line 46: I am not positive, but I think you can append to nil in go.  Because of this property, I think you may not need this "if" statement and you can just have this line handle all cases (even when the map is empty).

craiglowe (commented):
Looks great!  I had two quick comments that are not bugs, but one is an additional check and the other may shorten your code a little.

rimangan (approved):

craiglowe (approved):
Great!

cfauci (approved):

PR #737, opened 2022-11-05
wigMath

rimangan:
wigMath is a command to perform mathematical operations on wig format files.
Here are the operation supported:
First, wigMath supports scalar multipliction/division of wig data and element-wise addition and subtraction between two wigs. (useful for basic visualization of for example differential ATAC between samples using elementWiseSubtract. Another composite operation is to average several wig files by taking their sum and dividing by the number of wig files).
Second, the program can perform moving average smoothing on noisy input wig data. Seen below is a comparison of two tracks, the bottom is raw wig data with stochastic noise, and the top is the same data smoothed with a 20bp sliding window.
image

Third, the program can calculate the element-wise absolute error between two wigs A and B to make wig C, where C_i = abs(A_i - B_i). Also supports calculations of element-wise absolute percent error with the form C_i = 100% * abs((A_i-B_i)/A_i) s.t. A_i != 0.
Fourth, 'Pearson' allows the user to feed in two wig files and calculate the Pearson Correlation Coefficient between them to evaluate the degree of similarity in the two wig tracks. This program excludes values matching a specified 'missing' value float. The option 'samplingFrequency' allows the user to sample a random subset of bases in each wig to evalute the PCC. This is an important step for computational tractability.


cfauci (approved):

cmd/wigMath/wigMath.go
```diff
@@ -0,0 +1,197 @@
...
+		second = wig.Read(s.Pearson)
+		answer := wig.Pearson(records, second, s.Missing, s.SamplingFrequency)
+		out := fileio.EasyCreate(s.OutFile)
+		_, err = fmt.Fprintf(out, "PCC:\t%f\n", answer)
+		err = out.Close()
+		exception.PanicOnErr(err)
+	}
+}
+
+func multipleOptionCheck(s Settings) {
```
RavenYL, line 91: Maybe this could have been coded with a case switch loop, where each option is a case and the default is log.Fatalf? But this is just a stylistic comment

wig/smooth.go
```diff
@@ -0,0 +1,28 @@
...
+			sum += w.Values[k]
+		}
+		midPoint = (j + j + windowSize) / 2
+		answer.Values[midPoint] = sum / float64(windowSize)
+	}
+	return answer
+}
+
+// SmoothSlice performs moving average smoothing on an input slice of Wig structs using a user-specified windowSize.
+func SmoothSlice(w []Wig, windowSize int) []Wig {
```
RavenYL, line 22: I want to check that I understand Smooth and SmoothSlice. In cmd/wigMath, the input wig file becomes "records []Wig". SmoothSlice takes each wig in the []Wig, which is the wig block with a header line, and calls Smooth to assign the average wig value within a windowSize to the midpoint position of each window. Is this right?

Originally I was going to ask "Why is the midPoint = (j + j+ windowSize)/2, as opposed to something like midPoint=j+windowSize/2?" But then I realized that those 2 expressions are the same

RavenYL: Now there will be a change in the program about missing data.
-For operations like add/multiplication, missing data in an input wig will cause output to be missing data
-For smoothing, missing data will also cause output to be missing data, as opposed to a smoothed over average where answer.Values[midPoint] = sum / float64(windowSize-numberOfPositionsWithMissingData)

RavenYL (changes requested):

cmd/wigMath/wigMath_test.go
```diff
@@ -0,0 +1,228 @@
...
+		AbsoluteError:          "",
+		AbsolutePercentError:   "",
+		Missing:                -10,
+		Pearson:                "testdata/second.wig",
+		SamplingFrequency:      1,
+	},
+	{
+		InFile:                 "testdata/in.wig",
+		OutFile:                "testdata/tmp.Mult55.wig",
+		ExpectedFile:           "testdata/expected.Mult55.wig",
```
RavenYL, line 133: The actual file name is testdata/expected.mult55.wig so the capitalization of "M" here caused a mismatch and that's why go test can't find it

cfauci (approved):

RavenYL (approved):

PR #738, opened 2022-11-07
3dGREAT Part 2: Filling the space

rimangan:
Here we take the gene data condensed into a single point for each TSS as a bed and fill the whole genome with it's nearest gene record. We then take a bedpe and use the midpoint of a contact foot as a single point bed entry which contains the distance to the closest TSS (whichever region it fell in after the genome was filled with nearest gene records) as well as the gene name of that TSS. The foot of a contact with a lowest distance becomes the gene and distance contained at both feet of that contact. We then have a bed file that contains a bunch of single points, the TSS single points from earlier combined with the bedpe feet single points which contain hidden values that list how far that point is from either its or its partners nearest gene. we then check if the whole genome has the logically nearest gene considering neighboring bedpe feet and gene entries and remove any records which don't take up any space in the genome for one reason or another. Then we fill the space of the whole genome such that the whole genome is filled with bed entries with names that correspond to their nearest gene in 3d space. 

One problem: currently the output isn't really anything unless you use the nearestGene option. We should discuss the output and its function before merging.


bed/modify.go
```diff
@@ -70,3 +71,176 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+			currAnswer.ChromEnd = records[i].ChromEnd
+			currAnswer.Name = records[i].Name
+			currAnswer.Score = records[i].ChromStart
+		} else {
+			midpoint = (records[i].ChromStart + currAnswer.ChromEnd) / 2
+			currAnswer.ChromEnd = midpoint + 1
+			answer = append(answer, currAnswer)
+			currAnswer.Chrom = records[i].Chrom
+			currAnswer.ChromStart = midpoint + 1
+			currAnswer.ChromEnd = records[i].ChromEnd
```
craiglowe, line 152: Based on the last few lines, I think things might get funny here (we end up with invalid beds) if more than one TSS comes in with the same location (which will probably happen).  There is a test for that or is that something we know we handle?

bed/modify.go
```diff
@@ -70,3 +71,176 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+			out = append(out, records[i])
+		}
+	}
+	return out
+}
+
+// FillSpaceNoHiddenValue accepts a bed and reference genome as a map[string]chromInfo.ChromInfo and returns a bed
+// that assigns each genomic position to the nearest feature in the input bed using the Name field in the input bed.
+// Absolute start positions of the original input bed entries are stored in the Score field of the output bed entries.
+func FillSpaceNoHiddenValue(records []Bed, genome map[string]chromInfo.ChromInfo) []Bed {
```
craiglowe, line 125: This does not seem like a general function for modifying beds, but rather something more specific to great, so I don't think bed/modify.go is the correct place.  Probably the 3dgreat folder is where this should be

bed/modify.go
```diff
@@ -70,3 +71,176 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+	answer = append(answer, currAnswer)
+
+	return answer
+}
+
+//helper fucntion of FillSpaceHiddenValue. Define the 3dMidpoint between bed entries A and B (with hidden values
+//Ha and Hb) as (A.ChromEnd-Ha + B.ChromStart+Hb) / 2 s.t. A.ChromEnd < B.ChromStart.
+// if 3dMidpoint(A, B) < A.ChromEnd, then entry A can never have an entry in the output bed of FillSpaceHiddenValue,
+// and should be removed.
+func removeBedsWithNoTerritory(records []Bed) ([]Bed, bool) {
```
craiglowe, line 168: This should probably be in the cmd/3dgreat folder and not bed/modify.go

bed/modify.go
```diff
@@ -70,3 +71,176 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+			answer = append(answer, records[i-1])
+		}
+	}
+	answer = append(answer, records[len(records)-1])
+
+	return answer, violation
+}
+
+//runUntilNoNewHidden directs our mergeKeepLowNameAndScore function to keep updating until no changes occur
+func runUntilNoNewHidden(records []Bed) []Bed {
```
craiglowe, line 190: This should probably be in the cmd/3dgreat folder and not bed/modify.go

bed/modify.go
```diff
@@ -70,3 +71,176 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+	if newHidden {
+		runUntilNoNewHidden(records)
+	}
+	return records
+}
+
+// FillSpaceHiddenValue accepts a slice of Bed structs and a reference genome as a map[string]chromInfo.ChromInfo and returns a
+// slice of Bed structs that assigns each genomic position to the nearest feature in 3D space from the input bed, using
+// the input bed scores to represent "hidden values", or the distance from that position to its nearest TSS in 3D space.
+func FillSpaceHiddenValue(records []Bed, genome map[string]chromInfo.ChromInfo) []Bed {
```
craiglowe, line 202: This should probably be in the cmd/3dgreat folder and not bed/modify.go

bed/bedpe/modify.go
```diff
@@ -0,0 +1,21 @@
+package bedpe
+
+import "github.com/vertgenlab/gonomics/bed"
+
+//contactToMidpoint returns a bedpe where the contact has been collapsed to its midpoint
+func contactToMidpoint(bp BedPe) BedPe {
```
craiglowe, line 6: Do we need to allocate memory and make new ones, or can we just overwrite the values?  Form a quick glance at the code that calls this function, it doesn't seem like we go back to the original.

bed/bedpe/modify.go
```diff
@@ -0,0 +1,21 @@
...
+	aMidpoint := (bp.A.ChromStart + bp.A.ChromEnd) / 2
+	bMidpoint := (bp.B.ChromStart + bp.B.ChromEnd) / 2
+	A := bed.Bed{Chrom: bp.A.Chrom, ChromStart: aMidpoint, ChromEnd: aMidpoint + 1, FieldsInitialized: bp.A.FieldsInitialized}
+	B := bed.Bed{Chrom: bp.B.Chrom, ChromStart: bMidpoint, ChromEnd: bMidpoint + 1, FieldsInitialized: bp.A.FieldsInitialized}
+	return BedPe{A, B}
+}
+
+//contactsToMidpoints returns a bedpe where the contacts have been collapsed to their midpoints
+func contactsToMidpoints(bps []BedPe) []BedPe {
```
craiglowe, line 15: Do we need to allocate memory and make new ones, or can we just overwrite the values?  Form a quick glance at the code that calls this function, it doesn't seem like we go back to the original.

bed/compare.go
```diff
@@ -63,7 +63,7 @@ func Overlap(alpha Bed, beta Bed) bool {
 //OverlapCount returns the number of elements from list one that have any overlap with list two. Answers range from 0 to len(a).
 //Input bed slices must be presorted with SortByCoord.
 func OverlapCount(a []Bed, b []Bed) int {
-	var count int = 0
+	var count = 0
```
craiglowe, line 66: why do we remove the explicit type?

bed/compare.go
```diff
@@ -82,7 +82,7 @@ func OverlapCount(a []Bed, b []Bed) int {
 //OverlapLengthSum calculates the total number of overlapping bases between two sets of bed elements.
 //Input bed slices must be presorted with SortByCoord
 func OverlapLengthSum(a []Bed, b []Bed) int {
-	var sum int = 0
+	var sum = 0
```
craiglowe, line 85: why do we remove the explicit type?

bed/modify.go
```diff
@@ -54,7 +55,7 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
 		return records //empty and nil slices are returned as is.
 	}
 	SortByCoord(records)
-	var currentMax Bed = records[0]
+	var currentMax = records[0]
```
craiglowe, line 58: why do we remove the explicit type?

craiglowe (commented):
great work!  I took a first pass through this and the core looks really good.  My two high-level comments are that I think we have some functions in the bed package, that are not general functions for manipulating any bed struct, but are specific to 3dgreat, therefore should be located in the cmd/great3d folder.  The other is that I think it will help me (and others) review the code if there are more comments about the intended output of functions.  I find myself stuck in a cycle where I am reading the code to know what the intended outputs are, but then it is hard to find cases where a bug causes the code to produce output that does not match the intended output.

bed/modify.go
RavenYL, line 190: Note to self: review the runUntilNoNewHidden function

bed/modify.go
```diff
@@ -70,3 +71,176 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
 	outList = append(outList, currentMax)
 	return outList
 }
+
```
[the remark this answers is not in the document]
RavenYL: I don't see a function named newHidden around here anymore. Maybe these comments have been addressed?

bed/modify.go
RavenYL: If I remember correctly, the input to this function should already have been filtered such that no more than one TSS has the same location?

bed/modify.go
```diff
@@ -92,3 +92,178 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+	var outList = make([]Bed, 0)
+	SortByCoord(records)
+	var curr = records[0]
+	for i := 1; i < len(records); i++ {
+		if Overlap(curr, records[i]) {
+			if records[i].Score < curr.Score {
+				curr = records[i]
+			}
+		} else if curr.Chrom == records[i].Chrom {
+			dist = records[i].ChromStart - curr.ChromEnd
```
RavenYL, line 108: Note to self: I think this is where we "use nearest wormhole when it's nearer than bedpe's either foot's nearest gene"
curr is records[i] if it has the lower score
curr.ChromEnd=3 (the last base occupied by curr is 2), records[i].ChromStart=5 (the first base occupied by records[i] is 5). dist = 5-3 = 2 (the distance is length 2, aka base positions 3 and 4)

bed/modify.go
```diff
@@ -70,3 +71,176 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+		} else {
+			outList = append(outList, curr)
+			curr = records[i]
+		}
+	}
+	outList = append(outList, curr)
+	return outList, newHidden
+}
+
+func removeRecordsOnEmptyChrom(records []Bed, genome map[string]chromInfo.ChromInfo) []Bed {
```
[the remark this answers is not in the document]
RavenYL: This looks like "remove records that are on chromosomes that don't exist in the genome"

bed/modify.go
```diff
@@ -92,3 +92,178 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+			currAnswer.ChromEnd = genome[records[i-1].Chrom].Size
+			if currAnswer.ChromEnd < currAnswer.ChromStart {
+				log.Print(records[i-1])
+				log.Print(currAnswer)
+				log.Fatalf("Died on new chrom.")
+			}
+			answer = append(answer, currAnswer)
+			currAnswer.Chrom = records[i].Chrom
+			currAnswer.ChromStart = 0
+		} else if currAnswer.Name == records[i].Name && currAnswer.Chrom == records[i].Chrom {
```
RavenYL, line 245: Is `currAnswer.Chrom == records[i].Chrom` here unnecessary since it's implied in the else if?

Am I right to think that this is the case for "multiple points describing the same gene, possibly TSS + contact point, or contact point + contact point". In this case, expand the gene's bed to be the most inclusive (smallest chromStart, largest chromEnd, lowest score because that entire region can be considered that smallest distance to TSS)? If so, then I think already "remove the contact point / bedpe foot that is closer to the TSS to begin with" (the point Craig mentioned at lab meeting 6/14).

bed/modify.go
```diff
@@ -92,3 +92,178 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
...
+			currAnswer.Score = numbers.Min(currAnswer.Score, records[i].Score)
+		} else {
+			threeDMidpoint = (currAnswer.ChromEnd - records[i-1].Score + records[i].ChromStart + records[i].Score) / 2
+			currAnswer.ChromEnd = threeDMidpoint + 1
+			currAnswer.Name = records[i-1].Name
+			if currAnswer.ChromEnd-currAnswer.ChromStart < 0 {
+				log.Fatalf("Died in loop.")
+			}
+			answer = append(answer, currAnswer)
+			currAnswer.ChromStart = threeDMidpoint + 1
```
RavenYL, line 257: Note to self: note how currAnswer.ChromStart is set to midpoint (after writing out previous currAnswer entry to output file) so that the next currAnswer entry starts at the midpoint (and can end at the TSS of the next gene, or a calculated 3d midpoint for another distal gene based on the closest wormhole, but both cases are covered by the "else" aka threeDMidpoint case)

bed/modify.go
rimangan: This is fine but will need documentation

bed/bedpe/great3d_test.go
```diff
@@ -0,0 +1,17 @@
+package bedpe
+
```
cfauci, line 2: this is currently incomplete

cfauci (approved):

RavenYL (approved):

PR #741, opened 2022-11-21
multiFaToAxt

rimangan:
Simple conversion program to turn a chromosome-wide pairwise multiFa file into an AXT.
In my use case, I have a pairwise multiFa alignment of each human chromosome and the corresponding chromosome of the HCA. I want to make an AXT representation of this alignment in order to make a chain file, such that I can perform liftOver between human and HCA coordinates.


rimangan:
Could someone confirm if Rend and Qend in the AXT file format refers to alignment position, rather than reference positions? It seems that axtChain checks the number of characters in the sequence and compares it to Rend - Rstart

craiglowe:
My understanding is that the start and end are 1-based inclusive (e.g. closed) intervals.  The number is based on the reference genome, not the alignment.  It is also worth noting that I believe if the strand is minus, then the second set of intervals is from the reverse complement of the reference genome for that species.  Maybe my understanding if off because I am not sure what Rend-Rstart would give.  I think Rend-Rstart+1 would be the number of non-gap characters in the alignment for that species.

PR #747, opened 2022-12-13
small changes to lastZWriter

cfauci:
nothing particularly interesting, was just looking for a bug, but it wasn't in any of this


rimangan:
looks like there are a couple of empty files

cfauci:
> looks like there are a couple of empty files

I'm really not sure what's going on there because if you try to edit the file its not empty and it isn't empty on any push because testing would immediately fail, which it hasn't here. If you view the file it isn't empty, so I really don't know why it says that.

RavenYL (approved):

rimangan (approved):

PR #748, opened 2022-12-15
pileup terminal insertion fix

dasnellings:
There is an infrequent bug that I am running into with the pileup code that this PR fixes. If there are multiple reads with the same start position AND some (but not all of the reads) begin with an insertion pileup will panic because it tries to access the base before the start position which may not exist depending on the start position on previous reads. To fix this, this PR converts insertions at the start or end of reads to a soft clip. I think that this is the correct way to handle reads with terminal insertions as they are generally not informative.


cfauci (approved):

edotau (approved):
lgtm!

PR #751, opened 2023-01-03
tfMatchComp

rimangan:
tfMatchComp is a program that takes in a multiple sequence alignment in multiFa format, scans both the target and query sequences for TFBS motif occurrences, and provides the user with information on where sequence differences between the target and query have produced differences in motif matches. The output file is a bed with three annotation fields: the first is the motif score for the target, the second is the motif score for the query, and the third is the difference in motif score between the target and query. 
While this program is capable of scanning whole chromosomes, it performs an exhaustive enumerative search, so it is probably advisable to perform on sub-chromosome regions of interest.
Everything passes tests, but I would like to discuss the program in code review before we merge so we can discuss what other features and use cases we would like to see available for this program.
Dependent on the motif package in my last PR.


motif/match.go
```diff
@@ -10,30 +10,94 @@ import (
...
+			log.Fatalf("Error scoring consensus sequence.")
+		}
+		currCutoff = propMatch * currConsensusScore
+		currReverseMotif = ReverseComplement(motif)
 
+		for alnPos := range records[0].Seq {
+			if dna.DefineBase(records[0].Seq[alnPos]) {
+				currRefScore, couldScoreRefWindow = ScoreWindow(motif, records[0].Seq, alnPos)
+				currAltScore, couldScoreAltWindow = ScoreWindow(motif, records[1].Seq, alnPos)
+				if couldScoreRefWindow && couldScoreAltWindow && (currRefScore > currCutoff || currAltScore > currCutoff) {
```
craiglowe, line 44: Minor detail, but I usually think of cutoffs as being inclusive ranges.

motif/match.go
```diff
@@ -10,30 +10,94 @@ import (
...
+					if outputAsProportion {
+						currMotifScoreDifference = (currRefScore / currConsensusScore) - (currAltScore / currConsensusScore)
+						currAnnotation = []string{fmt.Sprintf("%f", currRefScore/currConsensusScore), fmt.Sprintf("%f", currAltScore/currConsensusScore), fmt.Sprintf("%f", currMotifScoreDifference)}
+					} else {
+						currMotifScoreDifference = currRefScore - currAltScore
```
craiglowe, line 49: What about when the motifs are aligned, but do not start at the same position.  For example, I am imagining a motif where the first two bases are not very important, but then the next four bases need to be ACGT.  I am also picturing the alignment:
ACCACGT
A-CACGT
I even though both have a strong match to this motif, it starts at alignment position 1 for the reference, but 0 for the query.
Am I understanding the code correctly?  Is there a smart way to deal with this?

craiglowe (commented):
Looks great!  It will be exciting to have this functionality!  I had a couple small comments.

PR #752, opened 2023-01-03
BedFormat Asymmetric Padding

rimangan:
New options to bedFormat enable the user to add an uneven left and right side pad to bed entries. Can be used additively with the pre-existing evenPadLength option.


cmd/bedFormat/bedFormat.go
```diff
@@ -81,7 +95,9 @@ func usage() {
 
 func main() {
 	var expectedNumArgs int = 2
-	var padLength *int = flag.Int("padLength", 0, "Add # of bases to both ends of each bed record. Requires noGapFile.")
+	var evenPadLength *int = flag.Int("evenPadLength", 0, "Add # of bases to both ends of each bed record. Requires chromSizeFile.")
+	var leftPadLength *int = flag.Int("leftPadLength", 0, "Add # of bases to the left end of the bed record. Requires chromSizeFile.")
```
craiglowe, line 99: I am not sure I understand the use case.  Should this be "upstream" and "downstream" instead of left and right so that it can look at the strand and extend up or downstream of a gene, or replication direction, or something else?  I am not sure when we would treat left and right different without knowing a strand (since in theory the assembly could have reported each chromosome as the reference complement, where left and right would be different).

craiglowe (commented):
I didn't see any bugs and I am happy to approve, but I don't totally understand the use case.  I left a comment in the code.

cmd/bedFormat/bedFormat.go
rimangan: Changed to upstream/downstream. Also added an option to trim a bed to its start position (strand-sensitive), called 'toTss'.

craiglowe (approved):
Looks good!

cfauci (approved):

PR #756, opened 2023-01-24
bedToWig useRange

rimangan:
This PR adds an option to bedToWig called useRange. By default, bedToWig assigns the bed value to the midpoint of the bed interval range to the output wig. With this option, the bed value will be applied across the bed range in the output wig.


convert/convert.go
```diff
@@ -128,25 +128,38 @@ func BedGraphToWig(inFile string, reference map[string]chromInfo.ChromInfo, miss
...
 			log.Fatalf("Two bed entries share the same midpoint. Unable to resolve ambiguous value assignment.")
 		}
-		if method == "Name" {
-			wigSlice[chromIndex].Values[midpoint] = common.StringToFloat64(b.Name)
-		} else if method == "Score" {
-			wigSlice[chromIndex].Values[midpoint] = float64(b.Score)
+
+		if useRange {
+			for i = b.ChromStart; i < b.ChromEnd; i++ {
+				if method == "Name" {
```
craiglowe, line 147: Before this "if statement" do we want to have a check (similar to newline 141) that makes sure the current value at this position is "missing"?

craiglowe (commented):
Looks great!  I just had one comment about checking for a missing value to make sure we don't have overlapping beds.

RavenYL (approved):

edotau (approved):

PR #757, opened 2023-01-28
updates to liftCoordinates

rimangan:
This PR fixes some bugs with reverse strand liftCoordinates and adds the option 'strictBorders', which sends records where either the chromStart or chromEnd fall in a TGap to the unmapped file.

### DIFF
--- a/chain/lift.go
+++ b/chain/lift.go
@@ -4,42 +4,53 @@ import (
 	"log"
 )
 
-//TPosToQPos converts a target position in a chain to the corresponding query position.
-func TPosToQPos(c Chain, TPos int) int {
+// TPosToQPos converts a target position in a chain to the corresponding query position.
+// The bool returns true if the returned TPos was found in a (Size) section of a Chain, and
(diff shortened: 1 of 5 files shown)


chain/lift.go
```diff
@@ -4,42 +4,55 @@ import (
...
+			}
+			currT += c.Alignment[i].TBases
+			currQ += c.Alignment[i].QBases
+		} else {
+			if currT+c.Alignment[i].Size > TPos {
+				return currQ - (TPos - currT), true
+			}
+			currT += c.Alignment[i].Size
+			currQ -= c.Alignment[i].Size
+			if currT+c.Alignment[i].Size > TPos {
```
craiglowe, line 48: Should this be:
currT+c.Alignment[i].TBases
instead of:
currT+c.Alignment[i].Size
?

craiglowe (commented):
Thank you!  Other than what I think might be one typo, this looks great!

craiglowe (approved):
Wonderful!

cfauci (approved):

PR #760, opened 2023-02-02
samConsensus Indel support

rimangan:
Update to samConsensus to support calling short insertions and deletions in the output fasta and vcf files. Here I have a new function in the sam package that identifies the "consensus" of a pileup, or the variant (be it an insertion, deletion, or base) that was seen the most times in a pileup. Can also return a multiFa alignment of each outputed chromosome with the reference. Still pending testing on real world data.
After some deliberation, I decided to remove the VCF functionality. As samConsensus creates a multiFa alignment, a VCF can always be made with multiFaToVcf using the samConsensus output.


sam/pileup.go
```diff
@@ -437,3 +438,140 @@ func (p *Pile) String() string {
...
+func PileConsensus(p Pile, substitutionsOnly bool) Consensus {
+	// first we check if consensus is a base
+	var max int = p.CountF[dna.A] + p.CountR[dna.A]
+	tiedConsensus := make([]Consensus, 1)
+	tiedConsensus[0] = Consensus{Base: dna.A, Type: Base}
+
+	max, tiedConsensus = getMaxBase(p, max, dna.C, tiedConsensus)
+	max, tiedConsensus = getMaxBase(p, max, dna.G, tiedConsensus)
+	max, tiedConsensus = getMaxBase(p, max, dna.T, tiedConsensus)
+	//question for Dan: main version checks dna.Gap here, is that necessary if we check the deletion field?
```
dasnellings, line 472: Not necessary for the new functionality, but it would be necessary to check when running with substitutionsOnly if you want to preserve legacy compatibility.

sam/pileup_test.go
```diff
@@ -253,3 +253,115 @@ func copyPile(p Pile) Pile {
...
+			Base:      1,
+			Insertion: []dna.Base{},
+			Deletion:  0,
+			Type:      Base,
+		},
+		substitutionsOnly: true,
+	},
+}
+
+func TestPileConsensus(t *testing.T) {
```
dasnellings, line 344: I _**strongly**_ advise testing this code with piles derived with GoPileup from sam records that contain both single base and multi-base deletions and insertions and check that you are getting the expected result. It may even be best to test this by generating a fasta consensus as you are planning. There is a high potential for gotchas here that will need to be carefully tested. For instance, the testing piles above (e.g. newline 326) list 42 bases having an insertion of AACTT on the F strand, but only 2 bases reported in the counts on the forward strand. The only way such a pile would be possible is if you had 40 reads that immediately begin with an insertion, which almost no aligner would allow. Piles are strictly generated per reference base, and since there is no index in the reference fasta for inserted sequence, the insertion is stored in the base immediately prior to the insertion. Therefore when you get an insertion as your consensus, you must also include a consensus for the bases/del in the same pile (i.e. you need 2 consensus structs for every pile where you are calling an insertion). I am disappointed in myself for not noting this in the documentation, but I am almost sure this is the case. If you confirm this to be true during your testing, I would appreciate you noting this in the documentation for the pile struct.

dasnellings (commented):
See my note above. Testing will be critical here, and it may be best tested by trying to generate a fasta consensus.

rimangan:
Note: Let's review VCF positions in code review monday

sam/pileup.go
```diff
@@ -437,3 +438,200 @@ func (p *Pile) String() string {
 	return fmt.Sprintf("RefIdx: %d\tPos: %d\tCountF: %v\tCountR: %v\tInsCountF: %v\tInsCountR: %v\tDelCountF: %v\tDelCountR: %v\tNext: %p\tPrev: %p",
 		p.RefIdx, p.Pos, p.CountF, p.CountR, p.InsCountF, p.InsCountR, p.DelCountF, p.DelCountR, p.next, p.prev)
 }
+
```
dasnellings, line 441: For the sake of my own future sanity as I am debugging, could I ask you to move the consensus stuff to a new file (maybe pileupConsensus.go?). pileup.go is already a pretty crowded file.

dasnellings (approved):

RavenYL (approved):

PR #762, opened 2023-02-08
handle single bp queries in interval

dasnellings:
This fixes a corner case when using single bp intervals as query for interval overlaps. A single bp interval (i.e. a point) violates the assumptions in the interval logic allowing it to satisfy multiple relationships. To fix this, we check if the input interval is a single bp and if so we do some extra work on return to sort and deduplicate the output.


rimangan (approved):
Awesome quick fix for this!!

rimangan (approved):

rimangan:
I'll make a note for everyone else: this change to interval.Query helped us catch another small bug in quickOrthologs. quickOrthologs takes a target TSS and tries to map to a query genome through a chain. If the TSS matches multiple chains, it goes to the unmapped file because the ortholog is ambiguous. The version on main aberrantly sent a TSS mapping to the same ortholog twice to unmapped, but our fix to sortAndDeduplicate allowed the program to find the right ortholog for that region. I pushed a new commit to this branch to update testing in quickOrthologs.

cfauci (approved):

PR #763, opened 2023-02-12
simulateEvol BranchLength with Indels

rimangan:
This PR introduces a new option to simulateEvol. Instead of simuating all sequences across a newick tree, the user may instead specify a 'branchLength' to make one child sequence from an input sequence at a specified degree of divergence. Notably, the output sequence will contain INDELs as well as substitutions. INDEL lengths are exponentially distributed, and the user may specify 'lambda', the rate parameter of the exponential distribution, to control the mean INDEL length. SimulateEvol also can now return a VCF in this mode, which records all mutations made in the child sequence.


simulate/simulateIntergenic.go
```diff
@@ -12,3 +20,187 @@ func RandIntergenicSeq(GcContent float64, lenSeq int) []dna.Base {
...
+	}
+
+	for inputPos < len(records[0].Seq) {
+		if inputPos%100_000 == 0 {
+			fmt.Printf("CurrentPos: %v.\n", inputPos)
+		}
+		currRand = rand.Float64() //this rand determines if there will be a mutation
+		if currRand < branchLength {
+			currRand2 = rand.Float64()         //this rand determines the mutation type
+			if currRand2 < (propIndel / 2.0) { //this half of INDELs will be deletions
```
RavenYL, line 69: I'm trying to understand why currRand2 < or >= (propIndel / 2.0) determines if the indel will be deletion or insertion. Is it just to divide up the indel cases equally, assuming it's equally likely to get any number?

PR #764, opened 2023-02-13
multiFaToVcf double counting bug

rimangan:
I found and fixed a small bug in multiFaToVcf.
When a substitution immediately precedes an INDEL, we should see a record like this: 
chr1	1431026	.	C	TT	100	PASS	.	.
In an alignment, this might look like this:
AAC-GTAC
AATTGTAC
On main, multiFaToVcf reports the following for this position:
chr1	1431026	.	C	T	100	PASS	.	.
chr1	1431026	.	C	TT	100	PASS	.	.

What appear to happen is the program sees the substitution, reports it to the VCF, but then the next iteration sees the indel, and reports that variant as well. 
Note that if the user uses the 'substitutionsOnly' option, the C -> T VCF line will still be written.

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -213,12 +213,12 @@ func getWigChromIndex(s string, wigSlice []wig.Wig) int {
 //entry as the reference and the second fasta entry as the alt allele.
 //This will have to be done by chromosome, as a pairwise multiFa will only have two entries, thus containing one chromosome per file.
 func PairwiseFaToVcf(f []fasta.Fasta, chr string, out *fileio.EasyWriter, substitutionsOnly bool, retainN bool) {
-	var pastStart bool = false //bool check to see if we have an insertion at the start of an alignment.
-	var insertion bool = false
-	var deletion bool = false
-	var insertionAlnPos int
-	var deletionAlnPos int
+	var pastStart, insertion, deletion bool = false, false, false //first bool checks to see if we have an insertion at the start of an alignment.
+	var insertionAlnPos, deletionAlnPos int
 	var currRefPos, currAlnPos int = 0, 0 //0 based, like fasta. Add 1 to get vcf pos.
+	if len(f) != 2 {
+		log.Fatalf("PairwiseFaToVcf expects a fasta input with two entries.")
+	}
 
 	for i := range f[0].Seq { //loop through alignment positions
 		if f[0].Seq[i] == dna.Gap { //reference is gap (insertion)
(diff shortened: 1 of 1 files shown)


convert/convert.go
```diff
@@ -270,7 +270,16 @@ func PairwiseFaToVcf(f []fasta.Fasta, chr string, out *fileio.EasyWriter, substi
...
+						if f[0].Seq[i+1] != dna.Gap && f[1].Seq[i+1] != dna.Gap { //if neither alt nor ref is a gap in the next pos.
+							vcf.WriteVcf(out, vcf.Vcf{Chr: chr, Pos: currRefPos + 1, Id: ".", Ref: dna.BaseToString(f[0].Seq[i]), Alt: []string{dna.BaseToString(f[1].Seq[i])}, Qual: 100.0, Filter: "PASS", Info: ".", Format: []string{"."}})
+						} else if substitutionsOnly { //we would also write the subsitution in the case where we don't care if the substitution precedes an INDEL because we are only reporting substitutions
+							vcf.WriteVcf(out, vcf.Vcf{Chr: chr, Pos: currRefPos + 1, Id: ".", Ref: dna.BaseToString(f[0].Seq[i]), Alt: []string{dna.BaseToString(f[1].Seq[i])}, Qual: 100.0, Filter: "PASS", Info: ".", Format: []string{"."}})
+						}
+						// Otherwise we won't report this substitution because it wil be part of the INDEL reporting.
+					} else { //for a substitution in the final position, we need not check for subsequent INDELs
+						vcf.WriteVcf(out, vcf.Vcf{Chr: chr, Pos: currRefPos + 1, Id: ".", Ref: dna.BaseToString(f[0].Seq[i]), Alt: []string{dna.BaseToString(f[1].Seq[i])}, Qual: 100.0, Filter: "PASS", Info: ".", Format: []string{"."}})
+					}
```
RavenYL, line 282: Since the final action (vcf.WriteVcf) seems to be the same, maybe the code can be condensed into 1 big if/else condition or fewer conditions than the current code?
It seems like we want to suppress reporting a substitution if: (There is a next base to look at AND (neither alt nor ref is a gap in the next position)) OR substitutionsOnly
In other words, we only want to report a substitution if: NOT((There is a next base to look at AND (neither alt nor ref is a gap in the next position)) OR substitutionsOnly)

RavenYL (approved):

craiglowe (approved):
Great!

PR #767, opened 2023-02-23
Recycle sam structs when seeking in bam file

dasnellings:
This PR adds a new function that allows you to input a slice of sam structs when seeking reads in a bam file. This will overwrite the input structs with the new reads dramatically reducing memory allocations on repeated seek calls.

### DIFF
--- a/sam/bamSeek_test.go
+++ b/sam/bamSeek_test.go
@@ -41,6 +41,48 @@ func TestSeekBamRegion(t *testing.T) {
 	}
 }
 
+func TestSeekBamRegionRecycle(t *testing.T) {
+	br, _ := OpenBam("testdata/rand.bam")
+	bai := ReadBai("testdata/rand.bam.bai")
+	reads := make([]Sam, 10)
+	reads = SeekBamRegionRecycle(br, bai, "chr7", 45000000, 45200000, reads)
+	for i := range reads {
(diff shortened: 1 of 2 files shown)


sam/bamSeek_test.go
```diff
@@ -41,6 +41,48 @@ func TestSeekBamRegion(t *testing.T) {
...
 }
 
+func TestSeekBamRegionRecycle(t *testing.T) {
+	br, _ := OpenBam("testdata/rand.bam")
+	bai := ReadBai("testdata/rand.bam.bai")
+	reads := make([]Sam, 10)
+	reads = SeekBamRegionRecycle(br, bai, "chr7", 45000000, 45200000, reads)
+	for i := range reads {
+		if reads[i].RName != "chr7" || !(reads[i].GetChromStart() < 45200000 && reads[i].GetChromEnd() > 45000000) {
+			t.Error("problem with SeekBamRegion")
```
RavenYL, line 51: Would the error message in this function be "problem with SeekBamRegionRecycle"? As far as I understand, since you test "SeekBamRegion" above, if only this function fails, you'll know it's the non-nil recycledSams []Sam that caused it to fail?

dasnellings: Good catch. Fixed.

rimangan (approved):

RavenYL (approved):

PR #770, opened 2023-03-02
simulateEvol with transition bias

rimangan:
This PR allows the user to simulate molecular evolution from an input sequence with a transition/transversion bias. On main, simulateEvol performs molecular evolution following "Jukes-Cantor with GC Preservation". In other words, transition probabilities are nearly equal, but expected GC content is preserved. This is done by biasing against C to G/ G to C substitutions and biasing towards T to A / A to T substitutions in cases where the expected GC is around 0.42, as is found in humans. 
Under a pure Jukes-Cantor model, in which all substitution probabilities are equal, sufficient branch lengths will drive the expected GC content to 0.5. 
I tested these molecular evolution models on hg38 chr22 with a branch length of 0.01, which gave me >300k variants to test the mutation spectrum.
I ran simulateEvol with a transitionBias of 1.001 to approximate pure Jukes-Cantor, and indeed saw much more even substitution frequencies.
Transition bias of 3 and 5, as expected, show increasing levels of transition bias (C to T/G to A and T to A/A to T). 
Note that in humans, the transition bias is about 3.

image


simulate/simulateIntergenic.go
```diff
@@ -35,7 +35,8 @@ const bufferSize = 10_000_000
 // lambda specifies the rate parameter for an exponential distribution, from which simulated INDEL sizes will be sampled.
 // gcContent specifies the expected value of GC content for inserted sequences.
 // vcfOutFile specifies an optional return, which records all variants made during the simulated mutation process.
-func SimulateWithIndels(fastaFile string, branchLength float64, propIndel float64, lambda float64, gcContent float64, vcfOutFile string) []fasta.Fasta {
+// transitionBias specifies the expected value of the ratio of transitions to transversions in the output sequence.
+func SimulateWithIndels(fastaFile string, branchLength float64, propIndel float64, lambda float64, gcContent float64, vcfOutFile string, transitionBias float64) []fasta.Fasta {
```
craiglowe, line 39: Convention is to have input-associated variables before output-associated variables, so it might be more intuitive to have transitionBias come before vcfOutFile

simulate/simulateIntergenic.go
```diff
@@ -204,3 +214,46 @@ func SimulateWithIndels(fastaFile string, branchLength float64, propIndel float6
 
 	return answer
 }
+
+// changebaseTransitionBias substitutes an input base b following the K80 model with a transitionBias parameter gamma.
```
craiglowe, line 218: typo, needs capital B

simulate/simulateIntergenic.go
```diff
@@ -204,3 +214,46 @@ func SimulateWithIndels(fastaFile string, branchLength float64, propIndel float6
...
+func changeBaseTransitionBias(b dna.Base, gamma float64) dna.Base {
+	var rand = rand.Float64()
+	switch dna.ToUpper(b) {
+	case dna.A:
+		if rand < 1.0/(2.0+gamma) { //p(a -> c) = 1 / (2 + gamma) (transversion)
+			return dna.C
+		} else if rand < 2.0/(2.0+gamma) { //p(a -> t) = 1 / (2+gamma) (transversion)
```
craiglowe, line 225: The code keeps calculating:
1.0/(2.0+gamma)
and
2.0/(2.0+gamma)
very minor, but might be best to do each of those calculations once and store each in a variable.  Then all your case/if statements just reference on of those two variables.  That can also help avoid typos in one place.

craiglowe (commented):
Two super minor comments.  I think this looks pretty much ready to merge in

simulate/simulateIntergenic.go
```diff
@@ -204,3 +214,47 @@ func SimulateWithIndels(fastaFile string, branchLength float64, propIndel float6
 
 	return answer
 }
+
+// changeBaseTransitionBias substitutes an input base b following the K80 model with a transitionBias parameter gamma.
+func changeBaseTransitionBias(b dna.Base, gamma float64) dna.Base {
+	var rand = rand.Float64()
+	var TvProb float64 = 1.0 / (2.0 + gamma) //p(a -> c) = 1 / (2 + gamma) (transversion)
+	switch dna.ToUpper(b) {
+	case dna.A:
```
RavenYL, line 223: Why does dna.A never return dna.A, and can only return C or T or G? Does this mean there will definitely be a substitution (as opposed to no substitution, A still returns A)?

cfauci (approved):

RavenYL (approved):

PR #774, opened 2023-03-07
Mask invalid bases in fasta

dasnellings:
The focus of this PR is on handling cases where there are non-ACGTN- characters in a fasta file. It adds 
1. `fasta.ReadForced` function to N-mask any invalid characters when reading in a fasta
2. an option in `faFormat` to N-mask invalid characters in the input file

The intention is for users to run `faFormat` on any invalid fasta before using it with gonomics (I also updated error messages to encourage this). IMO use of `fasta.ReadForced`, `fasta.NextFastaForced`, and `dna.StringToBasesForced` should be strongly discouraged in gonomics outside of their use in `faFormat`


fasta/fasta.go
```diff
@@ -27,14 +27,24 @@ type FastaMap map[string][]dna.Base
...
+	return read(filename, NextFasta)
+}
+
+// ReadForced functions identically to Read, but any invalid characters in the sequence will be masked to N.
+func ReadForced(filename string) []Fasta {
+	return read(filename, NextFastaForced)
+}
+
+// read is a helper function for Read and ReadForced to reduce duplication
+func read(filename string, nextFunc func(*fileio.EasyReader) (Fasta, bool)) []Fasta {
```
craiglowe, line 39: aahhhh, that is an interesting way to do it.  I was picturing passing a bool down through the functions, but I think this is probably a more elegant way to do it.

craiglowe (approved):
Great!  Thank you!

fasta/fasta.go
dasnellings: I figured since it we will not want to force the conversion 99.9% of the time, it would be less disruptive to just make a new function rather than adding a bool input to fasta.Read.

craiglowe: Yes, I understand.  I was thinking of something like:
fasta.Read(filename) calling: readHelper(filename, true)
and
fasta.ReadForced(filename) calling: readHelper(filename, false)
but I think your solution is more elegant since you don't need to interpret the bool inside the function

rimangan (approved):
Looks great! What do you think about catching bases in the sequence that are not a part of the IUPAC convention (say for example 'E') and throwing a different error? Especially since mskInvalid mentions the IUPAC included bases?

PR #775, opened 2023-03-10
mergeMultiFa

rimangan:
mergeMultiFa is able to take two multiFa alignment files on the same reference chromosome and create a merged output.


RavenYL (approved):

rimangan:
Line 52: check if ref in file 1 and file 2 is equal, if not, through fatal

cfauci (approved):

PR #776, opened 2023-03-10
Vcf context

dasnellings:
This PR adds a new cmd that reports the base context of variants in a vcf. The primary use for me is getting the trinucleotide context of mutations to see enrichments (e.g. at CpG sites), but I thought this might also be useful for others in the future. In theory you could use this to look at variants that arise in TF binding motifs which might be cool to see if mutations in regions under positive selection are enriched in certain motifs.


rimangan:
Very cool program! What do you think about incorporating this into the existing cmd vcfInfo?

dasnellings:
That is a good idea. It may be a minute before I have a chance to merge them, so please wait to approve until that is done.

cmd/vcfContext/vcfContext.go
```diff
@@ -0,0 +1,214 @@
...
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var ref *string = flag.String("r", "", "Reference fasta file (must be indexed). Required for -pad > 0.")
+	var input *string = flag.String("i", "", "Input vcf file.")
+	var output *string = flag.String("o", "stdout", "Output tsv file.")
+	var includeComplements *bool = flag.Bool("includeComplements", false, "Do not merge totals from complementary mutations (e.g. separate C > A and G > T counts). Should only be used when looking at single-stranded mutations.")
+	var pad *int = flag.Int("pad", 1, "Number of context bases on either side of variant (e.g. 0 == T, 1 == ATG, 2 == TATGA, ...")
```
RavenYL, line 33: Is there some kind of maximum threshold for pad, e.g. length of reference fasta sequence?

dasnellings: Yes, though the memory cost of pad scales as a factorial so you will run out of memory long before you reach the length of the fasta. In reality the soft limit of pad is probably ~10, but I have not tested it.

edotau (approved):
lgtm!

edotau (approved):

cfauci (approved):

PR #777, opened 2023-03-12
SamAssembler Part 1: Diploid Variant Calling from Pileup

rimangan:
First PR for my planned future program, to be called samAssembler. 
Essentially, I want to adapt samConsensus to make diploid variant calls and return two output sequences from input SAM/BAM sequencing data. Short reads won't give us phase information, so we won't be constructing two haplotypes, but the hope is we will be able to recover twice as much sequence information as from the haploid pseudoassemblies from samConsensus. 

In this first PR, I introduce a new file, assembly.go, in the SAM package, that makes diploid variant calls from input Piles. 
If we assume Piles are produced from diploid sequences, we can consider 10 genotypes: AA, AC, AG, AT, CC, CG, CT, GG, GT, and TT (plus NN as an 11th, secret genotype).

For a given pile, this program identifies the genotype with the highest posterior probability using a Bayesian framework. 
Briefly, using Bayes' Rule:

$p(Genotype | Pile) = \frac{P(Pile | Genotype)P(Genotype)}{P(Pile)}$

Different genotypes can be compared using the ratio of their posteriors:

$\frac{p(GenoA | Pile)}{p(GenoB | Pile)} =  \frac{P(Pile | GenoA)P(GenoA)}{P(Pile | GenoB)P(GenoB)}$

I will be uploading a README shortly that provides the complete mathematical explanation of the variant caller.
Note that in low depth, the variant caller is conservative, and will default to the reference base at that position. If we have some data, maybe 4 reads for C, no other reads, and the reference is A, the diploid call will be AC (one of the test cases does this, and this is an interesting but potential desirable property).

Newest draft of the mathematical explanation:
[samAssembly.pdf](https://github.com/vertgenlab/gonomics/files/10975122/samAssembly.pdf)


rimangan:
[samAssembly.pdf](https://github.com/vertgenlab/gonomics/files/10951628/samAssembly.pdf)
Current README draft

sam/assembly.go
```diff
@@ -0,0 +1,219 @@
...
+		maxDiploid = []DiploidBase{CC}
+		maxPosterior = logspace.Multiply(baseLikelihood(aCount, cCount, gCount, tCount, CC, epsilon, homozygousCache, heterozygousCache), priorCache[dna.C][CC])
+	case dna.G:
+		maxDiploid = []DiploidBase{GG}
+		maxPosterior = logspace.Multiply(baseLikelihood(aCount, cCount, gCount, tCount, GG, epsilon, homozygousCache, heterozygousCache), priorCache[dna.G][GG])
+	case dna.T:
+		maxDiploid = []DiploidBase{TT}
+		maxPosterior = logspace.Multiply(baseLikelihood(aCount, cCount, gCount, tCount, TT, epsilon, homozygousCache, heterozygousCache), priorCache[dna.T][TT])
+	default:
+		log.Fatalf("Reference base was not N, A, C, G, or T. Found: %s.\n", dna.BaseToString(refBase))
```
dasnellings, line 106: You may consider a panic here (and elsewhere in this file) instead of a fatal. This is a pretty handy function I can see being used in other places in gonomics. If this is used outside of samAssembly tracking down this error could be pretty hard. A stack trace from a panic would help a lot.

sam/assembly.go
```diff
@@ -0,0 +1,219 @@
...
+	case TT:
+		return "TT"
+	case NN:
+		return "NN"
+	}
+	log.Fatalf("Unrecognized DiploidBase: %v.", base)
+	return ""
+}
+
+//DiploidBaseCallFromPile determines which of 10 DiploidBase genotypes
```
dasnellings, line 58: it would be helpful for the documentation to define where to get or how to derive the caches.

sam/assembly_test.go
```diff
@@ -0,0 +1,163 @@
...
+		Delta:    0.1,
+		Gamma:    3,
+		Epsilon:  0.01,
+		Expected: AT},
+}
+
+func TestDiploidBaseCallFromPile(t *testing.T) {
+	var actual DiploidBase
+	var priorCache [][]float64
+	var heterozycousCache = make([][]float64, 0)
```
dasnellings, line 70: Where are the caches actually being defined? I am looking through the functions and I can't find where they are being built???

dasnellings (commented):
Made it part way through. I'll pick up later.

sam/assembly_test.go
rimangan: Right, so for the tests, we aren't using caches and calculating everything by hand. The plan is that the cmd (to be called samAssembler) will make caches and pass them to the package functions. You can see the functions to initialize the caches at the end of the file

dasnellings: I see now, so you check to see if the cache is available else calculate by hand. You should consider making the cache building functions public in assembly.go assuming the caches will be broadly applicable. Having DiploidBaseCallFromPile be a public library function, but making the functions to build the caches for it private makes it pretty hard for other people to use DiploidBaseCallFromPile.

dasnellings (approved):
I think this can go in as-is so I am approving, but there are some minor points I commented on that you should consider. Happy to re-approve if you make changes.

sam/assembly.go
rimangan: This is a great idea. Will implement in next commit

sam/assembly_test.go
rimangan: Ordinarily I would, but for the likelihood, I think cache building will be trivial: we'll make a 100 by 100 [][]float64, all initialized to zero. We expect sparse usage of the whole cache space (most piles should have the same approximate coverage, and generally be pretty neat either het or homozygous), so we won't precalculate everything, and fill the cache as we progress through the program

sam/assembly.go
```diff
@@ -0,0 +1,219 @@
...
+	if refBase == dna.N { //we will force a return of diploid NN whenever an N is found in the reference.
+		return NN
+	}
+	var aCount = p.CountF[dna.A] + p.CountR[dna.A]
+	var cCount = p.CountF[dna.C] + p.CountR[dna.C]
+	var gCount = p.CountF[dna.G] + p.CountR[dna.G]
+	var tCount = p.CountF[dna.T] + p.CountR[dna.T]
+	var baseCount = aCount + cCount + gCount + tCount
+
+	if baseCount < 1 {
```
RavenYL, line 72: Note to self: conservative because defaults to refBase

sam/assembly.go
```diff
@@ -0,0 +1,219 @@
...
+		if currPosterior > maxPosterior {
+			maxPosterior = currPosterior
+			maxDiploid = maxDiploid[:1] //clear ties
+			maxDiploid[0] = geno
+		} else if currPosterior == maxPosterior {
+			maxDiploid = append(maxDiploid, geno)
+		}
+	}
+
+	return maxDiploid[numbers.RandIntInRange(0, len(maxDiploid))] //if two genotypes have the same posterior density, pick one at random.
```
RavenYL, line 125: (For future cmd) Should there be a note in usage about "if two genotypes have the same posterior density, pick one at random"?

sam/assembly.go
```diff
@@ -0,0 +1,219 @@
...
+		f := logspace.Pow(math.Log(epsilon/3.0), float64(incorrectCount))
+		return logspace.Multiply(s, f)
+	}
+}
+
+// makePriorCache is a helper function used in samAssembler before running DiploidBaseCallFromPile.
+// Constructs a matrix ([][]float64) of form answer[refBase][outGenotype], where index of refBase: 0=a, 1=c, 2=g, 3=t.
+// Index of outGenotype follows the indices of the DiploidBase type.
+// Answers are log transformed.
+func makePriorCache(delta float64, gamma float64) [][]float64 {
```
RavenYL, line 204: Note to self: cache is made here

RavenYL (approved):
I remember Riley introduced the formulas at a previous code review, and the document is helpful. Trusting that the formulas were derived correctly, the implementation in the code looks good to me, I look forward to learning how the cache will be generated in the cmd, and any expansions to Indels

PR #778, opened 2023-03-15
samAssembler Part 2: Diploid INDEL calling from Pileup

rimangan:
Dependent on the last PR. This part adds functionality to call diploid insertion and deletion genotypes from Piles. Follows a similar Bayesian framework to the diploid substitution caller in the last PR. Mathematics are described in the document below.
[samAssembly.pdf](https://github.com/vertgenlab/gonomics/files/10975139/samAssembly.pdf)


sam/diploidIndel_test.go
```diff
@@ -0,0 +1,163 @@
...
+	ExpectedInsertion DiploidInsertion
+	ExpectedDeletion  DiploidDeletion
+}{
+	{P: Pile{CountF: [13]int{30, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}, //high coverage, clear complex het insertion
+		InsCountF: map[string]int{"AAT": 7, "AT": 6},
+		InsCountR: map[string]int{"AAT": 6, "AT": 5}},
+		Delta:             0.01,
+		Epsilon:           0.01,
+		Kappa:             0.05,
+		ExpectedInsertion: DiploidInsertion{Type: IaIb, Ia: "AAT", Ib: "AT"},
```
RavenYL, line 19: Note to self: running insertion prediction and deletion prediction on the same pile should get the same/consistent results

RavenYL (approved):
Approving because after trying to understand the pdf and reviewing the code to the best of my ability, this looks good, but happy to look into specific issues if needed!

PR #782, opened 2023-03-21
Add .bam functionality to samToWig

sweavs111:
uses sam.GoReadToChan to read either a sam or bam file as the input for samToWig


sweavs111:
for the testing files, I used 'samtools view -h -b' to turn the test.sam files into bams (needed to add a more realistic sam header). And made sure the bam and the sam gave the same output

RavenYL (approved):

cfauci (approved):

PR #783, opened 2023-03-23
Run gonomics cmd without source code

dasnellings:
This PR is dependent on my previous gonomics PR and is a bit meta.

The gonomics command parses the gonomics source code to assign commands to groups per the command group annotation in the cmd.go file. This was all working before this PR, but it meant that in order to use the gonomics cmd you must have all the source code files which may not be the case for users that only retrieve the binaries (after an eventual release). This PR parses the source code the same as it used to, but creates a cache of the information in the cmd/gonomics directory. This cache is then embedded in the gonomics executable using the `//go:embed command_cache.txt` keyword. The contents of the embedded file is then available to the program via a global variable. This enables the gonomics command to be used via the cached data rather than needing to parse the source code on each execution.

This does mean that the command_cache.txt file needs to be kept updated which would be rather tedious so I also added a part to the github workflow to update this file alongside the code formatting that it is already doing.

Apologies in advance as I will be making several commits to this PR to test the workflow is working.


dasnellings:
OK, looks like everything is finally working. Sorry for all the commits and any associated email spam. Unfortunately the only way to test the workflow is to push to an open PR. This is now ready for review.

.github/workflows/go.yml
```diff
@@ -35,17 +35,24 @@ jobs:
...
         cp -r ../gonomics /home/runner/go/src/github.com/vertgenlab/ &&
         cd /home/runner/go/src/github.com/vertgenlab/gonomics &&
+        go install ./... &&
         go test ./...
         
   format:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
```
edotau, line 44: I would just upgrade to v3, reference: [actions/checkout#checkout-v3](https://github.com/actions/checkout#checkout-v3)
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/go.yml
```diff
@@ -35,17 +35,24 @@ jobs:
...
         with:
           ref: ${{ github.head_ref }}
       - uses: actions/setup-go@v3
```
edotau, line 47: actions/checkout@v3 will force you to upgrade to [setup-go@v4](https://github.com/actions/setup-go#v4)
```suggestion
      - uses: actions/setup-go@v4
```

edotau (dismissed):
Left a couple suggestions, but otherwise has my node of approval.

.github/workflows/go.yml
```diff
@@ -76,6 +83,7 @@ jobs:
           mkdir --parents /home/runner/go/src/github.com/vertgenlab/ &&
           cp -r ../gonomics /home/runner/go/src/github.com/vertgenlab/ &&
           cd /home/runner/go/src/github.com/vertgenlab/gonomics &&
+          go install ./... &&
           go test -coverprofile=coverage.txt -covermode=atomic ./...
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v2
```
edotau, line 89: And if you upgrade to checkout@v3 and setup-go@v4, I'd check to make sure codecov-action@v2 is still compatible as well

edotau, line 89:
```suggestion
        uses: codecov/codecov-action@v3
```

edotau: codecov uses v3: https://github.com/codecov/codecov-action#migration-from-v1-to-v3

edotau: **actions/setup-go@v4** for lines:
- 16: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L16
- 30: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L30
- 47: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L47
- 78: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L78

actions/checkout@v3 for lines:
- 15: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L15
- 29: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L29
- 44: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L44
- 75: https://github.com/vertgenlab/gonomics/blob/fcd91e13a696afbf717f942bcae27ce27eb20f69/.github/workflows/go.yml#L75

dasnellings:
@edotau I tried updating the versions for checkout, setup-go and codecov-action, but it seemingly broke the workflow. I am not sure how to check the logs if the workflow fails to trigger so I'm not sure how to debug. Any advice on a better way to test the workflow?

edotau (approved):

edotau (approved):

cmd/gonomics/command_cache.txt
```diff
@@ -0,0 +1,255 @@
+##SourceDir:/home/runner/work/gonomics/gonomics/cmd/
```
RavenYL, line 1: My understanding is that if a user only retrieves binaries and not all source code files, they still won't be able to parse all the source code and run the gonomics cmd that way, but they will be able to use this cache to run the gonomics cmd. And the cache gets updated whenever another user that has all the source code files runs the gonomics cmd and makes a commit. Is this right?

RavenYL (approved):

craiglowe (commented):
This looks great!  My only comment/question is if the cache should get preference over the source code when both are present.  This means that as we update programs, the gonomics program may, or may not, update depending on if the cache is there.  I guess the cache will usually be there, so that means we will have to explicitly build the cache in order to have gonomics update.  Not sure which way people prefer.

cmd/gonomics/usage.go
```diff
@@ -4,12 +4,10 @@
 package main
 
 import (
-	"bufio"
-	"bytes"
 	"fmt"
 	"github.com/vertgenlab/gonomics/exception"
 	"github.com/vertgenlab/gonomics/fileio"
-	"io/ioutil"
+	"golang.org/x/exp/slices"
```
craiglowe, line 10: Do we need to update go.mod/num with this?

craiglowe (commented):
Quick comment about go.mod/num due to x/exp/slices

cmd/gonomics/usage.go
dasnellings: We already require x/exp so no update needed.

cmd/gonomics/gonomics.go
```diff
@@ -64,10 +69,41 @@ func getBin() (path string, binExists map[string]bool) {
...
-// tryPath tests a path for the presence of intervalOverlap (arbitrary chosen command)
+// getSrc returns the path to the source code files for commands
+func getSrc() string {
+	var expectedPath string
+	var cachedSrcPath string = getCachedSrcDir()
+	gopath := os.Getenv("GOPATH") + "/src/github.com/vertgenlab/gonomics/cmd/"
+	godefault := os.Getenv("HOME") + "/go/src/github.com/vertgenlab/gonomics/cmd/"
+
+	switch {
+	case cachedSrcPath != "" && tryPathDir(cachedSrcPath):
```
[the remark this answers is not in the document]
dasnellings: That is a good point. I think it is a fairly minor change so I am going to go ahead and merge this PR and open a new one in a few minutes with just this addition.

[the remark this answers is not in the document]
dasnellings: So this turns out to be non-trivial. This switch case is only a secondary decision on whether to use the cache or the source. The core branching point is at the start of getGonomicsCmds function where if the cache is full, we use the cache. To add the functionality you suggest we would need to define a trigger to decide whether we should use the cache or the source. Alternatively we would just always use the source when present, but running the cmd from source takes ~30s which was the initial impetus for making a cache, so not really a viable option. The only reasonable trigger I can think of is to check if any of the source code has been modified since the last time the gonomics cmd was built. That would require running os.Stat on each file in the source code before executing the command. That is a valid option, but it would probably add a couple seconds to the runtime and be quite a bit of work to code. I am in favor of just asking developers to rebuild the gonomics cache manually if they want to use uncommited commands which I don't think is too big an ask since I suspect most devs will rarely use the gonomics cmd anyways. If anyone feels strongly about this, I am happy to add this as a todo for me.

PR #786, opened 2023-03-27
dna: safer Extract

dolmen:
In dna.Extract, use the 3-arg slicing for safety: any append will force a copy instead of overwriting the original input.


dasnellings (approved):
I did not know 3-arg slicing was a thing. Thanks for the lesson!

rimangan (approved):

PR #789, opened 2023-03-27
/!\ Fix incorrect counting of 'n' in dna.Count, dna.CountMask

dolmen:
Fix two major bugs in `dna.Count` and `dna.CountMask`: the count of `n` was incorrect and the count of 'N' was reported instead.

This bug is fixed in b89aa8e2ac6012911bf07477ab32a3f7a2c660e1: just a typo `nCount` vs `NCount`.

This patches series also adds more testing (the first commit 192422863b9e88c771a7d99b86d771e17de798e3 adds failing tests) and simplify the implementation of the 2 functions `Count` and `CountMask` to reduce the risk of typos.


dasnellings (approved):
Nice catch!

PR #790, opened 2023-03-27
dna: faster BasesToString

dolmen:
Drop use of `strings.Builder` as we already know the final length of the string and each base is represented as a single ASCII byte.

### DIFF
--- a/dna/convert.go
+++ b/dna/convert.go
@@ -3,10 +3,10 @@ package dna
 import (
 	"errors"
 	"fmt"
-	"github.com/vertgenlab/gonomics/exception"
 	"log"
-	"strings"
 	"unicode"
+
+	"github.com/vertgenlab/gonomics/exception"
 )
 
 // RuneToBase converts a rune into a dna.Base if it matches one of the acceptable DNA characters.
@@ -179,16 +179,11 @@ var baseToByteArray = []byte{'A', 'C', 'G', 'T', 'N', 'a', 'c', 'g', 't', 'n', '
 
 // BasesToString converts a slice of DNA bases into a string. Useful for writing to files.
 func BasesToString(bases []Base) string {
-	var buffer strings.Builder
-	buffer.Grow(len(bases))
-	var err error
-	for i := range bases {
-		err = buffer.WriteByte(baseToByteArray[bases[i]])
-		if err != nil {
-			log.Panicf("problem writing byte")
-		}
+	buf := make([]byte, len(bases))
+	for i, b := range bases {
+		buf[i] = baseToByteArray[b]
 	}
-	return buffer.String()
+	return string(buf)
 }
 
 // ByteSliceToDnaBases will convert a slice of bytes into a slice of Bases.


dasnellings:
I'm not sure this is actually faster. I tried to benchmark the two cases with the following code
```
func BenchmarkBasesToStringBuilder(b *testing.B) {
	bases := StringToBases("AaCcGgTtNn")
	for i := 0; i < b.N; i++ {
		basesToStringBuild(bases)
	}
}

func BenchmarkBasesToStringCast(b *testing.B) {
	bases := StringToBases("AaCcGgTtNn")
	for i := 0; i < b.N; i++ {
		basesToStringCast(bases)
	}
}

func basesToStringBuild(bases []Base) string {
	var buffer strings.Builder
	buffer.Grow(len(bases))
	var err error
	for i := range bases {
		err = buffer.WriteByte(baseToByteArray[bases[i]])
		if err != nil {
			log.Panicf("problem writing byte")
		}
	}
	return buffer.String()
}

func basesToStringCast(bases []Base) string {
	buf := make([]byte, len(bases))
	for i, b := range bases {
		buf[i] = baseToByteArray[b]
	}
	return string(buf)
}
```

which does reports casting as the faster option
```
BenchmarkBasesToStringBuilder-8         42331587                26.72 ns/op           16 B/op          1 allocs/op
BenchmarkBasesToStringCast-8            52970290                22.27 ns/op           16 B/op          1 allocs/op
```
That was kind of confusing since I would have expected you need 2 allocations for casting (one to make the []byte and one to make the string). I tried modifying the benchmarks to actually do something with the string and I actually get the opposite result.

```
func BenchmarkBasesToStringBuilder(b *testing.B) {
	bases := StringToBases("AaCcGgTtNn")
	var s string
	for i := 0; i < b.N; i++ {
		s = basesToStringBuild(bases)
	}
	fmt.Sprint(s)
}

func BenchmarkBasesToStringCast(b *testing.B) {
	bases := StringToBases("AaCcGgTtNn")
	var s string
	for i := 0; i < b.N; i++ {
		s = basesToStringCast(bases)
	}
	fmt.Sprint(s)
}

BenchmarkBasesToStringBuilder-8         40734349                27.03 ns/op           16 B/op          1 allocs/op
BenchmarkBasesToStringCast-8            35017539                33.91 ns/op           32 B/op          2 allocs/op
```

All I added here was an assignment and a Sprint call after the loop and here we are using the expected 2 allocations per op and the function is slightly slower than using the builder. I would be curious to hear your thoughts on this, but I suspect that the compiler is doing some optimization to avoid an allocation for the string when it is not accessed later in the function, but requires a copy if the string is being accessed later to avoid situations where the string could be mutated by changing the byte slice.

dolmen:
OK. Could you commit the benchmarks?

dasnellings:
I don't have permissions to push to your fork

craiglowe:
These two implementations seem similar for the short sequences we are currently testing on.  Do we see a difference at longer sequences, 50 or 100?

dasnellings:
Good question Craig. I ran another benchmark converting 10Mb. Here are the results

```
func BenchmarkBasesToStringBuilderAssignLong(b *testing.B) {
	bases := make([]Base, 10000000)
	var s string
	for i := 0; i < b.N; i++ {
		s = basesToStringBuild(bases)
	}
	fmt.Sprint(s)
}

func BenchmarkBasesToStringCastAssignLong(b *testing.B) {
	bases := make([]Base, 10000000)
	var s string
	for i := 0; i < b.N; i++ {
		s = basesToStringCast(bases)
	}
	fmt.Sprint(s)
}

BenchmarkBasesToStringBuilderAssignLong-8            105          11186581 ns/op        10288232 B/op          1 allocs/op
BenchmarkBasesToStringCastAssignLong-8               140           8250868 ns/op        20219221 B/op          2 allocs/op
```

Casting is ~40% faster at the cost of 100% more memory. Not sure how best to balance these... Any thoughts here @edotau

dolmen:
@dasnellings What about pushing the benchmark to the main branch? It should be relevant for future maintenance even if my patch is not accepted.

dasnellings:
@dolmen agreed. They were added to main in https://github.com/vertgenlab/gonomics/pull/805

edotau (approved):
Yeah these are fine with me. I think another advantage is code simplicity.

rimangan (approved):

PR #791, opened 2023-03-28
First version of geneIdToName

sweavs111:
Both the NCBI RefSeq and Ensemble conversion tables are within the .go file so that issue is solved. Fully functional, I have used it.
To do / options:
- Write testing
- name to id
- option to only keep lines that had a successful conversion
- make a web app (?)
- open to other suggestions


sweavs111:
Here is the code without the explicitley defined NCBI / ENSEMBL maps:

`func makeMap(gnt string) map[string]string {
	mp := make(map[string]string)
	file := fileio.Read(gnt)

	for _, i := range file {
		columns := strings.Split(i, "\t")
		mp[columns[0]] = columns[1]
	}
	return mp
}

func geneIdToName(geneNameTable string, inFile string, outFile string, ncbi bool, ensembl bool) {
	var found bool
	var a, name string
	var err error
	found = false

	in := fileio.Read(inFile)
	out := fileio.EasyCreate(outFile)

	var mp map[string]string

	if ncbi {
		mp = makeNcbiMap()
	} else if ensembl {
		mp = makeEnsemblMap()
	} else {
		mp = makeMap(geneNameTable)
	}

	var z, x int
	for _, i := range in {
		found = false
		columns := strings.Split(i, "\t")
		for j := range columns {
			name, found = mp[columns[j]]
			if found {
				z++
				columns[j] = name
			}
		}
		if !found {
			x++
		}
		a = strings.Join(columns, "\t")
		fileio.WriteToFileHandle(out, a)
	}
	fmt.Printf("geneIDs lifted: %d\n", z)
	fmt.Printf("geneIDs not found: %d\n", x)
	err = out.Close()
	exception.PanicOnErr(err)
}

func usage() {
	fmt.Print("geneIdToName - finds and replaces gene IDs with gene names according to an input table \n" +
		"All files must be tab delimited. \n" +
		"The input table must be in the format geneID -tab- geneName \n" +
		"If a geneID in the input file is not in the inputTable if will remain unchanged in the output\n" +
		"Usage:\n" +
		"geneIdToName inputTable.txt in.txt out.txt\n" +
		"or\n" +
		"geneIdToName -ncbi/ensembl in.txt out.txt\n")
	flag.PrintDefaults()
}

func main() {
	var ncbi *bool = flag.Bool("ncbi", false, "Use the preset NCBI RefSeq ID to Gene Name conversion table (hg38).")
	var ensembl *bool = flag.Bool("ensembl", false, "Use the preset ENSEMBL ID to Gene Name conversion table (hg38).")
	var expectedNumArgs int

	flag.Usage = usage
	log.SetFlags(log.Ldate | log.Ltime)
	flag.Parse()

	if *ncbi || *ensembl {
		expectedNumArgs = 2
	} else {
		expectedNumArgs = 3
	}

	if len(flag.Args()) != expectedNumArgs {
		flag.Usage()
		log.Fatalf("Error: expecting %d arguments, but got %d\n\n", expectedNumArgs, len(flag.Args()))
	}
	var inputTable, in, out string
	if *ncbi || *ensembl {
		in = flag.Arg(0)
		out = flag.Arg(1)
	} else {
		inputTable = flag.Arg(0)
		in = flag.Arg(1)
		out = flag.Arg(2)
	}
	geneIdToName(inputTable, in, out, *ncbi, *ensembl)
}`

dasnellings:
Instead of adding the map to the .go file, you might consider using the embed functionality of go (see https://pkg.go.dev/embed).

That was you can keep the file separate so it is a lot easier to maintain (and can be gzipped).

RavenYL (commented):
Note to self: need to check out git branch to review code, since the diff is too large (probably because of the gene ids and names stored in the .go file) to be displayed on github

edotau (dismissed):

cfauci:
I think the function name "makeMap" should be more explicit considering our map usage and the existing Go function make. Maybe something like makeGeneNamMap.

I think to make this more useable we should at the very least move the makeMap function out to a package and then command would just be geneIdToName, but we can also make a wrapper for that to use as the command and make geneIdToName a package function that we can use throughout the codebase

having the option to not specify ncbi and ensembl and give an alternate gene name table is excellent good job with that!

in usage you have [options] notation for how to write out the program but you also write out the options as if they are required inputs which makes the instructions confusing. I would change this to:

`geneIdToNae [options] in.txt out.txt `

and then the flag that you have just below that will automatically write out what you've written in the main function for each option.

edotau (approved):
LGTM

- Note: I fixed some of your merge conflicts but then accidentally checked in some of local files I was playing around for proof of concept. I removed them but let me know if I mess anything else up

cfauci (approved):
Looks Great!

PR #801, opened 2023-04-01
samAssembler Part 4: The Complete Program

rimangan:
At last, samAssembler. This program takes in a fasta file of a reference genome and a bam file of reference-aligned short reads, assumed to be sampled from a diploid organism. The program them produces two assembled output sequences based on the inferred genomic sequence from which the reads were derived.
Complete mathematical description can be found in the attached document. Draft PR pending testing.
[samAssembly.pdf](https://github.com/vertgenlab/gonomics/files/11129951/samAssembly.pdf)


rimangan:
image
Here's the state of the output at 60x coverage and a small,  non repetitive genome. Top sequences is the reference sequence. Nex two are the diverged sequence from the generative model simulateEvol. The bottom two sequences are the reconstructed sequences from samAssembler.

rimangan:
> In the example screenshot, are the two simulate evol sequences 2 chromosomes of the same organism? And the Reconstructed genome is also diploid? Or are these technical replicates? Sorry if you answered this question elsewhere. If they are diploid, I am surprised that the GG dinucleotide in the 3rd row wasn't able to be phased correctly. Although overall the phasing looks decent to me. What was the simulated read length

All good questions. So here I used simulateEvol to make 2 different chromosomes and made reads from the 2 chromosomes and mixed them all together (like the experimental data we would start with from a diploid organism). I then use samAssembler to try to reconstruct the original diploid genome. So evol1 and evol2 represent the two simulated genomes, and Con_A and Con_B are the reconstructed contigs. Since this program works pile by pile, we won't have any correct phasing, except by chance. A potentially smarter program would look left and right along reads to try to capture local phasing, but this wasn't a priority for my program. 
Simulated reads are from a synthetic library with 60x coverage, 150x2 paired end reads with a mean fragment length of 400bp

edotau (changes requested):
Happy to talk about the review if you have questions or don't agree with anything I wrote. I usually do PR reviews in two passes:
1. **Programming cleanliness and technical aspects relating to software.** 
     - I look at language concepts, are you using the pointers correctly, importing the correct libraries, writing code in the correct format, and/or small typos, etc...
2. **Biological and experimental correctness**. 
    - Since I do not have a lot of insight into the specific  project, I didn't look at this part too carefully. I'm aware it's a small variant caller that performs that returns an assembly but I don't know all the corner cases you are looking for.

edotau (changes requested):
Most of these changes might be resolved once you fix your merge conflicts. Once I see that fixed, I'll take a look the function @dasnellings mentioned and help split that into smaller functions to merge this PR

cmd/samAssembler/samAssembler_test.go
```diff
@@ -0,0 +1,91 @@
...
+	ExpectedOutFileA     string
+	ExpectedOutFileB     string
+	OutMultiFaFileNames  []string
+	ExpectedMultiFaFiles []string
+	Verbose              int
+}{
+	{SamFileName: "testdata/diverged.RefAln.sorted.bam",
+		RefFile:              "testdata/ref.fa",
+		OutFileA:             "testdata/tmp.OutFileA.fa",
+		OutFileB:             "testdata/tmp.OutFileB.fa",
```
edotau, line 33: If possible, please add a test case with sequences of 20-50 bases (no more than 100), so I can have a simple example to follow the logic of your main function and perform the calculations in my head or by hand

PR #804, opened 2023-04-03
samAssembler Part 5: Scoring output alignments

rimangan:

### Testing

None

In response to Dan's comment, I cut up samAssembler into helper functions, and used structs to organize the many variables. This cut the size of the main program in half.

This PR adds the 'score' mode to samAssembler, which takes in a 5 way alignment, representing the haploid reference sequence, a 'true' diploid sequence, and the diploid sequence reconstructed from samAssembler. This mode then constructs confusion matrices based on the alignment, which will provide the basis for evaluating the precision and recall of the model. 

A confusion matrix is a 2d table like the one below, in which each column represents an actual genotype from our generative model of molecular evolution, and each row represents the corresponding predicted genotype. In other words, this matrix allows us to identify false positive and false negative variant calls, but allows us to further identify which genotype was incorrectly predicted from each pileup. 
Values along the diagonal represent correct base calls, and values off diagonal represent misclassification. Values below the diagonal represent situations where the model predicted a genotype that is more rare than the true mutation (a false positive). 
Tv1 and Tv2 refer to the two possible transversion mutations (for reference base A, this would be C or T). Tv1 and Tv2 are distinguished alphabetically, so for reference base "A", "C" would be Tv1.
image

Values above the diagonal represent when the model predicts a genotype that is more common that the true genotype (false negatives, in which the model is too conservative). For our use case, false positives are particularly bad, so when the model is uncertain, we expect that the diagonal line should decay up and to the right. 
This is indeed what we see on a series of simulated tests, in which we allow coverage to decay, or the mean read length to drop. Note on the far right how the predictions are all along the top row, homoRef. In other words, if the model does not have sufficient data to make a variant call, it will guess that the genotype is homozygous reference. In other words, this is a conservative, reference-biased variant caller.
image

We can compare our informative prior to a flat prior, in which all genotypes are equally weighted. This is the same as variant calling by likelihood ratio test. Notice how as errors accumulate, they do not show the same pattern of accumulating above the diagonal. In other words, the likelihood ratio test produces as many false positives as false negatives.
image

Finally, I ran this program on a forward evolution simulation on hg38 chr21 and chr22 to conduct a larger test. I've displayed proportions as well as the raw counts.

image

I'll mark this as ready for review, but keep in mind another part will be on the way to allow us to score INDELs. These confusion matrices score only diploid base genotypes.


cmd/samAssembler/testdata/makeTestData.readme.md
```diff
@@ -0,0 +1,79 @@
...
+~/go/bin/simulateEvol -qName evol2 -branchLength 0.05 -propIndel 0.1 -setSeed 17 -transitionBias 3 Sequence_0.fa Sequence_0.evol2.fa
+
+~/go/bin/faFilter -name Sequence_1 ref.fa Sequence_1.fa
+~/go/bin/simulateEvol -qName evol1 -branchLength 0.05 -propIndel 0.1 -setSeed 27 -transitionBias 3 Sequence_1.fa Sequence_1.evol1.fa
+~/go/bin/simulateEvol -qName evol2 -branchLength 0.05 -propIndel 0.1 -setSeed 29 -transitionBias 3 Sequence_1.fa Sequence_1.evol2.fa
+```
+  
+I then filtered out the evolved sequences from the resulting multiFa alignments and trimmmed out gaps:
+
+ ```
```
edotau, line 29:
```suggestion
 ```bash
```

cmd/samAssembler/testdata/makeTestData.readme.md
```diff
@@ -0,0 +1,79 @@
...
+I then filtered out the evolved sequences from the resulting multiFa alignments and trimmmed out gaps:
+
+ ``` 
+~/go/bin/faFilter -notName Sequence_0 Sequence_0.evol1.fa tmp.Sequence_0.evol1.fa
+~/go/bin/faFormat -noGaps tmp.Sequence_0.evol1.fa Sequence_0.evol1.diverge.noGaps.fa
+~/go/bin/faFilter -notName Sequence_1 Sequence_1.evol1.fa tmp.Sequence_1.evol1.fa
+~/go/bin/faFormat -noGaps tmp.Sequence_1.evol1.fa Sequence_1.evol1.diverge.noGaps.fa
+```
+  
+```
```
edotau, line 36:
```suggestion
```bash
```

cmd/samAssembler/testdata/makeTestData.readme.md
```diff
@@ -0,0 +1,79 @@
...
+We'll merge the diverged sequences and place them in the multiFa folder for validation later
+```
+~/go/bin/mergeMultiFa Sequence_0.evol1.fa Sequence_0.evol2.fa multiFa/Sequence_0.evol.fa
+~/go/bin/mergeMultiFa Sequence_1.evol1.fa Sequence_1.evol2.fa multiFa/Sequence_1.evol.fa
+```
+
+Next I generated 60x coverage short reads.
+(# of reads per chromosome = 2000 = 60xCoverage * 10kb Chrom Length / 2x150bp paired read length) 
+
+```
```
edotau, line 56:
```suggestion
```bash
```

cmd/samAssembler/testdata/makeTestData.readme.md
```diff
@@ -0,0 +1,79 @@
...
+~/go/bin/simulateSam -n 2000 -setSeed 9 -readLength 150 -fragmentLength 400 evol1.diverge.noGaps.fa evol1.Reads.bam
+~/go/bin/simulateSam -n 2000 -setSeed 9 -readLength 150 -fragmentLength 400 evol2.diverge.noGaps.fa evol2.Reads.bam
+rm evol1.diverge.noGaps.fa evol2.diverge.noGaps.fa
+```
+  
+Next I recovered fastq reads from the simulated bam files and merged the fastq files together
+
+```
```
edotau, line 64:
```suggestion
```bash
```

cmd/samAssembler/testdata/makeTestData.readme.md
```diff
@@ -0,0 +1,79 @@
...
+~/Software/bedtools2/bin/bedtools bamtofastq -i evol1.Reads.bam -fq evol1.readA.fastq -fq2 evol1.readB.fastq
+~/Software/bedtools2/bin/bedtools bamtofastq -i evol2.Reads.bam -fq evol2.readA.fastq -fq2 evol2.readB.fastq
+cat evol1.readA.fastq evol2.readA.fastq > merged.readA.fastq
+cat evol1.readB.fastq evol2.readB.fastq > merged.readB.fastq
+```
+  
+After this, I aligned these reads to the reference genome and sorted the reads
+
+ ```
```
edotau, line 73:
```suggestion
 ```bash
```

edotau (approved):
Looks good! 

Just generally small changes to stay consistent with good programming practices. I am a bit out of tune with regards to some of the biology so I just focused software practices and let the second review comment on the biological technical correctness.

edotau:
I also push a few lint formating lines of code. I didn't change the logic in any way, but be sure to do a git pull before you start anything else

edotau (approved):

dasnellings (approved):
The code looks good so I am approving, but from a scientific angle I think it will be important to thoroughly benchmark this against real data before making any major conclusions. Perhaps you could get a truth set from the genome in a bottle project or T2T and see how your calls compare to their final assembly using short-read illumina data they generated? 

From my experience calling somatic mutations I would expect a relatively high false-positive rate using the current approach, particularly in repetitive elements and other low complexity regions. You may come up against this more when you incorporate scoring for indels which can be particularly difficult.

It may well be that germline calling avoids many of the pitfalls we encounter with somatic calling, but some of them are likely to be universal. If you do encounter issues, you might consider weighting your posterior with information about read mapping quality, base quality scores, distance from the end of the read, and violation of phase with nearby heterozygous sites. You may also consider excluding piles in regions that are known to be difficult to map (see https://github.com/genome-in-a-bottle/genome-stratifications)

PR #810, opened 2023-04-07
Vcf filter clustered variants

dasnellings:
# Description

This PR adds a new flag to vcfFilter to remove clustered variants (i.e. any variant less than X bp away from another variant). This took a bit of gymnastics since vcfFilter was build to handle each record independent of other records, so I hijacked the input channel for vcfFilter into another function where I could do cross-record comparisons and output the passed variants to the original vcfFilter function. 

### Testing

Included


edotau (approved):

cfauci (approved):
I want other people to have a hance to go through this PR, but it looks good to me! I'm purposely not merging yet.

PR #817, opened 2023-04-13
style: gonomics spring cleanup refactor edition eau

edotau:
# Description
- Continuing to clean up files as I add a few more configurations
- added spell check to lint configs

### Testing

None

### DIFF
--- a/cmd/bedpeOverlap/bedpeOverlap.go
+++ b/cmd/bedpeOverlap/bedpeOverlap.go
@@ -14,7 +14,7 @@ import (
 )
 
 func bedpeOverlap(selectFile string, bedpeInFile string, contactOutFile string, bedSelect bool, overlapThreshold float64, overlapBoth bool) {
-	//bedpeOverlap will work with either a bedpe select file or a bed select file. First we determine which program to run.
+	// bedpeOverlap will work with either a bedpe select file or a bed select file. First we determine which program to run.
 	if bedSelect && overlapBoth {
 		SelectIsBedBoth(selectFile, bedpeInFile, overlapThreshold, contactOutFile)
 	} else if bedSelect {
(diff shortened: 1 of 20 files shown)


dasnellings (approved):

cmd/bedpeOverlap/bedpeOverlap.go
```diff
@@ -262,7 +262,7 @@ func main() {
 		log.Fatalf("Error: overlapThreshold must be used with bedSelect")
 	}
 
-	if *overlapBoth == true && *bedSelect == false {
+	if *overlapBoth {
```
rimangan, line 265:
```suggestion
	if *overlapBoth && !*bedSelect {
```

rimangan: Is this right? We don't want to fatal if overlapBoth is true and bedSelect is also true.

edotau: Ah nice catch, I misinterpreted the cases, these are positional flags by the user

cmd/bedpeOverlap/bedpeOverlap.go
```diff
@@ -257,12 +261,12 @@ func main() {
...
 
-	//should no longer be needed when overlapThresholdBedPe is implemented
-	if *overlapThreshold != 0 && *bedSelect == false {
+	// should no longer be needed when overlapThresholdBedPe is implemented
+	if *overlapThreshold != 0 && *bedSelect {
 		log.Fatalf("Error: overlapThreshold must be used with bedSelect")
 	}
 
-	if *overlapBoth == true && *bedSelect == false {
+	if *overlapBoth && *bedSelect {
```
rimangan, line 269: So I think this is going to fatal when both are true, but the fatal condition is overlapBoth==true, bedSelect==false.

edotau: my bad meant to put `!`

RavenYL (approved):
This looks good to me! Noting that we can write "return bool check" instead of "if bool check, return true"

dasnellings (approved):

rimangan (approved):

PR #823, opened 2023-04-17
Add "keepCellsOption" to fastqFilter

sweavs111:
Added an option to the fastqFilter command to filter a single cell read pair by a user-input list of cell barcodes. This is similar in structure to the keepNames option but compares on barcodes. The use case for this option is to filter the fastq files from the scSTARR-seq targeted enrichment sequencing for cells that were picked up and made it through cell ranger in the initial scRNA-seq run.

I have added testing and my option seems to pass testing, though it looks like there is another test that is failing (that looks like it is trying to compare the sequence of a read1/read2 scPair


sweavs111:
Fixed the the R2 test file, my apologies. Disregard my message about test failing. All tests pass currently

RavenYL (approved):

edotau (approved):

edotau (approved):

rimangan (approved):

PR #836, opened 2023-04-27
simulateSam: Ancient DNA Damage

rimangan:
# Description

Dependent on #830 

This pull request introduces functionality to simulate DNA damage caused by postmortem cytosine deamination present in ancient DNA libraries.
The figure below, from Briggs et al PNAS 2007, shows that cytosine deamination events (read C to T mutations on the forward strand or as G to A on the opposite strand) are concentrated near the ends of reads (note that their library prep chewed back 3' overhangs, hence the asymmetric distribution). 
image

In this PR, the user can specify an "ancientErrorRate", which is the probability that a C or G in the organism is misread as a T or A, respectively. We model the spatial concentration of deaminations with a geometric distribution, which can be controlled with the "geometricParam" option. 

Also, the spatial distribution of cytosine deaminations can be written out to an optional file using a new option.
As this program is closely related to my work on samAssembler, I have added a section to the samAssembler documentation to explain some technical details of ancientError generation. 

[samAssembly.pdf](https://github.com/vertgenlab/gonomics/files/11348297/samAssembly.pdf)

Now for some results. As the rate of cytosine deamination increases, more and more cytosine deamination events appear (area under the curve in the histogram below). As rates increase, we see an S shaped curve emerge from the geometric distribution as we saturate cytosine deaminations in the terminal positions (in other words, every C on the last base is likely to become a T if the cytosine deamination rate is too high). 
image
Here I show a density histogram instead to better highlight the change in shape of the spatial disribution of synthetic cytosine deamination events.
image

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


craiglowe (commented):
This looks good to me.  I see one place where I think we can simplify things by having a function call instead of an if-statement

simulate/sam.go
```diff
@@ -53,7 +55,7 @@ func IlluminaPairedSam(refName string, ref []dna.Base, numPairs, readLen, avgFra
 
 // generateSamReadNoFlag generates a sam record for the input position.
```
RavenYL, line 56: Can you update this documentation to reflect that generateSamReadNoFlag now generates not just a sam record but also a []int that is deaminationDistributionSlice?

simulate/sam.go
```diff
@@ -86,9 +88,14 @@ func generateSamReadNoFlag(readName string, refName string, ref []dna.Base, star
 	}
 	copy(currSam.Seq[realSeqStartIdx-start:len(currSam.Seq)-(end-realSeqEndIdx)], ref[realSeqStartIdx:realSeqEndIdx])
 
+	// If ancient DNA, we simulate cytosine deamination error
+	if ancientErrorRate > 0 {
```
RavenYL, line 92: Note to self that ancient error needs to happen before sequencing error for the simulation to make sense

simulate/sam.go
```diff
@@ -169,3 +176,42 @@ func sequencingError(currSam sam.Sam, alias numbers.BinomialAlias) sam.Sam {
...
+	numAncientErrorAttempts := numbers.RandBinomial(ancientAlias)
+	damagedPositions := make(map[int]int, numAncientErrorAttempts)
+	var currError int = 0
+	for currError < numAncientErrorAttempts {
+		distanceToEnd = numbers.RandGeometric(geometricParam)
+		for distanceToEnd >= len(currSam.Seq) {
+			distanceToEnd = numbers.RandGeometric(geometricParam)
+		}
+		if !forward {
+			currRandPos = len(currSam.Seq) - distanceToEnd - 1 // distance from end, instead of distance from start
```
RavenYL, line 193: At code review we talked about how "ancient damage" is on the DNA molecule and should be detected on both forward and reverse sam reads, unlike "sequencing error".  Does this mean there needs to be code here that says that "if !forward" not only the currRandPos needs to be mutated, but also the matching position on the forward read (and vice versa for "else", mutate matching position on both forward read and reverse read?)

simulate/sam.go
```diff
@@ -13,9 +14,10 @@ import (
 )
 
 // IlluminaPairedSam generates a pair of sam reads randomly distributed across the input ref.
-func IlluminaPairedSam(refName string, ref []dna.Base, numPairs, readLen, avgFragmentSize int, avgFragmentStdDev float64, flatErrorRate float64, binomialAlias numbers.BinomialAlias, out *fileio.EasyWriter, bw *sam.BamWriter, bamOutput bool) {
+func IlluminaPairedSam(refName string, ref []dna.Base, numPairs, readLen, avgFragmentSize int, avgFragmentStdDev float64, flatErrorRate float64, ancientErrorRate float64, flatBinomialAlias numbers.BinomialAlias, ancientBinomialAlias numbers.BinomialAlias, geometricParam float64, out *fileio.EasyWriter, bw *sam.BamWriter, bamOutput bool, deaminationDistributionSlice []int) {
```
craiglowe, line 17: Right now ancient DNA damage of a molecule is independent across the fwd and rev reads, but in reality, I think if a position is damaged, and read by both the forward and reverse reads, both should show the damage.  Is that correct?  If so, I think this would be important to capture.

craiglowe (commented):
One more comment on fwd and rev DNA damage

simulate/sam.go
RavenYL: (Note that this refers to fwd and rev reads from Illumina of the same strand of DNA, not the 2 strands of 1 DNA molecule)

rimangan: Resolved in latest commit

simulate/sam.go
```diff
@@ -13,21 +13,50 @@ import (
...
+// output file handles for sam, bam, and a bool denoting if bam (or sam) should be the output.
+// Whichever handle (sam or bam) is not being used can be nil.
 func IlluminaPairedSam(refName string, ref []dna.Base, numPairs, readLen, avgFragmentSize int, avgFragmentStdDev float64, flatErrorRate float64, ancientErrorRate float64, flatBinomialAlias numbers.BinomialAlias, ancientBinomialAlias numbers.BinomialAlias, geometricParam float64, out *fileio.EasyWriter, bw *sam.BamWriter, bamOutput bool, deaminationDistributionSlice []int) {
-	var fragmentSize, midpoint, startFor, startRev, endFor, endRev int
+	var fragmentSize, midpoint, startFor, endRev, newCapacity int
 	var currFor, currRev sam.Sam
+	var fragment []dna.Base = make([]dna.Base, 0, avgFragmentSize+int(5*avgFragmentStdDev)) //set initial fragment slice capacity to 5 deviations above average fragment size
+	var newFragment []dna.Base                                                              //this will be used if we run out of capacity in 'fragment'
+	if avgFragmentSize < readLen {
+		log.Fatalf("Error: average fragment size %v is less than read length %v.\n", avgFragmentSize, readLen)
```
RavenYL, line 28: Can you help me better understand what "fragment" is? Is it a subset of the whole input DNA sequence (e.g. a region on Neanderthal chr1)? Is fragment used as a "unit" to generate matching ancient damage on both forward and reverse reads?

Why is it a fatal error if the average fragment size is less than read length? Is this something the user should take care to avoid before they feed in the inputs to this IlluminaPairedSam function?

simulate/sam.go
```diff
@@ -13,21 +13,50 @@ import (
...
+
+		//here we check if fragment has sufficient capacity for the next fragment we drew. If not, we extend capacity.
+		if len(fragment)+fragmentSize > cap(fragment) {
+			newCapacity = len(fragment) + fragmentSize
+			newFragment = make([]dna.Base, fragmentSize, newCapacity)
+			fragment = newFragment
+		} else {
+			fragment = fragment[:fragmentSize]
+		}
+		copy(fragment, ref[startFor:endRev])
```
RavenYL, line 49: It looks like the fragmentSize can be smaller than readLen, in which case readLen becomes fragmentSize (but average fragmentSize can't be smaller than readLen). Why is this?

What do "fragment capacity, fragment length, and fragment size" mean?

simulate/sam.go
RavenYL: Thank you! I am approving and merging since I don't think the code needs to change, but I have 2 followup questions.

Now I understand that the capacity refers to the memory allocated to the fragment. Does fragmentSize represent the number of bases (int), and len(fragment) measures how much memory the fragment takes up (length of []dna.Base))?

My experience has been that when fragmentSize < readLength, full-length reads are still produced, but the bases after the fragment ends are all empty calls (e.g. G calls for certain chemistries). Do you experience truncated reads?

RavenYL (approved):

PR #842, opened 2023-05-24
fix issue trying to write cigar * for unmapped reads

dasnellings:
Currently there is an error when trying to write bam files with unmapped reads that use "*" as cigar instead of nil. This fixes the issue


edotau (approved):

craiglowe (approved):
Thank you!

PR #843, opened 2023-05-25
flag fix

dasnellings:
Our sam flags are backwards for IsPosStrand. Bits 16 and 32 are true if the sequence is being reverse complemented so it should be false if the read is mapping to the positive strand.


edotau (approved):

cmd/samToBed/testdata/test2.bed
```diff
@@ -1,2 +1,2 @@
-ref	28	45	r003
-ref	36	45	r001
+ref	3	33	r003
```
craiglowe, line 1: I am not sure why the samToBed test files are changing.  I don't think messing up the seq being from the reverse strand should mess with the coordinates.  I also quickly looked at test.sam, and it seems like 28 and 36 should be the start coordinates in test2.bed.

cmd/samToWig/testdata/test2.wig
```diff
@@ -2,31 +2,6 @@ fixedStep chrom=chr1 start=1 step=1
 0
 0
 0
-0
```
craiglowe, line 5: I didn't look into the details, but I am surprised that test2.wig is changing (similar to test2.bed above where I did look into it quick).  I wouldn't think this output is dependent on the dna seq in the sam

craiglowe (commented):
Great to get this bug in the flag fixed!  Thank you!  I am not sure I understand why the test files are being changed.  I quickly looked into test2.bed, and the present start coordinates look correct to me compared to the new start coordinates, which are being proposed in this PR.  I am also unsure why test2.wig would change with this bug being fixed.

cmd/samToBed/testdata/test2.bed
RavenYL: I wonder if the changes have something to do with fragLength?
Since test1 didn't change in samToBed or samToWig, and only test2 changed,
and looking in the _test.go files fragLength is set to 30 (not the default -1) only in test2 not test1.
fragLength matters in a function in gonomics/convert/convert.go: SamToBedFrag (as opposed to SamToBed).
The way SamToBedFrag works does depend on positive/negative strand.

cmd/samToBed/testdata/test2.sam
@SQ	SN:ref	LN:45
r003	16	ref	29	30	6H5M	*	0	0	TAGGC	*
r001	83	ref	37	30	9M	=	7	-39	CAGCGCCAT	*

convert/convert.go
		answer = bed.Bed{Chrom: s.RName, Name: s.QName, FieldsInitialized: 4}
		if sam.IsPosStrand(s) {
			answer.ChromStart = int(s.Pos - 1) **// this was the 29-1=28 in the old test2.bed**
			answer.ChromEnd = numbers.Min(answer.ChromStart+fragLength-cigar.NumInsertions(s.Cigar)+cigar.NumDeletions(s.Cigar), reference[answer.Chrom].Size)
			answer.Strand = bed.Positive
		} else {
			answer.ChromEnd = int(s.Pos-1) + cigar.ReferenceLength(s.Cigar)
			answer.Strand = bed.Negative
			answer.ChromStart = numbers.Max(answer.ChromEnd-(fragLength-cigar.NumInsertions(s.Cigar)+cigar.NumDeletions(s.Cigar)), 0) **// this is the numbers.Max(33-(30-0+0),0) = 3 in the new test2.bed**
		}
		return answer

So maybe the new test2.bed is right?

dasnellings: Yes, as Raven pointed out, they changed because both call the `convert.SamToBedFrag` function which is dependent on strand (specifically for handling reference length when the cigar contains indels).

craiglowe:
ok, yes, I see that now.  I had forgotten about the frag length option

craiglowe (approved):
Great fix!

RavenYL (approved):

PR #850, opened 2023-06-20
Example tool

cfauci:
example function for readme and to point out in the readme for users.


RavenYL (dismissed):

craiglowe:
Can we use one of our simpler existing tools at the example?  Maybe add a comment to each line in the program about what that line does.  Maybe something like bedToFasta?
I worry that this being called newTool instead of something like fastaTranslate will have it gather dust and not be used.

cmd/bedDistanceFromChrEnds/bedDistanceFromChrEnds.go
```diff
@@ -13,26 +13,26 @@ import (
...
-		lengthFromEnd = ref[records[i].Chrom].Size - records[i].ChromEnd
-		if lengthFromEnd < 0 {
+		lengthFromEnd = ref[records[i].Chrom].Size - records[i].ChromEnd //calculates the distance from the right termination of the chrom to the end of this bed record
+		if lengthFromEnd < 0 {                                           //throw an error if the length is less than 0
 			log.Fatalf("inputBed coordinates are outside chrom.sizes coordinate range, %s", records[i])
 		}
-		records[i].Score = numbers.Min(lengthFromEnd, records[i].ChromStart)
-		if records[i].FieldsInitialized < 5 {
+		records[i].Score = numbers.Min(lengthFromEnd, records[i].ChromStart) //calculate if the beginning of the bed is closer to the start than the end of the bed is to the end of the chrom and store that distance in the score field of the bed that will be written out
+		if records[i].FieldsInitialized < 5 {                                // output bed should have 5 fields (more information about the bed fields int eh codebase can be found in gonomics/bed/bed.go
```
craiglowe, line 31: typo

cmd/exampleTool/exampleTool.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	fileio.Write(output, outRecord) //write the contents of the outRecord variable to the file specified in the arguments.
+
+}
+
+// example of usage functions for commands in the gonomics/cmd directory
+func usage() {
+	fmt.Print(
+		"newTool - takes a bed and fasta and converts the bed sequences into amino acide sequences\n" +
```
craiglowe, line 43: can we call this faTranslate instead?

cmd/exampleTool/exampleTool.go
```diff
@@ -0,0 +1,69 @@
...
+			"newTool bedFile fastaFile\n" +
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+//example main function for gonomics/cmd directory
+func main() {
+	var expectedNumArgs int = 3
+
+	flag.Usage = usage
```
craiglowe, line 54: Can we add comments for the lines in this function as well?

cmd/exampleTool/exampleTool.go
```diff
@@ -0,0 +1,69 @@
...
+	"flag"
+	"fmt"
+	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/dna"
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/fileio"
+	"log"
+)
+
+// bedToAminoAcid is  complete function made from gonomics packages, it takes a bed and a fasta and returns the amino acid sequence of all the beds together.
```
craiglowe, line 13: extra space

craiglowe (commented):
Wonderful job!  I had just a few small comments about typo and some tiny changes

cmd/exampleTool/exampleTool.go
cfauci: good catch , I'll match the names

cmd/exampleTool/exampleTool.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	fileio.Write(output, outRecord) //write the contents of the outRecord variable to the file specified in the arguments.
+
+}
+
+// example of usage functions for commands in the gonomics/cmd directory
+func usage() {
+	fmt.Print(
+		"bedToAminoAcid - takes a bed and fasta and converts the bed sequences into amino acid sequences\n" +
```
craiglowe, line 43: bedToAminoAcid sounds like a good name.  Can we get a:
mkdir cmd/bedToAminoAcid
git mv cmd/exampleTool/exampleTool.go cmd/bedToAminoAcid/bedToAminoAcid.go
so that it becomes a program with that name?

craiglowe (commented):
One final ask

cmd/exampleTool/exampleTool.go
craiglowe: Actually, I take that back.  Maybe it is good to leave it in the exampleTool directory, but let's rename the actual file

craiglowe (approved):
Wonderful!

RavenYL (approved):

PR #852, opened 2023-06-23
restoring keepAllNames option

cfauci:
allows beds to be merged and names to be kept from all merged regions in a comma separated list in the names column. I've used this successfully for months, and it existed previously, but somehow never made it onto main.

### DIFF
--- a/bed/modify.go
+++ b/bed/modify.go
@@ -70,7 +70,7 @@ func MergeLowMem(b <- chan Bed, mergeAdjacent bool) <- chan Bed {
 
 // MergeHighMem retains input Bed entries that are non-overlapping with other input bed entries and merges together overlapping bed entries.
 // Merged bed entries will retain the maximum score in the output.
-func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
+func MergeHighMem(records []Bed, mergeAdjacent bool, keepAllNames bool) []Bed {
 	var outList []Bed
 	if len(records) == 0 {
 		return records //empty and nil slices are returned as is.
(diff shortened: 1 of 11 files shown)


cmd/bedMerge/bedMerge.go
```diff
@@ -50,9 +49,29 @@ func bedMergeLowMem(infile string, outfile string, mergeAdjacent bool) {
 	exception.PanicOnErr(err)
 }
 
-func bedMergeHighMem(infile string, outfile string, mergeAdjacent bool) {
-	var records []bed.Bed = bed.Read(infile)
-	outList := bed.MergeHighMem(records, mergeAdjacent)
```
RavenYL, line 55: The code looks good to me but I just wonder if the bed.MergeHighMem function is now removed from the bed package?

bed/modify.go
```diff
@@ -84,6 +84,9 @@ func MergeHighMem(records []Bed, mergeAdjacent bool) []Bed {
 				currentMax.Score = records[i].Score
 			}
 			currentMax.ChromEnd = numbers.Max(records[i].ChromEnd, currentMax.ChromEnd)
+			if keepAllNames && records[i].Name != "" || keepAllNames && currentMax.Name != "" {
```
craiglowe, line 87: Does this work when currentMax.Name is blank?  It seems like we would then end up with extra commas at that start.  I think we want to something like:
if keepAllNames && records[i].Name != "" { // do we have a name to add
   if currentMax.Name != "" { // is there already a name there to add on to
      currentMax.Name = currentMax.Name + "," + records[i].Name
   } else { // there is no name and this will be the first one
      currentMax.Name = records[i].Name
   }
}

craiglowe (commented):
I had one small comment on the logic.  It looks like there is also some change to the spacing in lastzwriter?

bed/modify.go
cfauci: good catch, I was only using this with files that absolutely all had names

craiglowe (approved):
Wonderful!

RavenYL (approved):

PR #858, opened 2023-07-03
Godocs for the interval package

sweavs111:
# Description

[Description]

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


interval/query.go
```diff
@@ -10,12 +10,14 @@ import (
 	"github.com/vertgenlab/gonomics/vcf"
 )
 
+// GoReadToChan reads Lift interfaces to a channel from an input file (bed, axt, vcf, sam, chain).
```
craiglowe, line 13: Should this say "Interval" interfaces instead of "Lift" interfaces?

craiglowe (commented):
Had one quick comment

craiglowe (approved):
Thank you!

interval/interval.go
```diff
@@ -12,6 +12,7 @@ import (
 	"golang.org/x/exp/slices"
 )
 
+// Interval is a type interface for any genomic datatype that has chromosomal coordinate info (chromosome, start, end)
```
dasnellings, line 15:
```suggestion
// Interval is a type interface for any genomic datatype that has chromosomal coordinate info (chromosome, start, end)
// Note that the methods must return positions for a 0-base half-closed interval [start, end)
```

We should also give some specifics about the implementation since this is something users will need to interface with if they want to use interval for unsupported file types.

craiglowe: To help people out we could say something like, "In order to satisfy this interface, a type must have GetChrom(), GetChromStart(), and GetChromEnd() methods."  Would that be good?

craiglowe (approved):
Added a quick response to a comment

craiglowe (dismissed):
Great!

cfauci (approved):

cfauci:
conflict with one of the comments in this package, should be easy to fix on your local

craiglowe (approved):
Great work!

dasnellings (approved):

PR #859, opened 2023-07-03
Tragedy of the Commons

rimangan:
# Description
This PR removes the common package from Gonomics and moves its remaining functions to "cast", a subpackage of numbers.
[Description]

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


craiglowe:
Thank you for doing this!  I wonder if "cast" is a little vague on what something like cast.StringToInt("9") would return.  Maybe 10 or maybe 57 (ascii value).  It seems unclear if it is keeping the value of the bits the same, but changing how they are interpreted, or if it is changing the value of the bits to keep the high-level value of the variable the same.  Would something like "convert" or "parse" (I think they are all StringToSomething) be more clear.  Maybe it doesn't live in "numbers" since it also does things like convert to bool

cfauci (dismissed):

craiglowe (approved):
Great to finally remove "common" from the code base!

dasnellings (approved):

PR #860, opened 2023-07-05
intervalOverlap test fix

dasnellings:
Fixes broken test in intervalOverlap


craiglowe (commented):
All I see here is changing the default behavior from unmerged output to merged output, but I am not sure we want to make that change.  Can we instead change the test?

dasnellings:
> All I see here is changing the default behavior from unmerged output to merged output, but I am not sure we want to make that change. Can we instead change the test?

This is just changing the test. The default setting for the program is still to not have merged output. One of the optimizations I made was to not return a bool instead of the set of all overlapping intervals if we are not writing the overlapping set (i.e. merged output == true). Since the test was checking the identity of the overlap, the test needs to be run with merged output = true.

craiglowe (approved):
Thank you!

PR #879, opened 2023-07-07
updates to vcf docs

craiglowe:
# Description
Added documentation to the vcf package

## Relevant Links

### Testing

None

### Screenshots & Media

None

None

### DIFF
--- a/vcf/fix.go
+++ b/vcf/fix.go
@@ -6,12 +6,20 @@ import (
 	"github.com/vertgenlab/gonomics/dna"
 )
 
+// FixAllVcf runs FixVcf on each element in a slice of vcf structs.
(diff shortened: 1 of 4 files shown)


rimangan:
Can gVcf be removed? It is not called elsewhere. I'm not sure of the history of the alleleSplit cmd,  which appears to be the only cmd that uses the other code in the gVcf.go file

craiglowe:
I do think we should remove gVcf.  I think we expanded the vcf struct and functions to do what gVcf was doing.  I think it might be good to leave it in place for now, but it we should spend one of the next code reviews on the vcf package because that is one of the messy areas in the code base right now

cfauci (approved):

rimangan (approved):

PR #896, opened 2023-07-24
bedFormat FdrScoreAnnotation

rimangan:
# Description
This is a companion PR to the 'faFindFast' long output PR. The idea here is in faFindFast, we wind up with a raw P value (a log transformed float64) in the first annotation field.
This PR adds an option to bedFormat to accept any bed file with raw p values in an annotation field and convert them to FDR adjusted p values. Note that an FDR pAdj follows this formula:

$$
pAdj = rawP \cdot \frac{totalComparisons}{Rank}
$$

rawP is provided in the input bed lines. totalComparisons can be calculated as the number of bed lines in the input file. To calculate rank, we must first generate a list of counts for every score (ex. -log10pvalue 4: count 3, 3: 12, 2:21, 1:45, 0:205).
From this example count vector, we can calculate ranks (for example, -log10pvalue 2 would be rank 21+12+3=36).
This program does a first pass through the input file to get the counts, allowing it to calculate the Rank. Then, on a second pass, the program is able to append the FDR-adjusted P values.

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


cfauci (approved):

cmd/bedFormat/testdata/expected.fdrAnnotation.bed
```diff
@@ -0,0 +1,96 @@
+chr1	0	10	chr1_0	2	+	2.000000e+01	1.019461e+00	5.423397e-01
```
RavenYL, line 1: Can I ask what are the 3 columns after + (strand)? Are they all annotation? I thought there would be 2 columns of annotation, raw p value and adjusted p value?

cmd/bedFormat/bedFormat.go
```diff
@@ -99,7 +133,64 @@ func bedFormat(s Settings) {
...
+		var fdrSlice []FdrConverter = make([]FdrConverter, 0)
+		for i := range fdrMap {
+			fdrSlice = append(fdrSlice, fdrMap[i])
+		}
+
+		sort.Slice(fdrSlice, func(i, j int) bool { return compareFdrConverter(fdrSlice[i], fdrSlice[j]) == -1 })
+
+		for i := len(fdrSlice) - 1; i >= 0; i-- {
+			currRank += fdrSlice[i].Count
+			fdrSlice[i].Rank = currRank
```
RavenYL, line 170: Note to self: 
- fdrSlice was sorted, by compareFdrConverter (	-1 means a.RawPValue < b.RawPValue, so rawValues are increasing aka ascending order after sorted, reference: https://pkg.go.dev/sort). 
- Then, attach currRank to fdrSlice entry in descending order, because we will start with the last entry, which is the largest value, and has rank 0 (and after that, smaller values, rank 1, 2, 3, etc.)
- take into account that currRank needs to include count of that entry
- then write out fdrSlice[i]'s true currRank

cmd/bedFormat/bedFormat.go
```diff
@@ -99,7 +133,64 @@ func bedFormat(s Settings) {
...
+		var currRank int = 0
+		var fdrSlice []FdrConverter = make([]FdrConverter, 0)
+		for i := range fdrMap {
+			fdrSlice = append(fdrSlice, fdrMap[i])
+		}
+
+		sort.Slice(fdrSlice, func(i, j int) bool { return compareFdrConverter(fdrSlice[i], fdrSlice[j]) == -1 })
+
+		for i := len(fdrSlice) - 1; i >= 0; i-- {
+			currRank += fdrSlice[i].Count
```
RavenYL, line 169: Does the absolute number of currRank matter in adjusting p-value?

Right now, it seems like if i=len(fdrSlice)-1, fdrSlice[i].Count=2, since we start with currRank=0, then currRank becomes currRank += fdrSlice[i].Count (0+2=2), so "the 1st highest raw p-value is now ranked 2nd because it appears 2 times". Is this intended behavior? Or do we want "the 1st highest raw p-value is ranked 1st, but the 2nd highest raw p-value will be ranked 3rd because the 1st highest raw p-value appears 2 times", in which case I think we would do fdrSlice[i].Rank = currRank before currRank += fdrSlice[i].Count, aka swapping lines 169 and 170

cmd/bedFormat/bedFormat.go
```diff
@@ -99,7 +133,64 @@ func bedFormat(s Settings) {
...
+			fdrSlice[i].Rank = currRank
+
+			// note here I move from -log10 space to log10 space for readability
+			fdrSlice[i].RawPValue = fdrSlice[i].RawPValue * -1                                                          //from -log10p to log10p
+			fdrSlice[i].AdjPValue = fdrSlice[i].RawPValue + math.Log10(float64(totalWindows)/float64(fdrSlice[i].Rank)) //note that addition is logspace multiply
```
RavenYL, line 174: Can I ask why adjPValue is calculated with rawPValue*(totalWindows)/(rank)? I think this means that the more total windows there are, the bigger adjPValue becomes, and the less likely it will be significant. I thought it would be the other way around?

cmd/bedFormat/bedFormat.go
RavenYL: (note to self: higher rank means lower p-value)

RavenYL (approved):

PR #898, opened 2023-07-24
Straw to bedpe

cfauci:
# Description
made the hic package, created some reading and comparison functions and made testing for Read function, passes tests
## Relevant Links

### Testing

None

### Screenshots & Media

None

None


hic/hic.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	for i = range a {
+		equal = Equal(a[i], b[i])
+		if !equal {
+			return equal
+		}
+	}
+
+	return equal
```
rimangan, line 81: It might be more readable to 'return true' here and to 'return false' on line 77.

hic/hic.go
```diff
@@ -0,0 +1,82 @@
...
+	return current
+}
+
+// Equal compares two Straw structs to see if the values are identical and returns a bool
+func Equal(a Straw, b Straw) bool {
+	var equal bool
+	if a.Bin1Start == b.Bin1Start && a.Bin2Start == b.Bin2Start && a.ContactScore == b.ContactScore {
+		equal = true
+	} else if a.Bin1Start == b.Bin2Start && a.Bin2Start == b.Bin1Start && a.ContactScore == b.ContactScore { //bins don't have restrictions of ordering
+		equal = true
```
rimangan, line 57: instead of equal=true this line could just read 'return true'. No need for the equal variable I think

hic/testdata/strawTestFile.straw
```diff
@@ -0,0 +1,2 @@
+1000	5000	5
```
rimangan, line 1: Might be good to switch Bin1Start and Bin2Start for one of these (to test for the lack of restriction on ordering)

hic/hic.go
cfauci: good point, I'll update

hic/testdata/strawTestFile.straw
cfauci: I changed the values in the data struct instead of the file

rimangan (approved):

sweavs111 (approved):
looks great!

PR #902, opened 2023-07-25
bedShear

rimangan:
# Description
This PR introduces a new command called bedShear, which splits an input bed file into fixed size fragments. Think of this as an inverse to 'bedMerge'. The motivation for this was that liftOver often fails to lift bed elements with INDELs between references. By splitting beds into small fragments (or even individual bases), we can circumvent this restriction. The specific use case is I have a bed of all regions in hg38 where we had alignments between hg38/panTro6/etc. I wanted to visualize these beds in T2T, but noticed many of these beds failed to lift.

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


craiglowe (approved):
Great!

RavenYL (approved):

PR #903, opened 2023-07-27
Biased Reconstruction and INDEL support in reconstructSeq

rimangan:
# Description
This PR replaces 'primateRecon', which performs ancestral reconstruction of the HCA node based on a fixed topology alignment, with the existing program 'reconstructSeq', which can perform ancestral reconstruction with any newick tree and multiple alignment. 
However, primateRecon has several features which are not supported by reconstructSeq on main.
First, reconstructSeq estimates ancestral bases by maximum likelihood. However, when estimating divergence between the modern human and HCA sequence, we have previously biased our reconstruction towards the modern human sequence, to create a more conservative divergence estimate (the human sequence essentially serves as our prior). 
This PR replicates the reference-biased reconstruction options in primateRecon in reconstructSeq.
There are two types of bias filters the user can apply: nonBiasProbThreshol and highestProbThreshold. With highestProbThreshold, an ancestral base will be converted to dna.N unless the probability of that base is above some threshold (in the HAQER paper we used this for phasing ancestral/derived segregating sites for dAFS analysis).
With nonBiasProbThreshold, we return the reference base unless the sum of probabilities for non-reference bases is above a threshold (this is the option used to generate the HCA sequence we use to screen for HAQERs).

The second problem with reconstructSeq is it does not support INDELs, unlike primateRecon. This PR adds INDEL support to reconstructSeq, using the same rule as in primateRecon. 

All nodes in an Etree have three edges: Left, Right, and Up. A base (A,C,G,T,or N) is considered present (as opposed to absent, where node.Fasta.Seq[pos] == dna.Gap) if two of these three edges lead to a base in that position somewhere in the tree.
For example, consider the tree human, chimp, bonobo, gorilla, and orangutan. The human-chimp ancestor would have a base if human and bonobo have a base and chimp,gorilla, and orangutan have gaps, as two of the three edges lead to bases (left leads to human, right leads eventually to bonobo).

Note that with this implementation, reconstructSeq now fully recapitulates the functionality of primateRecon, but on general trees (see figure).
image
image

As such, I have move primateRecon and mouseRecon, to a subdirectory of cmd called 'deprecated'. These programs are still functional, but i strongly advise the exclusive use of reconstructSeq going forward.

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


reconstruct/reconstruct.go
```diff
@@ -326,31 +337,115 @@ func BubbleUp(node *expandedTree.ETree, prevNode *expandedTree.ETree, scrap []fl
...
-		fix := FixFc(root, internalNodes[k])
-		yHat := Yhat(fix)
-		internalNodes[k].Fasta.Seq = append(internalNodes[k].Fasta.Seq, []dna.Base{dna.Base(yHat)}...)
+	DescendentBaseExists(root, position)
+	BaseExists(root, position)
+	for k := range internalNodes {
+		fix = FixFc(root, internalNodes[k])
+
+		if internalNodes[k].BasePresent {
+			if internalNodes[k].Name == biasParentName {
```
RavenYL, line 439: If biasLeafName == "", then biasParentName is initialized string, which is also empty (?), so "if internalNodes[k].Name == biasParentName" will always be false, and will get unbiased estimate. Is that correct?

cfauci (dismissed):

RavenYL (approved):

reconstruct/reconstruct.go
```diff
@@ -326,31 +337,118 @@ func BubbleUp(node *expandedTree.ETree, prevNode *expandedTree.ETree, scrap []fl
...
-		fix := FixFc(root, internalNodes[k])
-		yHat := Yhat(fix)
-		internalNodes[k].Fasta.Seq = append(internalNodes[k].Fasta.Seq, []dna.Base{dna.Base(yHat)}...)
+	DescendentBaseExists(root, position)
+	BaseExists(root, position)
+	for k := range internalNodes {
+		fix = FixFc(root, internalNodes[k])
+
+		if internalNodes[k].BasePresent {
+			if biasParentName != "" && internalNodes[k].Name == biasParentName {
```
craiglowe, line 442: Would it simplify the code a little bit to have this if-else statement just set biasBase to either base in the fasta or dna.N, and then have the function call after the if-else?  It seems like biasBase is the only thing that needs to be set.  Right now we overwrite nonBiasThreshold with a zero if biasParentName is not given, but maybe we don't need that?  Maybe instead we have a check and a fatal if someone calls this with a no parentName, but a non-zero value of nonBiasBaseThreshold?

reconstruct/reconstruct.go
```diff
@@ -326,31 +337,118 @@ func BubbleUp(node *expandedTree.ETree, prevNode *expandedTree.ETree, scrap []fl
...
+				biasBase = biasLeafNode.Fasta.Seq[position]
+				answerBase = LikelihoodsToBase(fix, nonBiasBaseThreshold, biasBase, highestProbThreshold) //biased estimate
+			} else {
+				answerBase = LikelihoodsToBase(fix, 0, dna.N, highestProbThreshold) //unbiased estimate
+			}
+		} else {
+			answerBase = dna.Gap
+		}
+
+		internalNodes[k].Fasta.Seq = append(internalNodes[k].Fasta.Seq, []dna.Base{answerBase}...)
```
craiglowe, line 452: I think we are just adding one base to the slice, right?  So we can do:
internalNodes[k].Fasta.Seq = append(internalNodes[k].Fasta.Seq, answerBase)

craiglowe (commented):
Wonderful!  I had just a few very small comments.

reconstruct/reconstruct.go
```diff
@@ -326,31 +337,118 @@ func BubbleUp(node *expandedTree.ETree, prevNode *expandedTree.ETree, scrap []fl
...
+		if node.Right != nil {
+			DescendentBaseExists(node.Right, pos)
+		}
+		node.DescendentBasePresent = node.Left.DescendentBasePresent || node.Right.DescendentBasePresent
+	}
+}
+
+// LoopNodes performs ancestral sequence reconstruction for all nodes of an input tree, specified by an input root node,
+// at a user-specified alignment position.
+// Options: the user may specify a 'biasLeafName'. When specified, the reconstruction of this sequence's immediate ancestor
```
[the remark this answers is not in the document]
rimangan: We have developed theory for lineage-biased reconstruction off internal nodes (this was the dot product between normalized likelihood vectors), but i haven't implemented it in code. This PR just does leaf-biased reconstruction. I plan to handle that generalization in a separate PR.

reconstruct/reconstruct.go
```diff
@@ -326,31 +337,118 @@ func BubbleUp(node *expandedTree.ETree, prevNode *expandedTree.ETree, scrap []fl
...
+// will be biased towards that descendent. The degree of this bias is controlled by the option 'nonBiasBaseThreshold'.
+// The user may also specify a 'highestProbThreshold'. If the program is uncertain about ancestral reconstruction for a particular
+// node, this option will allow LoopNodes to return an 'N' for that node instead.
+func LoopNodes(root *expandedTree.ETree, position int, biasLeafName string, nonBiasBaseThreshold float64, highestProbThreshold float64) {
```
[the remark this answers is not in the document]
rimangan: Next commit catches this at the level of the cmd file.

reconstruct/reconstruct.go
rimangan: resolved in coming commit

reconstruct/reconstruct.go
rimangan: ParentName is determined earlier in the function based on whether the bias node has a parent, and fails otherwise. We have the check and fatal for no LeafName / non-zero value of nonBiasBaseThreshold in the cmd. I'm not sure if that simplification would give us the same behavior, and I plan to generalize this in a coming PR anyway and expect much of this to be replaced.

RavenYL (approved):

edotau (approved):

PR #904, opened 2023-07-31
Numerical Stability for faFindFast longOutput

rimangan:
# Description
This PR performs RawPValue calculations for faFindFast in logSpace, instead of in normal space, which ensures numerical stability for small values.
## Relevant Links

### Testing

None

### Screenshots & Media

None

None

### DIFF
--- a/cmd/faFindFast/efficient.go
+++ b/cmd/faFindFast/efficient.go
@@ -5,9 +5,9 @@ import (
 	"github.com/vertgenlab/gonomics/dna"
 	"github.com/vertgenlab/gonomics/exception"
 	"github.com/vertgenlab/gonomics/numbers"
+	"github.com/vertgenlab/gonomics/numbers/logspace"
 	"io"
 	"log"
-	"math"
 )
 
 // incrementWindowEdge takes two aligned fasta sequences and a current index into the alignment.
@@ -73,6 +73,8 @@ func speedyWindowDifference(windowSize int, reference []dna.Base, query []dna.Ba
 	var gapOpenCloseRef, gapOpenQuery, gapClosedQuery, numRefNs, numQueryNsGap, numQueryNsMatch, numSubst int // ints we will get back when moving the window one ref base
 	var err error
 	var percentDiverged, rawPValue float64
+	var scorePValueCache map[int]float64 = make(map[int]float64) //for each score, we cache the RawPValue for performance
+	var foundInMap bool
 
 	for lastAlnIdxOfWindow < len(reference) { // this check could also be "!done", I am not sure what is more clear
 		// we always move the lastBaseOfTheWindow (right side) and add on what we find to the counters because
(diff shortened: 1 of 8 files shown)


numbers/distribution_test.go
```diff
@@ -31,13 +31,13 @@ func TestBinomialDist(t *testing.T) {
...
+	input1 := BinomialLeftSummation(20, 1, 0.6, false)
 	expected1 := 3.408486e-07
-	input2 := BinomialLeftSummation(20, 20, 0.6)
+	input2 := BinomialLeftSummation(20, 20, 0.6, false)
 	expected2 := 1.000000e+00
-	input3 := BinomialRightSummation(20, 4, 0.6)
+	input3 := BinomialRightSummation(20, 4, 0.6, false)
 	expected3 := 9.999527e-01
-	input4 := BinomialRightSummation(20, 16, 0.4)
+	input4 := BinomialRightSummation(20, 16, 0.4, false)
```
RavenYL, line 40: Maybe a test for logOutput = true would be helpful for the future?

RavenYL (dismissed):

RavenYL (dismissed):

cmd/faFindFast/efficient.go
```diff
@@ -73,6 +73,10 @@ func speedyWindowDifference(windowSize int, reference []dna.Base, query []dna.Ba
 	var gapOpenCloseRef, gapOpenQuery, gapClosedQuery, numRefNs, numQueryNsGap, numQueryNsMatch, numSubst int // ints we will get back when moving the window one ref base
 	var err error
 	var percentDiverged, rawPValue float64
+	var scorePValueCache map[int]float64
```
craiglowe, line 76: I think we could use some comments for this line to know what values the cache is mapping between, and also what a divergence rate of -1 means.  I think this function could also use a comment.

cmd/faFindFast/efficient.go
```diff
@@ -120,3 +124,25 @@ func speedyWindowDifference(windowSize int, reference []dna.Base, query []dna.Ba
...
+
+	for k := n - 1; k >= 0; k-- {
+		answer[k] = logspace.Add(numbers.BinomialDistLog(n, k, p), answer[k+1])
+	}
+
+	// now convert to -log10 space
+	for k := 0; k <= n; k++ {
+		answer[k] = -1 * logspace.ToBase10(answer[k])
+	}
+	answer[0] = 0 //hardcoded to avoid numerical noise
```
craiglowe, line 145: How much is this noise?  Maybe there should be a check here that if answer[0] is too far from 1 or 0 (depending on when the check happens) that we die.  Maybe we check to make sure it is between 0.99 and 1.01?

numbers/distribution.go
```diff
@@ -224,75 +228,76 @@ func PoissonSum(left int, right int, lambda float64) float64 {
...
+	var answer, _ = BinomialDist(n, left, p, logOutput)
 	var curr float64
 	for i := left; i <= right; i++ {
-		curr, _ = BinomialDist(n, i, p)
-		answer += curr
+		curr, _ = BinomialDist(n, i, p, logOutput)
+		if logOutput {
+			answer = logspace.Add(answer, curr)
+		} else {
+			answer += curr
```
craiglowe, line 269: Is there an advantage to doing all this math in non-log-space?  Seems like we could do it all in log-space and then just convert at the end?  Maybe that is slower since logspace.Add is slower than normal addition?

numbers/distribution.go
```diff
@@ -224,75 +228,76 @@ func PoissonSum(left int, right int, lambda float64) float64 {
...
 }
 
 // BinomialRightSummation calculates the sum of binomial probabilities to the right of k successes for a binomial distribution with n experiments and a success probability of p, inclusive.
-func BinomialRightSummation(n int, k int, p float64) float64 {
-	if k > n/2 {
-		return evaluateRightBinomialSum(n, k, p)
-	} else {
-		return 1 - evaluateLeftBinomialSum(n, k-1, p)
+func BinomialRightSummation(n int, k int, p float64, logOutput bool) float64 {
+	if n == k {
```
craiglowe, line 244: Do we need this special case?

numbers/distribution.go
craiglowe: This comment also applies to some of the functions below, but the more I think about it the less clear I am that we want to be in log-space if we will eventually get out of it and have to do lots of addition operations

craiglowe (commented):
Wonderful!  Just a few small comments

cmd/faFindFast/efficient.go
rimangan: I tested without this line and the answer I got back was 3.616228e-17. In other words, very minor noise

numbers/distribution.go
rimangan: Yes, exactly like you said, logspace.Add is significantly slower than addition. Normal space calculations are therefore much faster. The big question is whether the user is concerned if an area under the curve is reported as 0 if we hit underflow. If they don't care, the regular space calculation will be much faster. Note that when logOutput is true, we never get out of logSpace, so we're not doing all the logspace.Add operations for nothing

RavenYL (approved):

PR #905, opened 2023-07-31
draft of Cellranger bam

sweavs111:
This is a draft of a program that takes in a bam file from cellranger and will pull out reads that are "representative" for that UMI. Default it will print the cellbarcode for that cell and the construct that that UMI is associated with. There is an option where the program will add up all the construct counts to create a pseudobulk and there is a further option for input normalization.

The intended use case for this program is the create psuedobulk values for STARR-seq data. I should probably change the name of the command but I wasn't really thinking ahead when I created it.

### DIFF
--- a/cmd/cellrangerBam/cellrangerBam.go
+++ b/cmd/cellrangerBam/cellrangerBam.go
@@ -1,31 +1,61 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"github.com/vertgenlab/gonomics/fileio"
 	"github.com/vertgenlab/gonomics/sam"
+	"log"
 )
 
-func main() {
+func parseBam(inSam string, outTable string) {
 	var k int = 0
 	var bit int32
-	var umis []string
+	var constructName, cellString string
+	var output []string
 
-	//ch, _ := sam.GoReadToChan("/Users/sethweaver/Downloads/2dayBrainIueOut/lowe/possorted_genome_bam.chrSS.bam") //lowe library analysis
-	//ch, _ := sam.GoReadToChan("/Users/sethweaver/Downloads/2dayBrainIueOut/silver/silver_possorted_genome_bam.chrSS.bam") //silver library analysis
-	ch, _ := sam.GoReadToChan("/Users/sethweaver/Downloads/2dayBrainIueOut/haqer/haqer_possorted_genome_bam.chrSS.bam") //haqer reanalysis
+	ch, _ := sam.GoReadToChan(inSam)
+	out := fileio.EasyCreate(outTable)
 
 	for i := range ch {
 		num, _, _ := sam.QueryTag(i, "xf") //xf: extra flags
 		bit = num.(int32)
 
 		if bit&8 == 8 { // bit 8 is the flag for UMI used in final count.
-			construct, _, _ := sam.QueryTag(i, "GX") // gene associated with read
-			name := construct.(string)
-			umis = append(umis, name)
+			construct, _, _ := sam.QueryTag(i, "GX") // gene associated with UMI
+			constructName = construct.(string)
+			cell, _, _ := sam.QueryTag(i, "CB") // cell associated with UMI
+			cellString = cell.(string)
+			write := fmt.Sprintf("%s\t%s", cellString, constructName)
+			fileio.WriteToFileHandle(out, write)
 			k++
 		}
 	}
-	fmt.Println(k)
-	fileio.Write("/Users/sethweaver/Downloads/2dayBrainIueOut/haqer/cellrangerUsedUMIs.txt", umis)
+	fmt.Println("Found this many valid UMIs: ", k)
+	fileio.Write(outTable, output)
+}
+
+func usage() {
+	fmt.Print("cellrangerBam -- Takes in a cellranger bam file of STARR-seq reads and parses the extra flag field to pull out the" +
+		"representitive read for each UMI and which construct it belongs to. It will also report the cell which it was found in.\n" +
+		"Usage: \n" +
+		"cellrangerBam inFile outFile\n\n")
+	flag.PrintDefaults()
+}
+func main() {
+	var expectedNumArgs int = 2
+
+	flag.Usage = usage
+	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	flag.Parse()
+
+	if len(flag.Args()) != expectedNumArgs {
+		flag.Usage()
+		log.Fatalf("Error: expecting %d arguments, but got %d\n",
+			expectedNumArgs, len(flag.Args()))
+	}
+
+	a := flag.Arg(0)
+	b := flag.Arg(1)
+	parseBam(a, b)
 }
--- a/cmd/cellrangerBam/cellrangerBam.go
+++ b/cmd/cellrangerBam/cellrangerBam.go
@@ -7,28 +7,25 @@ import (
 )
 
 func main() {
-	var j int = 0
 	var k int = 0
 	var bit int32
 	var umis []string
 
-	ch, _ := sam.GoReadToChan("/Users/sethweaver/Downloads/2dayBrainIueOut/lowe/possorted_genome_bam.chrSS.bam")
+	//ch, _ := sam.GoReadToChan("/Users/sethweaver/Downloads/2dayBrainIueOut/lowe/possorted_genome_bam.chrSS.bam") //lowe library analysis
+	//ch, _ := sam.GoReadToChan("/Users/sethweaver/Downloads/2dayBrainIueOut/silver/silver_possorted_genome_bam.chrSS.bam") //silver library analysis
+	ch, _ := sam.GoReadToChan("/Users/sethweaver/Downloads/2dayBrainIueOut/haqer/haqer_possorted_genome_bam.chrSS.bam") //haqer reanalysis
 
 	for i := range ch {
-		if j >= 10 {
-			break
-		}
-		num, _, _ := sam.QueryTag(i, "xf")
+		num, _, _ := sam.QueryTag(i, "xf") //xf: extra flags
 		bit = num.(int32)
 
-		if bit&8 == 8 {
-			construct, _, _ := sam.QueryTag(i, "GX")
+		if bit&8 == 8 { // bit 8 is the flag for UMI used in final count.
+			construct, _, _ := sam.QueryTag(i, "GX") // gene associated with read
 			name := construct.(string)
 			umis = append(umis, name)
 			k++
 		}
-		//j++
 	}
 	fmt.Println(k)
-	fileio.Write("/Users/sethweaver/Downloads/2dayBrainIueOut/lowe/cellrangerUsedUMIs.txt", umis)
+	fileio.Write("/Users/sethweaver/Downloads/2dayBrainIueOut/haqer/cellrangerUsedUMIs.txt", umis)
 }
--- a/cmd/cellrangerBam/cellrangerBam.go
+++ b/cmd/cellrangerBam/cellrangerBam.go
@@ -5,33 +5,94 @@ import (
 	"fmt"
 	"github.com/vertgenlab/gonomics/exception"
 	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/numbers/parse"
 	"github.com/vertgenlab/gonomics/sam"
 	"log"
+	"strings"
 )
 
-func parseBam(inSam string, outTable string) {
+//inputNormalize takes in the psuedobulk map and an input normalization table and normalizes all the raw count values in the map
+func inputNormalize(mp map[string]float64, normalize string) {
+	var total float64
+	var found bool
+	var columns []string
+
+	if normalize != "" {
+		inputNormValues := fileio.Read(normalize)
+		if len(inputNormValues) != len(mp) {
+			fmt.Println("The input normalization table doesn't have the same number of constructs as was found in the input bam.")
+			// trying to find the best way to throw an error if there is a raw count value that doesn't get normalized
+		}
+		for _, i := range inputNormValues {
+			columns = strings.Split(i, "\t")
+			total, found = mp[columns[0]]
+			if found {
+				mp[columns[0]] = total * parse.StringToFloat64(columns[1])
+			} else {
+				fmt.Println("Construct not found in map for normalization: ", columns[0])
+			}
+		}
+
+	}
+}
+
+//writeMap simply writes out the pseudobulk and/or input normalized values to an io.writer
+func writeMap(mp map[string]float64, writer *fileio.EasyWriter) {
+	var total float64
+	var write string
+	for i := range mp {
+		total, _ = mp[i]
+		write = fmt.Sprintf("%s\t%f", i, total)
+		fileio.WriteToFileHandle(writer, write)
+	}
+}
+
+//parseBam takes in a cellranger bam and pulls out reads that are representative of the UMI and also returns the construct associated with the UMI.
+func parseBam(inSam string, outTable string, pb bool, normalize string) {
 	var k int = 0
 	var bit int32
-	var constructName, cellString string
+	var constructName, cellString, write string
+	var count float64
+	var found bool
 
 	ch, _ := sam.GoReadToChan(inSam)
 	out := fileio.EasyCreate(outTable)
+
+	mp := make(map[string]float64)
 
 	for i := range ch {
 		num, _, _ := sam.QueryTag(i, "xf") //xf: extra flags
 		bit = num.(int32)
 
 		if bit&8 == 8 { // bit 8 is the flag for UMI used in final count.
+			k++
 			construct, _, _ := sam.QueryTag(i, "GX") // gene associated with UMI
 			constructName = construct.(string)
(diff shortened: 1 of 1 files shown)


sweavs111:
latest commit on binning cells has a weird bug where the first bin is much more likely to be selected, especially at higher bin numbers. Not working on bug fixes tonight

sweavs111:
image

sweavs111:
open question for those reviewing. There are a lot of print statements while the program is running. Are those helpful? or should I take those out. A lot of them started as sanity checks/debugging but ive left some of them in because I thought they may be useful

RavenYL:
> open question for those reviewing. There are a lot of print statements while the program is running. Are those helpful? or should I take those out. A lot of them started as sanity checks/debugging but ive left some of them in because I thought they may be useful

Usually we do use print statements for debugging and then remove them when they are ready to be merged. But I can imagine situations where it's good to keep fatal prints for failures (as opposed to having prints for successes)

sweavs111:
Screenshot 2023-08-23 at 11 51 19 PM
Screenshot 2023-08-23 at 11 51 45 PM

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+func determineIdealBins(s Settings, cellTypeSlice []string) int {
+	var count, j int
+	var found, stop bool
+	var z, l, m string
+	var binSlice []int
+	var matrix [][]string
+
+	nc := fileio.Read(s.determineBins)
+
+	for i := 0; i <= 10; i++ { //do this whole loop 10 times (essentially 10 replicates)
```
RavenYL, line 343: I think this loop is done 11 times because it gets executed for i==0,1,2,...10

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+					columns := strings.Split(l, "\t") //split an entry in from the matrix
+					count, found = ncMap[columns[1]]  //search the 2nd column to see if it is a negative control
+					if found {                        //if it is, add a count to the map for that particular negative control
+						ncMap[columns[1]] = count + 1
+					}
+				}
+				for m = range ncMap {
+					count, _ = ncMap[m]
+					if count < 1 {
+						binSlice = append(binSlice, j-1)
```
RavenYL, line 363: How is the ideal bin number determined? It looks like ideal bin number = average of binSlice []int, and binSlice records the j (number of bins, some number between 1 and 100) that makes ncMap[m] count <1 (gets 0 counts for negative control constructs) in 10 trials?

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+		}
+		if randNum >= prob*float64(j) && randNum < prob*float64(j+1) { // check to see if our random number belongs to that bin
+			bin = j
+		}
+	}
+	return bin
+}
+
+// distributeCells takes in a slice of string that is created in parseBam that has a list of cellBarcodes and constructs. It will partition those cells and constructs into separate slices. The number of bins is a user-input variable
+func distributeCells(s Settings, cellTypeSlice []string, fromDB bool) [][]string {
```
RavenYL, line 397: What does fromDB mean?

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+	var count float64
+	var outMatrix, columns, columns2 []string
+	var gfpNormFactorMap map[string]float64
+
+	cellTypeMap := make(map[string]string) // [cellBarcode]cellType
+	allCellTypes := make(map[string]int)   // [cellType]placeholderInt
+	//countsPerCellType := make(map[string]int)
+
+	cto := fileio.Read(s.scAnalysis) // read in the tab delimited file with cell and cell type info
+	for _, i := range cto {          //make a map where the cell barcode is the key and the cell type is the value
```
RavenYL, line 454: Note: makes cellTypeMap, allCellTypes

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+			if s.byCell || sc || s.binCells > 0 || s.determineBins != "" {
+				cell, _, _ := sam.QueryTag(i, "CB") // get the cell barcode associated with valid UMI
+				cellString = cell.(string)
+				cellByConstructName = fmt.Sprintf("%s\t%s", cellString, constructName) //list of all construct reads and what cell barcodes they belong to
+				if s.byCell {                                                          //if you only want to print out the construct and the cell it was found in
+					fileio.WriteToFileHandle(out, cellByConstructName)
+				}
+				if sc || s.binCells > 0 || s.determineBins != "" { //get some data for the singleCellAnalysis and distributeCells functions
+					allConstructs = append(allConstructs, constructName) //list of all constructs found in the bam
+					cellTypeSlice = append(cellTypeSlice, cellByConstructName)
```
RavenYL, line 623: I thought cellTypeSlice is a []string of cell types. Is cellByConstructName a string of cell type?

interval/tools.go
```diff
@@ -8,3 +8,17 @@ import (
 func CoordsToString(i Interval) string {
 	return fmt.Sprintf("%s:%v-%v", i.GetChrom(), i.GetChromStart(), i.GetChromEnd())
 }
+
+func AreEqual(a Interval, b Interval) bool {
+	var isEqual bool = false
+
+	if a.GetChrom() == b.GetChrom() {
```
RavenYL, line 15: Style: Would && be more efficient than the nested if checks?

numbers/stat.go
```diff
@@ -48,3 +56,29 @@ func Pearson(a []float64, b []float64) float64 {
 	}
 	return num / (math.Sqrt(sumA) * math.Sqrt(sumB))
 }
+
+/*
```
RavenYL, line 60: Is this code block old and now commented out for some reason, or is it new code that are notes?

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+		if interval.AreEqual(currEntry, i) {
+			if currEntry.Qual >= i.Qual {
+				continue
+			} else {
+				currEntry = i
+				continue
+			}
+		}
+		overlappedBedEntry := interval.Query(selectTree, currEntry, "any")
+		if len(overlappedBedEntry) != 1 {
```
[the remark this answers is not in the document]
sweavs111: This is all post-PCR duplicate removal. If the unique sam entry overlaps either 0 or > 1 bed, we throw it away and go onto the next sam entry. If the unique sam entry overlaps exactly 1 bed entry, then we add one to the count map and go onto the next entry

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+	var prevCluster string
+	var ncAvg []float64
+
+	columns = strings.Split(matrix[0], "\t")
+	currCluster = columns[0]
+
+	clusterIdx := 0
+	for p, i := range matrix {
+		found = false
+		columns = strings.Split(i, "\t")
```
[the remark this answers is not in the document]
sweavs111: you might be right that a custom struct would be helpful

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+// cell type
+func getNcAvg(matrix []string, ncSlice []string, ncVals [][]float64) []float64 {
+	var found bool
+	var currCluster string
+	var columns []string
+	var prevCluster string
+	var ncAvg []float64
+
+	columns = strings.Split(matrix[0], "\t")
+	currCluster = columns[0]
```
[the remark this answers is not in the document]
sweavs111: Yeah good question i don't exactly know. My reasoning was: I wanted to set the first entry to "currCluster" because i have logic in the for loop that compares currCluster to prevCluster. If they are both nil then on the first time through the loop the will be equal (?) and move onto the next cluster when i don't want it to

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+func normScToNegativeCtrls(matrix []string, s Settings, numCellTypes int) []string {
+	var ncSlice, columns []string
+	var prevCluster, newLine string
+
+	ncVals := make([][]float64, numCellTypes)
+	nc := fileio.Read(s.ncNorm)
+	for _, i := range nc {
+		ncSlice = append(ncSlice, i)
+	}
+	ncAvg := getNcAvg(matrix, ncSlice, ncVals)
```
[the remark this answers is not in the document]
sweavs111: yes good idea i will do that

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+func umiSaturation(umiBxSlice []string, out *fileio.EasyWriter) {
+	var perc, randNum float64
+	var j string
+	var count int
+
+	fileio.WriteToFileHandle(out, "totalReads\tumis")
+
+	for i := 1; i <= 10; i++ {
+		perc = float64(i) / 10
+		mp := make(map[string]bool)
```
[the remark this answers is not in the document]
sweavs111: i want to erase the map and create a new one for every time in the loop. I thought this would be the best way to do that. Is that true / not true?

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+		count = 0
+		for _, j = range umiBxSlice {
```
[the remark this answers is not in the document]
sweavs111: broadly, this for loop is counting unique UMIs at different proportions of the sam file. You are correct that if the random number is larger than the proportion we continue "continue" which stops the current iteration of the for loop and goes onto the next one (aka: the next read in the sam file). If the read passes the initial test then count is the total reads and the length of the map will be the number of UMIs (if the same UMI is given to the map it will just overwrite)

numbers/stat.go
sweavs111: this was something that I was thinking about writing to add stats to the program. I think i am going into a different direction, so i will delete

interval/tools.go
sweavs111: good point

cmd/cellrangerBam/cellrangerBam.go
sweavs111: haha good catch

cmd/cellrangerBam/cellrangerBam.go
sweavs111: yes that is correct

cmd/cellrangerBam/cellrangerBam.go
sweavs111: that is "from determine bins". Normally the distributeCells function prints out how many cells make it into each bin, but i didn't want that to be printed out each time that determineBins runs the loop. So if fromDB is true, it doesn't print out the summary statistics for bin allocation

cmd/cellrangerBam/cellrangerBam.go
```diff
@@ -0,0 +1,823 @@
...
+		var a int = 0
+		singleCellTypeMap := make(map[string]float64) // make a map which will hold the construct counts for the current cell type ([construct]counts)
+		for _, j = range allConstructs {              // loop over all construct that were found in the bam file to make every construct start at 0 reads
+			singleCellTypeMap[j] = 0
+		}
+		for _, i := range cellTypeSlice { //loop over the slice created in cellrangerBam() that that contains: cellBarcode \t construct
+			columns2 = strings.Split(i, "\t")
+			cluster, found = cellTypeMap[columns2[0]] //search for that cell barcode in the cellType map
+			if found && cluster == cellType {         // if that cell is found and the cell type in the map matches the cell type we are looping through, search for the construct corresponding with that cellBarcode in the singleCellCount map and add one to the value
+				count, _ = singleCellTypeMap[columns2[1]]
```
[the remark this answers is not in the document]
sweavs111: yes i think you are right i recently learned this

cmd/cellrangerBam/cellrangerBam.go
sweavs111: cellByConstructName is maybe poorly named that is the slice that has cellBx \t construct

cmd/cellrangerBam/cellrangerBam.go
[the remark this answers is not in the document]
sweavs111: looks like the correct way to do this is: singleCellTypeMap[columns2[1]] += 1

sweavs111:
note to current reviewers: I am preparing a fairly large PR that introduces multiple custom structs and the starr-seq package

starrSeq/binCells.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		sort.Strings(outSlice)
+		for _, i := range outSlice {
+			fmt.Println(i)
+		}
+	}
+	return binnedCells
+}
+
+// DetermineIdealBins will determine how many bins should be used for the pseudobulk -binCells option.
```
hcnapier, line 72: This is not a super well thought out idea, but since you're trying to get the ideal bin number mainly to make sure you have some reads from each negative control construct in each bin, could it make sense to just make this the bin number equal to the minimum number of cells for any of the constructs? For example, if you have NCa with 5 cells, NCb with 7 cells, and NCc with 8 cells, choose 5 bins to ensure you have at least one cell from each NC construct in each bin. Perhaps the main downside here is that you'd need to write a separate bin distribution function for the NC cells to make sure that each NC cell for a given construct will go to a new bin before you start repeating bins. I feel like that might end up being more efficient then simulating different bin sizes many times though.

starrSeq/singleCellStarrSeq.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		for m = range singleCellTypeMap { // write out the results from that cell type
+			outLine = scStarrSeqMatrix{counts: singleCellTypeMap[m], cluster: cellType, construct: m}
+			outMatrix = append(outMatrix, outLine)
+		}
+		fmt.Println(fmt.Sprintf("Found %d raw counts in the cell type: %s", a, cellType))
+	}
+
+	sortScCountMatrixByConstruct(outMatrix)
+	sortScCountMatrixByCluster(outMatrix) //sort the slice before writing
```
RavenYL, line 75: Note to self: the 2 sorts probably don't cancel each other out, but means the output will be sorted first by cluster, then by construct

starrSeq/normalize.go
```diff
@@ -0,0 +1,190 @@
...
+	var ncAvg []float64
+
+	clusterIdx := 0
+	for p, i := range matrix {
+		found = false
+		currCluster = i.cluster
+		if p > 0 && currCluster != prevCluster {
+			clusterIdx++
+		}
+		for _, j := range ncSlice {
```
RavenYL, line 59: Note to self: for each element of matrix, check if its "construct" is matches any one of the "ncSlice constructs"

starrSeq/alt.go
```diff
@@ -0,0 +1,121 @@
...
+	var outSlice []Read
+
+	umiMap := make(map[string]Read) //make a map that has cellBarcode/UMI as the key and the corresponding Read entry is the value
+	for _, i := range ssrs {        // loop through slice of Read
+		bxUmi = fmt.Sprintf("%s_%s", i.Bx, i.UMI) //append cell barcode and UMI together
+		read, found = umiMap[bxUmi]               //check to see if the bx/umi is already in the map
+		if !found {
+			umiMap[bxUmi] = i //if it's not, add it with the Read to the map
+		} else {
+			if read.Construct == i.Construct { // if its found, check to make sure the current bx/umi maps to the same construct as the one in the map
```
RavenYL, line 108: I thought you had a function where "if multiple reads have the same bxUmi, then keep the longer read". Was that ever a function?

starrSeq/alt.go
```diff
@@ -0,0 +1,121 @@
...
+				starrSeqRead = Read{Bx: "NA", UMI: umi.(string), Cluster: "undefined", Construct: overlap[0].(bed.Bed).Name}
+			case cbcFound && found:
+				starrSeqRead = Read{Bx: cbcUpdate, UMI: umi.(string), Cluster: cluster, Construct: overlap[0].(bed.Bed).Name}
+			case cbcFound && !found:
+				starrSeqRead = Read{Bx: cbcUpdate, UMI: umi.(string), Cluster: "undefined", Construct: overlap[0].(bed.Bed).Name}
+			}
+			starrSeqReadSlice = append(starrSeqReadSlice, starrSeqRead)
+		}
+	}
+	collapsedUmiSlice := collapseUmi(starrSeqReadSlice)
```
RavenYL, line 83: Alt does include collapseUmi, just with gonomics, not cellranger. Would it be helpful to users to change the cmd/starrSeqAnalysis.go usage to something like "collapseUmi by only removing identical cell barcode - UMI combinations through gonomics", rather than "by hand"?

starrSeq/alt.go
RavenYL: From Seth: part of samFilter not cellrangerBam

PR #906, opened 2023-08-02
Straw to bedpe

cfauci:
# Description
strawToBedpe takes the output of the aidenlabs straw command and returns a bedpe. This is how I've found is the best way to transfer hic data into bedpe format

## Relevant Links

- [aidenlabs/straw git](https://github.com/aidenlab/straw)

here is the straw usage:
straw <observed/oe/expected> <normalization> <hic file> <chrom bin1> <chrom bin2> <Bp/Frag> <binSize>
for our uses the normalization will be set to NONE and the BP/Frag will be set to BP. This output will give us the raw values for everything.

this is what the straw output looks like:
startBin1 startBin2 score (number of contacts)
3000000	3000000	4
3000000	3005000	28
3005000	3005000	9
3000000	3010000	15
3005000	3010000	24
3010000	3010000	8
3000000	3015000	1
3005000	3015000	5
3010000	3015000	7
3000000	3020000	4

straw requires an argument that specifies which chromosomes you want in the comparison and can do interchromosomal contacts, which is why strawToBedpe has the option to provide a second chrom string. You can also specify the bin size (resolution) that straw prints out at and so that is also an option with a default of 5000. 

Future additions:
*Filtering for Score
*Filtering out any contacts that are within the same bin, or within bins of a certain distance (option to specify)


rimangan (approved):

cmd/strawToBedpe/strawToBedpe.go
```diff
@@ -0,0 +1,70 @@
...
+			"\n" +
+			"Usage:\n" +
+			"\n" +
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var expectedNumArgs int = 4
+	var binSize *int = flag.Int("binSize", 5000, "A binSize must be provided for the resolution of the straw file output if it is not 5000.")
```
sweavs111, line 52: what is the behavior if the input bin size is different than the straw bin size? Is there any way to autodetect bin size from the straw file and keep it in the straw struct?

cmd/strawToBedpe/strawToBedpe.go
```diff
@@ -0,0 +1,70 @@
+// Command Group: "Data Conversion"
```
sweavs111, line 1: This is a random question not really pertaining to the PR directly, but have straw files already gone through sequencing filters and expected/observed interaction normalization for distance between bins? I know there can be quite significant corrections for hic sequencing data (though I don't know exactly what they entail) and it would be good to know if these have already been applied at the 'straw' stage as it will effect interaction calling I'm sure

sweavs111 (commented):
Really like this command, just had 1.5 minor comments. The one about bin size is probably more pertinant

cmd/strawToBedpe/strawToBedpe.go
sweavs111: maybe a better way of phrasing this question is: does the input binSize from the option have to equal the bin size from the straw file?

cfauci: yes it does if you want the data to be accurate to the actual contact information

cmd/strawToBedpe/strawToBedpe.go
cfauci: From my understanding, you can get any of this information from the straw file but you can also set everything to get the raw data which I believe I've done and I have an example usage listed in the description here.

cmd/strawToBedpe/strawToBedpe.go
sweavs111: Should the program throw an error if the actual vs input bin sizes aren't equal?

cfauci: The straw file doesn't contain any information about window size so there's no straight forward way to check for it.

sweavs111 (approved):

PR #907, opened 2023-08-04
3dGreat Part III: Gaf Parser

rimangan:
# Description
First part of the ontology package: a parser for the GAF (Go Annotation File) format. Full documentation is here: http://geneontology.org/docs/go-annotation-file-gaf-format-2.2/
We only need a few fields here. Field 3 is the gene name and field 5 is the Gene Ontology ID. We'll need these to relate genes to traits for trait enrichment analysis. More to come, testing here ensures I/O fidelity.
Test file is the first 200 lines of the Gene Ontology Catalog for humans.

## Relevant Links

### Testing

None

### Screenshots & Media

None

None

### DIFF
--- a/bed/bed.go
+++ b/bed/bed.go
@@ -34,7 +34,7 @@ const (
 	None     Strand = '.'
 )
 
-// String converts a bed struct to a string so it will be automatically formatted when printing with the fmt package.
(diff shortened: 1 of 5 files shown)


craiglowe (approved):
Great!

PR #908, opened 2023-08-04
normal distribution from bedpe data 

cfauci:
# Description
this is still a work in progress, but as of now functions properly (besides the cmd/bedMath/bedMath.go). this is designed to calculate a normal distribution from a bedpe for the curve that describes the number of contacts as a function of distance from any bin in the genome.

## Relevant Links

### Testing

None

### Screenshots & Media

None

Need to add capability of - strand to functionality

### DIFF
--- a/numbers/curveFit/curveFit.go
+++ b/numbers/curveFit/curveFit.go
@@ -1,0 +1,8 @@
+package curveFit
+
+import "github.com/vertgenlab/gonomics/numbers"
+
+func gauss() {
+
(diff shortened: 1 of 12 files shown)


numbers/curveFit/curveFit.go
```diff
@@ -0,0 +1,8 @@
+package curveFit
```
cfauci, line 1: this file will probably be deleted but I'm still working on how to do the sampling and fitting that I need to do

PR #909, opened 2023-08-05
3dGreat Part IV: Obo Package

rimangan:
# Description
New package, obo, is a subpackage of ontology. This package handles Open Biomedical Ontologies (OBO) format files. Specs for this format are here: http://owlcollab.github.io/oboformat/doc/obo-syntax.html 
Really only the first six fields of the struct are useful. This format encodes gene ontologies, definitions and human readable names, and how ontologies relate to each other. Ontologies are organized into a tree, with parent and child nodes. For example, a term like "Neurodegenerative Disease" may have a parent term "Nervous System Disease" and a child term "Alzheimer's Disease". 
This package can also arrange Obo structs into a tree, based on annotated relationships between terms. These trees can be exported into DOT format, which is used for plotting. 
As the entire GO catalog has tens of thousands of terms, we will want to pick specific subtrees for plotting. This package can identify the size of subtrees rooted on any node, allowing us to pick well sized trees for plotting. Below I show a subtree for the term "Reproduction", which has only a handful of nodes.
image

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


cfauci (dismissed):

ontology/obo/testdata/mannosyltransferaseActivity.dot
```diff
@@ -0,0 +1,12 @@
+digraph G{
```
RavenYL, line 1: I think the 3 test out files could be removed?
Otherwise, waiting for new commit before merging

RavenYL (approved):

ontology/obo/tree.go
```diff
@@ -0,0 +1,160 @@
...
+// calculates the size of subtrees rooted on a particular Obo struct in a larger
+// Obo tree.
+func numberOfDescendentsRecursive(term *Obo, visitedNode map[string]bool) {
+	if _, foundInMap := visitedNode[term.Id]; foundInMap {
+		return
+	}
+	visitedNode[term.Id] = true
+	//leaf case
+	if len(term.Children) == 0 {
+		term.SubTreeSize = 0
```
craiglowe, line 138: I think we want to adjust counting or vocabulary so that I leaf is a subtree of size one, or has descendents of zero.  Right now we have a subtree size of 0.  I think the concept of a subtree is more common in math/cs, so I would vote for changing the counting.

ontology/obo/tree.go
```diff
@@ -0,0 +1,160 @@
...
+	// Now we iterate through terms and set up the tree
+	for i := range terms {
+		term := &terms[i] //get a memory copy
+		for _, isAItem = range term.IsA {
+			parentId = isAItem.ParentId
+			if parentTerm, foundInMap = termMap[parentId]; foundInMap {
+				term.Parents = append(term.Parents, parentTerm)
+				parentTerm.Children = append(parentTerm.Children, term)
+			} else {
+				//log.Fatalf("Error: The term with ID \"%s\" has a parent with ID \"%s\" that "+
```
craiglowe, line 29: I think this should be a fatal.  If they say a node exists, that does not exist, that seems like an error

ontology/obo/tree.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"fmt"
+	"github.com/vertgenlab/gonomics/exception"
+	"github.com/vertgenlab/gonomics/fileio"
+	"log"
+)
+
+// BuildTree takes a slice of Obo structs and builds a tree, setting
+// the Parents and Children fields to point to related nodes.
+func BuildTree(terms []Obo) map[string]*Obo {
```
craiglowe, line 12: I think there is a tension here where it is unclear if we are treating the obo as a tree, slice, or map.  We can get many slice benefits from a map (e.g. looping through keys), so maybe the question is whether it is a tree or a map.  From the code I am seeing, I think we want the reader to get a root (or set of roots) at the end and treat the ontology as a tree.

ontology/obo/tree.go
```diff
@@ -0,0 +1,160 @@
...
+				//	"is not found in the Obo file.\n", term.Id, parentId)
+			}
+		}
+	}
+	return termMap
+}
+
+// termToDot is a helper function of ToDot that converts and individual Obo struct to a
+// DOT format digraph descriptor, and recursively does so for all children of the input Obo struct.
+func termToDot(term Obo, out *fileio.EasyWriter, visitedNode map[string]bool) {
```
craiglowe, line 39: We have some complexity in the code (e.g. visitedNode map) because we don't know the root.  I think this should probably take a pointer to the root instead of a term and visitedNode

ontology/obo/obo.go
```diff
@@ -0,0 +1,276 @@
...
+// new ontology element.
+func peekIsTermLine(peek []byte) bool {
+	if len(peek) != 6 {
+		log.Fatalf("Error: peekIsTermLine expected 6 bytes. Received %v.\n", len(peek))
+	}
+	peekString := string(peek)
+	return peekString == "[Term]"
+}
+
+// Read creates a slice of Obo structs and a Header struct from an input filename.
```
craiglowe, line 92: Do we want a list without the tree.  Maybe the read function should build the tree and return the tree?

ontology/obo/tree.go
```diff
@@ -0,0 +1,160 @@
...
+	termMap := makeTermMap(terms)
+
+	var parentId string
+	var foundInMap bool
+	var parentTerm *Obo
+	var isAItem IsADescription
+
+	// Now we iterate through terms and set up the tree
+	for i := range terms {
+		term := &terms[i] //get a memory copy
```
craiglowe, line 22: Why do we make a memory copy?  Is it because we may be editing the slice, but want the tree to be separate data that is not linked back to the slice?  If obo files are big, maybe we read it into a slice of pointers to structs, then link up the structs, with no copies.  We then return the tree to the person who asked us to read the file.
I think due to history, we first thought about treating ontologies as a list, rather than a tree.  Now that we have decided to treat them like a tree, it probably makes sense to use a tree data structure.  Right now we have some tension where we are wanting to treat it like a tree, but are holding onto the list structure.

ontology/obo/tree.go
```diff
@@ -0,0 +1,160 @@
...
+	err = out.Close()
+	exception.PanicOnErr(err)
+}
+
+// makeTermMap creates a map to find nodes based on their ID and AltIDs
+func makeTermMap(terms []Obo) map[string]*Obo {
+	var altId string
+	termMap := make(map[string]*Obo)
+	for i := range terms {
+		term := &terms[i] // get a memory copy
```
craiglowe, line 82: I think this function should take a pointer to the root and then use recursion to make a map of pointers to the nodes.  I don't think we want/need the copying

ontology/obo/tree.go
```diff
@@ -0,0 +1,160 @@
...
+// SubtreeToDot makes a DOT format tree for the subtree rooted at the input nodeId.
+// BuildTree must have been run on this tree already, as termMap is an argument.
+func SubtreeToDot(outFile string, nodeId string, termMap map[string]*Obo) {
+	var err error
+	out := fileio.EasyCreate(outFile)
+	//write header line for Dot
+	_, err = fmt.Fprintf(out, "digraph G {\n")
+	exception.PanicOnErr(err)
+
+	var visitedNode = make(map[string]bool)
```
craiglowe, line 107: This is another place where we have some added complexity because of storing a tree in a list data structure.

craiglowe (commented):
Great work!  My overall comment is that we have some complexity where we are trying to store the obo in a list data structure, but treat it like a tree.  I realize this is due to the history of first thinking of ontologies like lists instead of like trees, but I think it is better to fix this now, or it may continue to cause complex code in the future.

ontology/obo/tree.go
rimangan: next commit resolves

ontology/obo/tree.go
rimangan: Next commit adds the "force" option to BuildTree, allowing us to control this behavior

craiglowe: Is the "force" option needed for the Gene Ontology files?  It is a good option to have if there are bugs in the GO files, but I am worried we are giving researchers wiggle room that they do not need, which may lead to silent errors.

fileio/fileio.go
```diff
@@ -174,16 +175,41 @@ func AreEqual(a string, b string) bool {
 	return equal(a, b, true)
 }
 
+// AreEqualIgnoreOrder returns true if input files contain the same lines,
+// although the order of the lines does not matter.
+// This program sorts the two files and compares the contents, so it is not well
+// suited for large files as the whole contents are read into memory.
+func AreEqualIgnoreOrder(a string, b string) bool {
```
craiglowe, line 182: This is a cool function!

craiglowe (approved):
Great!

edotau (approved):
LGTM

PR #913, opened 2023-08-09
feat: yamlfmt-golint-eau

edotau:
# Description

- Small PR to remove a few annoying golintci settings like import order, etc.
- A few clean ups removing whitespace at end of functions that dont add a whole lot of value
- I've reduce the lint rules to these 3, please feel free to adjust them as it fits the labs goals/standards in future PRs
    - https://github.com/vertgenlab/gonomics/blob/ee5acb189c76b676d41b59dc7467ba66ad23e77f/.golangci.yml#L17-L20
- Added yamlfmt which is used the same way on yaml files

## Relevant Links

- google's yamlfmt which is the same functionality as go fmt but on yaml files contained inside the repo
- https://github.com/google/yamlfmt
- lintci configuration documentation can be found here:
    - https://golangci-lint.run/usage/linters/

### Testing

None

### Screenshots & Media

None

None

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -58,6 +58,6 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: false
+          fail_ci_if_error: true
           files: coverage.txt
           verbose: true
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,14 +1,11 @@
 name: Go
-
 on:
   push:
     branches: main
   pull_request:
     branches: main
-
 env:
   GOLANG: 1.20.3
-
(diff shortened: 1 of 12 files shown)


.github/workflows/go.yml
```diff
@@ -61,6 +58,6 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: false
+          fail_ci_if_error: true
```
edotau, line 61: Should this be true or false? @dasnellings

dasnellings: I believe we decided this should be false. We want to be notified if coverage is going down, but we don't want to outright fail the PR.

cfauci (approved):

RavenYL (approved):

PR #922, opened 2023-09-01
faFindFast from 2-sequence to 3-sequence

RavenYL:
# Description
- faFindFast can now accept a multiFa file with more than 2 sequences,
- faFindFast will make the 1st sequence in the multiFa file the reference sequence (output bed file positions will be based on this sequence)
- By default, faFindFast will make the 1st and 2nd sequences in the multiFa file the firstQuery and secondQuery sequences (output bed file will report divergence between these 2 sequences)
- Options allow the user to define firstQuery and secondQuery sequences by name

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


cmd/faFindFast/efficient.go
```diff
@@ -4,133 +4,210 @@ import (
...
-		// usually increment the baseBeforeWindow, but not at the beginning when we have not yet incremented the end
-		// enough to have a full "windowSize" of bases in the window
-		if lastRefIdxOfWindow-refIdxBeforeWindow > windowSize {
-			alnIdxBeforeWindow, _, _, _, numRefNs, _, numQueryNsMatch, numSubst = incrementWindowEdge(reference, query, alnIdxBeforeWindow)
-			refIdxBeforeWindow++
+		// usually increment the baseBeforeWindow,
+		// but not at the beginning when we have not yet incremented the end enough to have a full "windowSize" of bases in the window
+		if lastFirstQueryIdxOfWindow-firstQueryIdxBeforeWindow > s.WindowSize {
+			alnIdxBeforeWindow, _, _, _, numFirstQueryNs, _, numSecondQueryNsMatch, numSubst = incrementWindowEdge(firstQuery, secondQuery, alnIdxBeforeWindow)
+			alnIdxBeforeWindowForRef = updateAlnIdxBeforeWindow(firstQuery, alnIdxBeforeWindow)
```
RavenYL, line 144: first main new function in this PR: updateAlnIdxBeforeWindow

cmd/faFindFast/efficient.go
```diff
@@ -4,133 +4,210 @@ import (
...
-					percentDiverged = 100 * (float64(totalSubst+totalGaps) / float64(windowSize))
-					if totalSubst+totalGaps > windowSize {
-						log.Fatalf("Error: total number of mutations exceeds windowSize. This may or may not be a bug, but your sequence has deviated from our use case.")
-					}
-					rawPValue = scorePValueCache[totalSubst+totalGaps]
-					_, err = fmt.Fprintf(out, "%s\t%d\t%d\t%s_%d\t%d\t%s\t%e\t%e\n", refChrName, refIdxBeforeWindow+1, lastRefIdxOfWindow+1, refChrName, refIdxBeforeWindow+1, totalSubst+totalGaps, "+", percentDiverged, rawPValue)
-					exception.PanicOnErr(err)
+		if lastFirstQueryIdxOfWindow-firstQueryIdxBeforeWindow == s.WindowSize && lastAlnIdxOfWindow < len(firstQuery) {
+
+			// convert position from firstQuery to reference
```
RavenYL, line 161: second main new function in this PR: convert position from firstQuery to reference

cmd/faFindFast/efficient.go
```diff
@@ -4,133 +4,210 @@ import (
...
+				// call fasta.AlnPosToRefPosCounterSeq on the alnPos of chromStart and chromEnd
+				// as opposed to alnIdxBeforeWindowForRef and lastAlnIdxOfWindow
+				// this is because fasta.AlnPosToRefPosCounterSeq rounds position up when it encounters gap
+				// in order to only scan the genome once, call fasta.AlnPosToRefPosCounterSeq on saved refStart and alnStart once refStart and alnStart both exceed 0
+				if lastRefIdxOfWindowPlusOne < 0 || lastAlnIdxOfWindow+1 < 0 {
+					refIdxWindowStart = fasta.AlnPosToRefPosCounterSeq(reference, alnIdxBeforeWindowForRef+1, 0, 0)
 				} else {
-					_, err = fmt.Fprintf(out, "%s\t%d\t%d\t%s_%d\t%d\n", refChrName, refIdxBeforeWindow+1, lastRefIdxOfWindow+1, refChrName, refIdxBeforeWindow+1, totalSubst+totalGaps)
-					exception.PanicOnErr(err)
+					refIdxWindowStart = fasta.AlnPosToRefPosCounterSeq(reference, alnIdxBeforeWindowForRef+1, lastRefIdxOfWindowPlusOne, lastAlnIdxOfWindow+1)
```
RavenYL, line 172: hopefully I've achieved memory-caching here

cmd/faFindFast/efficient.go
```diff
@@ -4,133 +4,210 @@ import (
...
+
+				// an option/flag can tell us not to print if there are Ns in the firstQuery or secondQuery
+				if !s.RemoveN || totalNs == 0 {
+					if s.LongOutput && !s.OutputAlnPos {
+						percentDiverged = 100 * (float64(totalSubst+totalGaps) / float64(s.WindowSize))
+						if totalSubst+totalGaps > s.WindowSize {
+							log.Fatalf("Error: total number of mutations exceeds windowSize. This may or may not be a bug, but your sequence has deviated from our use case.\n")
+						}
+						rawPValue = scorePValueCache[totalSubst+totalGaps]
+						_, err = fmt.Fprintf(file, "%s\t%d\t%d\t%s_%d\t%d\t%s\t%e\t%e\n", s.RefChromName, refIdxWindowStart, lastRefIdxOfWindowPlusOne, s.RefChromName, refIdxWindowStart, totalSubst+totalGaps, "+", percentDiverged, rawPValue)
```
RavenYL, line 184: in old code would print (start, end) as (refIdxBeforeWindow+1, lastRefIdxOfWindow+1). note that ref in old code is firstQuery in new code. Now, directly use fasta.AlnPosToRefPosCounterSeq to calculate the bed entry's (start, end) as (refIdxWindowStart, lastRefIdxOfWindowPlusOne). This is because if we convert from firstQueryBeforeWindow to refIdxWindowStart-1, we can't guarantee that firstQueryBeforeWindow+1 converts to refIdxWindowStart. Smilarly, if we convert from lastRefIdxOfWindow to lastRefIdxOfWindow, we can't guarantee that lastRefIdxOfWindow+1 converts to lastRefIdxOfWindowPlusOne

cmd/faFindFast/efficient.go
RavenYL: I wonder if this is the major slowdown. For example, there may be big gaps in real sequences and it will take a long time to find the next nonGap base?

cmd/faFindFast/efficient.go
```diff
@@ -4,133 +4,210 @@ import (
...
+		// keep track of changes between the original and next non-gap positions in the firstQuery sequence
+		// 6. secondQuery
+		if secondQuery[alnIdx] == dna.N {
+			numSecondQueryNsGap++
 		}
-		if seqTwo[alnIdx] != dna.Gap {
-			gapOpenCloseRef = 1
+		// 2. gapOpenClosedFirstQuery
+		if secondQuery[alnIdx] != dna.Gap {
+			gapOpenCloseFirstQuery = 1 // Rationale: if the for loop is executed, this means there is a gap in the firstQuery, and if secondQuery does not have a gap, then it is indeed a firstQuery gap compared to secondQuery
```
RavenYL, line 47: This file has the biggest changes. Most of the changes in this file are renaming anything to do with "reference" to "firstQuery", which does not impact code efficiency.

cmd/faFindFast/efficient.go
```diff
@@ -4,133 +4,210 @@ import (
...
+	// divergenceRate = math.MaxFloat64 is a reserved value that signifies that the user has not set a divergence rate. If divergenceRate != -1,
 	// we initialize the scorePValueCache.
-	if divergenceRate != -1 {
-		scorePValueCache = binomialDistCacheLog10(windowSize, divergenceRate)
+	if s.DivergenceRate != math.MaxFloat64 {
+		scorePValueCache = binomialDistCacheLog10(s.WindowSize, s.DivergenceRate)
 	}
 
-	for lastAlnIdxOfWindow < len(reference) { // this check could also be "!done", I am not sure what is more clear
+	for lastAlnIdxOfWindow < len(firstQuery) { // this check could also be "!done", I am not sure what is more clear
```
hcnapier, line 131: I like the phrasing you have written right now, I think it makes the condition more explicit

cmd/faFindFast/faFindFast.go
```diff
@@ -6,29 +6,69 @@ package main
...
+	records := fasta.Read(s.InFile)    // fasta.Read already makes sure that sequence names are unique
+	recordsMap := fasta.ToMap(records) // generate map[string][]dna.Base. No order but can do key-value search
+
+	if len(records) < 2 {
+		log.Fatalf("Error: There must be at least 2 fasta records in the input file.\n")
+	}
+
+	//if reference and query names were not specified in the command, then use 1st and 2nd fasta records' names, respectively
+	//allows backward compatibility with previously-written code
+	if s.FirstQueryName == "" {
```
edotau, line 44: #### For setting the firstQuery and secondQuery, I'd get rid of the redundancy code. We can separate this logic into a helper function like this:

```golang
func getSequenceFromName(recordsMap map[string][]dna.Base, name string, defaultSeq []dna.Base) []dna.Base {
	if name == "" {
		return defaultSeq
	}
	seq, exists := recordsMap[name]
	if !exists {
		log.Fatalf("Error: %s name is not found in the input file.\n", name)
	}
	return seq
}
```
#### Once you have helper function you can just call `getSequenceName()` and set `firstQuery` and `secondQuery`:

```suggestion
	firstQuery = getSequenceFromName(recordsMap, s.FirstQueryName, records[0].Seq)
	secondQuery = getSequenceFromName(recordsMap, s.SecondQueryName, records[1].Seq)
```

cmd/faFindFast/faFindFast.go
```diff
@@ -6,29 +6,69 @@ package main
...
+	}
+	if s.SecondQueryName == "" {
+		secondQuery = records[1].Seq
+	} else {
+		_, found = recordsMap[s.SecondQueryName]
+		if found {
+			secondQuery = recordsMap[s.SecondQueryName]
+		} else {
+			log.Fatalf("Error: second query name is not found in the input file.\n")
+		}
```
edotau, line 62: The `faFindFast()` function would be reduced to something like this:

```golang
func faFindFast(s Settings) {
	var reference, firstQuery, secondQuery []dna.Base

	records := fasta.Read(s.InFile)    // fasta.Read already makes sure that sequence names are unique
	recordsMap := fasta.ToMap(records) // generate map[string][]dna.Base. No order but can do key-value search

	if len(records) < 2 {
		log.Fatalf("Error: There must be at least 2 fasta records in the input file.\n")
	}

	firstQuery = getSequenceFromName(recordsMap, s.FirstQueryName, records[0].Seq)
	secondQuery = getSequenceFromName(recordsMap, s.SecondQueryName, records[1].Seq)
	reference = records[0].Seq // reference will always be the 1st fasta record in a multiFa input file

	if !(len(reference) == len(firstQuery) && len(reference) == len(secondQuery)) {
		log.Fatalf("Error: Reference, first query, and second query sequences are not all of equal length.\n")
	}

	speedyWindowDifference(reference, firstQuery, secondQuery, s)
}

```

cmd/faFindFast/faFindFast.go
```diff
@@ -6,29 +6,69 @@ package main
...
+	RefChromName    string
+	RemoveN         bool
+	LongOutput      bool
+	DivergenceRate  float64
+	OutputAlnPos    bool
+}
+
+func faFindFast(s Settings) {
+	var reference, firstQuery, secondQuery []dna.Base
+	var found bool
```
edotau, line 33: - You can also check against the map if the key exists key, which will return a bool without using the found flag.

edotau (dismissed):
Great job! I I only had one comment about the redundancy of code getting the name from the records so i suggested that become a helper function to improve readability and simplify the main function a bit.

rimangan (dismissed):
A monumental achievement! Great to see this update!

edotau (dismissed):

edotau:
You might have to run go fmt on the files and check them in once again, at least that's the error I am seeing

edotau (approved):

rimangan (approved):

PR #923, opened 2023-09-05
tfMatchComp

rimangan:
# Description

Two new commands: tfMatch and tfMatchComp.
tfMatch scans a fasta for motif occurrences.
tfMatchComp scans an alignment of fasta sequences and quantifies where motifs have been gained or lost between sequences.
Here are some slides describing the algorithm:
https://docs.google.com/presentation/d/1Y9odAFzi8bL244BqeITUUHFX3d87duuEZIElRUAUA_Q/edit?usp=sharing 

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


motif/match.go
```diff
@@ -0,0 +1,304 @@
...
+	var answer = make([][]rankTensorElement, 4)
+	for row = 0; row < 4; row++ {
+		answer[row] = make([]rankTensorElement, len(p.Mat[row]))
+		for column = 0; column < len(p.Mat[row]); column++ {
+			answer[row][column] = rankTensorElement{Value: p.Mat[row][column], Base: dna.Base(row)} //cast row to dna.Base (0 -> A, 1 -> C, 2 -> G, 3 -> T)
+		}
+	}
+	//sort matrix columns by rank
+	for column = 0; column < len(p.Mat[0]); column++ {
+		for currRank = 0; currRank < 3; currRank++ {
```
RavenYL, line 220: Style: This is a small consideration but what is the most efficient way to sort the matrix? I was thinking maybe only 1 for loop would be necessary?

motif/match.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+		answer = answer + "]\n"
+	}
+	answer = answer + "\n"
+
+	return answer
+}
+
+// buildKmerHash produces a hash mapping 2bit encoded kmer sequences to their corresponding motif score for a PositionMatrix.
+// Only kmers with a motif score above an input thresholdProportion are stored in the
```
RavenYL, line 254: are stored in the output map?

motif/match.go
```diff
@@ -0,0 +1,304 @@
...
+	//first score sequence
+	//we look at parent score, and we look at changed base.
+	currValue := parentValue + rankMatrix[rankVector[index]][index].Value - rankMatrix[rankVector[index]-1][index].Value
+	// if we pass threshold. Everything else is in this if, so if we fail threshold, we just return
+	if currValue >= threshold {
+		//if we've passed, we add current sequence to the map
+		var currKey uint64 = dnaTwoBit.BasesToUint64RightAln(currSeq, 0, len(currSeq))
+		answer[currKey] = currValue
+		for i := index; i < len(rankMatrix[0]); i++ { //generate children to the right of the current index.
+			if rankVector[i] < 3 { //we only have to generate children if the rank vector is less than 3 at this position (only 4 base identities)
```
RavenYL, line 294: Would it be a small efficiency boost if the if and for lines are swapped so the for loop only gets executed if the "if" is met?

motif/matchComp.go
```diff
@@ -0,0 +1,288 @@
...
+		altEndsConsidered := make(map[int]bool, 0)
+
+		kmerHash = buildKmerHash(motifs[i], s.PropMatch)
+		scanRefSequenceComp(s.Records, kmerHash, motifs[i], s.ChromName, out, s.ResidualWindowSize, consensusScore, bed.Positive, s.RefStart, s.EnforceStrandMatch, s.OutputAsProportion, altEndsConsidered, s.ResidualFilter)
+
+		revCompMotif = ReverseComplement(motifs[i])
+		revKmerHash = buildKmerHash(revCompMotif, s.PropMatch)
+		scanRefSequenceComp(s.Records, revKmerHash, revCompMotif, s.ChromName, out, s.ResidualWindowSize, consensusScore, bed.Negative, s.RefStart, s.EnforceStrandMatch, s.OutputAsProportion, altEndsConsidered, s.ResidualFilter)
+
+		//now we scan the alt sequences for any motifs lost in ref
```
RavenYL, line 75: is "alt" the same as "query", aka the 2nd sequence in the multiFa (e.g. chimp), as opposed to the reference/target /1st sequence in the multiFa (e.g. human)?

motif/matchComp.go
```diff
@@ -0,0 +1,288 @@
...
+				log.Fatalf("Unrecognized base.")
+			}
+		}
+
+		// now we check if the current key is a significant motif hit
+		if currRefScore, inKmerHash = kmerHash[currRefKey]; inKmerHash {
+			//fmt.Printf("CurrRefKey: %v. CurrRefScore: %v.\n", currRefKey, currRefScore)
+			minResidual = math.Inf(1)
+			minResidualAltScore = 0
+			for currAltStart = numbers.Max(alnPos-len(pm.Mat[0])-residualWindowSize+1, 0); currAltStart <= alnPos+residualWindowSize-len(pm.Mat[0])+1; currAltStart++ {
```
RavenYL, line 139: would the for loop's end condition also need a numbers.Min between the "lnPos+residualWindowSize-len(pm.Mat[0])+1" and the end/length of the multiFa sequence?

motif/matchComp.go
```diff
@@ -0,0 +1,288 @@
...
+				if !couldScoreSequence {
+					break
+				}
+				currResidual = math.Abs(currRefScore - currAltScore)
+				if currResidual < minResidual {
+					minResidual = currResidual
+					minResidualAltScore = currAltScore
+				}
+				if !enforceStrandMatch {
+					currAltScore, currAltEnd, couldScoreSequence = ScoreWindow(revCompPm, records[1].Seq, currAltStart)
```
RavenYL, line 151: Is it necessary to do revCompPm, when in lines 73 and 77 you already do the scan sequence function for the reverse complement sequence?

motif/matchComp.go
```diff
@@ -0,0 +1,288 @@
...
+				}
+				altEndsConsidered[currAltEnd] = true
+			}
+			if outputAsProportion {
+				currRefScore = currRefScore / consensusScore
+				minResidualAltScore = minResidualAltScore / consensusScore
+				minResidual = math.Abs(currRefScore - minResidualAltScore)
+			}
+			if minResidual >= residualFilter {
+				currBed = bed.Bed{
```
RavenYL, line 170: In motif/match.go you have the below code, which I thought was right (chr1 0 10, motif is positions 0-9 aka length is 10, currPos is 10 after scanning the genome window, so chrEnd is 10, chrEnd is 10-10=0). Is that consistent with the code here? Here, is refPos 9 after scanning the genome window (maybe doesn't get ++ at the end of a loop?), and chrEnd is 9+1=10, chrStart is 9-10+1=0)?
```
					currBed = bed.Bed{Chrom: records[currChrom].Name,
						ChromStart:        currPos - motifLen,
						ChromEnd:          currPos,
						Name:              motifName,
						Score:             0,
						Strand:            strand,
						FieldsInitialized: 7,
						Annotation:        []string{fmt.Sprintf("%f", currScore)}}
```

cmd/tfMatchComp/tfMatchComp.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func main() {
+	var expectedNumArgs int = 4
+	var propMatch *float64 = flag.Float64("propMatch", 0.8, "Specifies the minimum motif score (as a proportion of the consensus sequence score) required for a match to be retained in the output.")
+	var matrixFileType *string = flag.String("matrixFileType", "Frequency", "Specify the type of position matrix file. Can be 'Frequency', 'Probability', or 'Weight'.")
+	var pfmPseudocounts *float64 = flag.Float64("pfmPseudocounts", 0.1, "If a Position Frequency Matrix is provided, this pseudocount value will be applied when converting to a PWM.")
+	var refStart *int = flag.Int("refStart", 0, "Set the reference position for the beginning of the input multiFa alignment.")
+	var outputAsProportion *bool = flag.Bool("outputAsProportion", false, "Display the output motif scores as proportions of the consensus score. Motif difference score will thus be a change in consensus proportion.")
+	var residualWindowSize *int = flag.Int("residualWindowSize", 5, "Set the number of offset bases to consider when searching for motif in other species.")
```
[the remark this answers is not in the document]
RavenYL: Maybe specify that only the window with the min residual in the query sequence will be taken into account (I think that's how the program behaves?) ?

motif/match.go
```diff
@@ -0,0 +1,304 @@
...
+		case dna.G:
+			answer += pm.Mat[2][motifPos]
+			motifPos++
+			seenAnswer += "G"
+		case dna.T:
+			answer += pm.Mat[3][motifPos]
+			motifPos++
+			seenAnswer += "T"
+		case dna.N:
+			return -1, -1, false //this will just ignore this position of the sequence and move on.
```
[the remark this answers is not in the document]
rimangan: Because we return, there is no need to increment the motifPos

motif/match.go
```diff
@@ -0,0 +1,304 @@
...
+	var answer uint64
+	var motifPos = 0
+	for motifPos < motifLen {
+		if alnPos >= len(record.Seq) {
+			return 0, 0, false
+		}
+		switch record.Seq[alnPos] {
+		case dna.N: //if we see an N, we skip this base and clear out answer and the partial key.
+			alnPos++
+			motifPos = 0
```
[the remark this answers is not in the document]
rimangan: A valid key must have all bases (As, Cs, Gs, and Ts). If we see an N, we want to skip over the Ns and start building a new key from scratch in the position following the n. So we reset our answer and set motifPos back to zero, so it's like we're starting over again in the position after the N

motif/match.go
```diff
@@ -0,0 +1,304 @@
...
+				needNewKey = false
+			} else {
+				switch records[currChrom].Seq[currPos] {
+				case dna.N:
+					needNewKey = true
+					continue
+				case dna.Gap:
+					continue
+				case dna.A:
+					currKey = currKey << 2
```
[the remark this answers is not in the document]
rimangan: Exactly right. 01001011 translates to CAGT, for example

motif/match.go
```diff
@@ -0,0 +1,304 @@
...
+		case dna.N: //if we see an N, we skip this base and clear out answer and the partial key.
+			alnPos++
+			motifPos = 0
+			answer = 0
+		case dna.Gap: //keep searching, motifs can span gaps, which are ignored.
+			alnPos++
+		case dna.A:
+			answer = answer << 2 //left shift two bases, clearing a space for the new base.
+			answer = answer | 0  //append 00 in last two spots with bitwise OR.
+			motifPos++
```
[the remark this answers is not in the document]
rimangan: Oh wow, important catch! I'll fix in my next commit.

motif/match.go
rimangan: This is performing Insertion Sort, which runs in O(n^2). Indeed not the fastest, many sorting algorithms run in O(nlogn). However, since N is guaranteed to be 4, the difference should be minimal. I chose this sorting method for ease of programming and readability here

motif/match.go
rimangan: good catch!

motif/match.go
rimangan: This would change the logic. The 'if' will be true for some loop iterations and false for others

motif/matchComp.go
rimangan: I apologize if I have used this inconsistently. Should we standardize with 'target' 'query'?

motif/matchComp.go
rimangan: Great catch, I'll change at 243 as well

motif/matchComp.go
rimangan: The situation for this one is that if our motif is CCC, and the ref is AAACCCAAA, and the alt is AAAGGGAAA, we should count no difference in motif, as the same position has the motif in the ref and alt, albeit the alt has the motif on the opposite strand. So we're doing the rapid 2bit kmer searching for the forward motif for the ref, but then checking both the forward and revcomp motif in the alt. However, this would still miss a GGG in the ref seq, so we'll need to rescan. Hope this clears it up, this is a funny nuance I'm happy to talk about

motif/match_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+func TestBuildKmerHash(t *testing.T) {
+	var motifs []PositionMatrix
+	var answer map[uint64]float64
+	for _, v := range BuildKmerHashTests {
+		motifs = ReadJaspar(v.PwmFile, "Weight")
+		for i := range motifs {
+			answer = buildKmerHash(motifs[i], v.ThresholdProportion)
+			if len(answer) != v.ExpectedLengths[i] {
+				t.Errorf("Error in buildKmerHash. Expected Hash Size: %d. Found: %d.", len(answer), v.ExpectedLengths[i])
```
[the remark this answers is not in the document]
rimangan: Yes. It might be difficult to compare identity here

cmd/tfMatchComp/tfMatchComp.go
```diff
@@ -0,0 +1,84 @@
...
+			"options:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var expectedNumArgs int = 4
+	var propMatch *float64 = flag.Float64("propMatch", 0.8, "Specifies the minimum motif score (as a proportion of the consensus sequence score) required for a match to be retained in the output.")
+	var matrixFileType *string = flag.String("matrixFileType", "Frequency", "Specify the type of position matrix file. Can be 'Frequency', 'Probability', or 'Weight'.")
+	var pfmPseudocounts *float64 = flag.Float64("pfmPseudocounts", 0.1, "If a Position Frequency Matrix is provided, this pseudocount value will be applied when converting to a PWM.")
+	var refStart *int = flag.Int("refStart", 0, "Set the reference position for the beginning of the input multiFa alignment.")
```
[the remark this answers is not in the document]
rimangan: The idea here was you could extract an alignment for your favorite gene, or for a STARR-seq construct, and find all motifs on just that area (maybe just 1kb out of a whole chromosome). However, if we find motifs, we'll want to report their reference positions, not alignment positions. So the user here can tell the program what the starting position of the alignment corresponds to in reference positions

motif/matchComp.go
RavenYL: No worries! 'target' 'query' seems more intuitive and generalized to me, because when I see 'ref' 'alt' I think of segregating sites in vcfs so they give me the feeling that they can only describe 2 versions/populations of the same species, not 2 species, but tfMatchComp can be used to compare any 2 fastas including 2 different species.

motif/matchComp.go
RavenYL: Can you help me understand why missing the GGG (the revcomp motif) in the ref seq is a problem? I thought if you can find the CCC in the ref and the GGG in the alt that's enough, but I'm probably missing something

RavenYL: revCompPm is to calculate revComp sequence score!

RavenYL (approved):

cmd/tfMatch/tfMatch.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	motif.RapidMatch(motifs, records, s.PropMatch, s.OutFile, s.OutputAsProportion)
+}
+
+func usage() {
+	fmt.Print(
+		"tfMatch - Genome-wide scanning of TFBS occurrences." +
+			"Input DNA sequences must be upper case.\n" +
+			"Usage:\n" +
+			"tfMatch input.fa matrices.pfm output.bed\n" +
```
cfauci, line 49: is this .pfm supposed to say .pmf for position matrix file?

cfauci (approved):
I left a reminder comment about the necessary new line character for the usage statement that we saw yesterday with Nicole

PR #926, opened 2023-09-24
bedSimilarity.go

sweavs111:
bedSimilarity.go. This is something that Craig, Raven, and I have been talking about. It's functionality is to provide summary statistics of similarity between bed files. The use case in mind for this program is to compare similarities of haqer sets with different window sizes. It will create a text file with data. You can either present 2 bed files or a file that has list of file names. In the second case, all comparisons will be done. If you provide a file list you can also create an optional heatmap of the results.


interval/tools.go
```diff
@@ -2,9 +2,114 @@ package interval
...
+		return false
+	}
+}
+
+// SortByCoord takes in a slice of Interval and sorts the slice according to coordinates
+func SortByCoord(in []Interval) {
+	sort.Slice(in, func(i, j int) bool { return compare(in[i], in[j]) == -1 })
+}
+
+func compare(a Interval, b Interval) int {
```
RavenYL, line 71: Can you add a description for compare? It looks like you use it in sort, and prioritize either chromStart or chromEnd being smaller (return -1, would match SortByCoord line 68 condition)?

interval/tools.go
```diff
@@ -2,9 +2,114 @@ package interval
...
+	if a.GetChromEnd() < b.GetChromEnd() {
+		return -1
+	}
+	if a.GetChromEnd() > b.GetChromEnd() {
+		return 1
+	}
+	return 0
+}
+
+// Unique takes in a slice of Interval and returns a slice of Interval with only unique interval entries (determined by coordinates). The last instance of each interval is kept.
```
RavenYL, line 91: Would you be keeping the first instance of each interval, since "if AreEqual(currRegion, i)", then the code continues to the next iteration of the for loop without appending currRegion to uniqueRegions?

bed/info.go
```diff
@@ -61,3 +61,22 @@ func IsSelfOverlapping(b []Bed, verbose int) bool {
 	}
 	return false
 }
+
+func OverlapSize(a Bed, b Bed) int {
+	if !Overlap(a, b) {
+		return 0
+	}
+	switch {
```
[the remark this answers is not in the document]
sweavs111: yes this is a great idea!

cmd/haqerBedComps/haqerBedComps.go
```diff
@@ -0,0 +1,194 @@
...
+	for i := range beds {
+		var l matrixLine
+		l.name = beds[i]
+		a = bed.Read(beds[i])
+		intervalsA = interval.BedSliceToIntervals(a)
+		for j = range beds {
+			if beds[i] == beds[j] {
+				//fileio.WriteToFileHandle(out, fmt.Sprintf("%s\t%s\t%f\t%f\t%f", beds[i], beds[j], 1.0, 1.0, 1.0))
+				if s.matrixComponents != "" || s.matrixAverage != "" {
+					l.vals = append(l.vals, 1.0)
```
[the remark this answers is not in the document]
sweavs111: The matrix components are only if you want to make a heat-map. In the default output which is just a table of values, i decided it would not be necessary to output the self comparisons, which are all just 1

cmd/haqerBedComps/haqerBedComps.go
```diff
@@ -0,0 +1,194 @@
...
+			allFiles = append(allFiles, i)
+		}
+		tail := strings.Join(allFiles, "\t")
+		fileio.WriteToFileHandle(outMatrix, tail)
+		err = outMatrix.Close()
+		exception.PanicOnErr(err)
+	}
+}
+
+func writeMatrixLine(line matrixLine, outMatrix *fileio.EasyWriter) {
```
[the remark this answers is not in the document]
sweavs111: I am not exactly sure what you mean by this, but I am all for efficiency! If theres a better way to do this I would love to know

cmd/haqerBedComps/haqerBedComps.go
```diff
@@ -0,0 +1,194 @@
...
+func doWork(s settings) {
+	if s.list == "" {
+		compareTwo(s)
+	} else {
+		multipleComparisons(s)
+	}
+}
+
+func usage() {
+	fmt.Print("haqerBedComps -- Takes in 2 bed files or a list of bed files and give similarity statistics based on number of overlaps" +
```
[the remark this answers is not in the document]
sweavs111: yes you are exactly correct, I am planning on changing it, but I wanted it to match the file name for now

interval/tools.go
sweavs111: good catch!

cmd/haqerBedComps/haqerBedComps.go
```diff
@@ -0,0 +1,194 @@
...
+	if s.matrixAverage != "" {
+		outMatrix = fileio.EasyCreate(s.matrixAverage)
+	}
+	if s.matrixComponents != "" {
+		outMatrix = fileio.EasyCreate(s.matrixComponents)
+	}
+
+	beds := fileio.Read(s.list)
+	for i := range beds {
+		var l matrixLine
```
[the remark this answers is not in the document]
sweavs111: I need the variable 'l' to be nil at the start of every j loop. I thought the best way to do it was to redeclare it here. Is there a better way to do this?

cmd/haqerBedComps/testdata/expected.matrixComp.txt
```diff
@@ -0,0 +1,4 @@
+testdata/smallAJ.bed	1.000000	0.750000	0.750000
+testdata/largeAJ.bed	0.666667	1.000000	0.333333
+testdata/thirdBed.bed	0.500000	0.500000	1.000000
+x	testdata/smallAJ.bed	testdata/largeAJ.bed	testdata/thirdBed.bed
```
RavenYL, line 4: Is it easy to move this line to the beginning of the file, rather than the end? it might make R processing easier. Thank you!

cmd/haqerBedComps/haqerBedComps.go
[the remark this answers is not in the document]
sweavs111: i just set l.vals to an empty slice of float64. the code looks a lot better. good suggestion!

cmd/haqerBedComps/testdata/expected.matrixComp.txt
sweavs111: done!

RavenYL (commented):
I think there are a few more code efficiency comments that haven't been discussed/addressed but after that I will approve this because the program works well for me on real data!

hcnapier (commented):
I can't speak to the efficiency tweaks Raven mentioned, but everything here makes sense and looks good to me.

bed/info.go
```diff
@@ -61,3 +61,11 @@ func IsSelfOverlapping(b []Bed, verbose int) bool {
 	}
 	return false
 }
+
+// OverlapSize will return an int corresponding to the overlap size between two Bed elements. If the elements are non-overlapping it will return 0.
```
RavenYL, line 65: I just found that there's a function called OverlapLength in bed/compare.go. Should these 2 be merged?

cmd/bedSimilarity/bedSimilarity.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if s.matrixAverage != "" || s.matrixComponents != "" {
+		for _, i := range files {
+			allFiles = append(allFiles, separatePath(i))
+		}
+		header := strings.Join(allFiles, "\t")
+		fileio.WriteToFileHandle(outMatrix, header)
+	}
+
+	for i := range files {
```
edotau, line 78: If I am following your variable names correctly, something like this should work:

```golang
func multipleComparisons(s settings) {
	// var a, b []bed.Bed
	var intervalsA, intervalsB []interval.Interval
	var percA, percB, avg float64
	var outMatrix *fileio.EasyWriter
	var err error
	var j int
	var l matrixLine
	var aName, bName string
	var allFiles []string = []string{"x"}

	out := fileio.EasyCreate(s.outFile)
	fileio.WriteToFileHandle(out, "A\tB\tproportion overlaps of A in B\tproportion overlaps of B in A\tbedSimilarityScore")

	if s.matrixAverage != "" {
		outMatrix = fileio.EasyCreate(s.matrixAverage)
	}
	if s.matrixComponents != "" {
		outMatrix = fileio.EasyCreate(s.matrixComponents)
	}

	files := fileio.Read(s.list)

	if s.matrixAverage != "" || s.matrixComponents != "" {
		for _, i := range files {
			allFiles = append(allFiles, separatePath(i))
		}
		header := strings.Join(allFiles, "\t")
		fileio.WriteToFileHandle(outMatrix, header)
	}

	// Read data from files once and store
	numFiles := len(files)
	fileData := make([][]bed.Bed, numFiles)
	intervalData := make([][]interval.Interval, numFiles)
	names := make([]string, numFiles)

	for i, file := range files {
		fileData[i] = bed.Read(file)
		intervalData[i] = interval.BedSliceToIntervals(fileData[i])
		names[i] = separatePath(file)
	}

	for i := 0; i < numFiles; i++ {
		l.name = names[i]
		aName = names[i]
		l.vals = []float64{}
		// a = fileData[i]
		intervalsA = intervalData[i]
		for j = range files {
			if files[i] == files[j] {
				if s.matrixComponents != "" || s.matrixAverage != "" {
					l.vals = append(l.vals, 1.0)
				}
				continue
			}
			// b = fileData[j]
			bName = names[j]
			intervalsB = intervalData[j]
			percA, percB, avg = interval.IntervalSimilarity(intervalsA, intervalsB)
			if j > i {
				fileio.WriteToFileHandle(out, fmt.Sprintf("%s\t%s\t%f\t%f\t%f", aName, bName, percA, percB, avg))
			}

			if s.matrixAverage != "" {
				l.vals = append(l.vals, avg)
			} else if s.matrixComponents != "" {
				l.vals = append(l.vals, percA)
			}
		}
		if s.matrixAverage != "" || s.matrixComponents != "" {
			writeMatrixLine(l, outMatrix)
		}
	}

	err = out.Close()
	exception.PanicOnErr(err)
	if s.matrixAverage != "" || s.matrixComponents != "" {
		err = outMatrix.Close()
		exception.PanicOnErr(err)
	}
}
```
- I added the loop to read and process the files and storing the data/names/etc. and commented out `fileStorage` so you can easily follow where I made modifications. This should be a working implementation (i think)

edotau (dismissed):
### Left one suggestion:

A major optimization we can implement here is to avoid reading the files repeatedly in a nested loop, but just looping over the files once, save them and access the index at a later time.

```golang
// Read data from files once and store
numFiles := len(files)
fileData := make([][]bed.Bed, numFiles)
intervalData := make([][]interval.Interval, numFiles)
names := make([]string, numFiles)

for i, file := range files {
	fileData[i] = bed.Read(file)
	intervalData[i] = interval.BedSliceToIntervals(fileData[i])
	names[i] = separatePath(file)
}
```

bed/info.go
edotau: Yeah I think @RavenYL is right, if you if you look at [vertgenlab/gonomics/bed/compare.go](https://github.com/vertgenlab/gonomics/blob/main/bed/compare.go#L105-L113), something similar has been implemented. I am in favor of updating `OverlapLength` to:

`return numbers.Min(a.ChromEnd, b.ChromEnd) - numbers.Max(a.ChromStart, b.ChromStart)`

bed/info_test.go
```diff
@@ -41,3 +41,23 @@ func TestIsSelfOverlapping(t *testing.T) {
 		}
 	}
 }
+
+func TestOverlapSize(t *testing.T) {
```
edotau, line 45: Once you update OverlapLength, you can prob get rid of this

edotau: Actually, should try to sync with @RavenYL, because there's something similar in PR: #928

bed/info.go
sweavs111: lol i tried to look to see if this function already existed aparently i just missed it. I will delete overlapSize and edit overlapLength to what you suggested Eric

cmd/bedSimilarity/bedSimilarity.go
sweavs111: this looks awesome, I'll dive into this deeper today!

sweavs111: Made these improvements thanks! The only think i changed from your suggested code is that instead of having the intermediate [][]bed.Bed to create the [][]interval.Interval, I put the bed.Read(file) right into the bedSliceToIntervals function, which eliminated the need for the fileData variable

RavenYL (approved):

edotau (approved):

PR #927, opened 2023-09-27
Gaf filter to remove entries if the qualifier contains NOT 

hcnapier:
# Description

Gaf filter to remove entries if the qualifier contains NOT.

## Relevant Links

### Testing

None

### Screenshots & Media

None

None


RavenYL (dismissed):

cmd/gafFilter/gafFilter.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+type Settings struct {
+	InFile    string
+	OutFile   string
+	RemoveNot bool
+}
+
+func gafFilter(s Settings) {
+	var pass bool = false
```
sweavs111, line 23: style: if you set pass to true at the begining of each loop below, do you have to explicitely define it as false here?

cmd/gafFilter/testdata/expected.gaf
sweavs111, line 1: do the test files need to be this big? Not a big issue but it seems like they could just be a few lines

sweavs111 (commented):
Can't comment on if this is the "desired feature" like Raven asked, but the code looks good! Just had 2 little style points / questions.

cmd/gafFilter/testdata/expected.gaf
hcnapier: You're right, they don't have to be this big.

cmd/gafFilter/gafFilter.go
hcnapier: Probably not, good point

cmd/gafFilter/gafFilter.go
```diff
@@ -0,0 +1,76 @@
+// Command Group:
```
craiglowe, line 1: Should we pick a command group for this program?

cmd/gafFilter/gafFilter.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+	var err error
+	err = out.Close()
+	exception.PanicOnErr(err)
+}
+
+func usage() {
+	fmt.Print(
+		"bedFilter\n" +
```
craiglowe, line 49: This usage message needs an update from "bedFilter" to "gafFilter"

craiglowe (commented):
Looks great!  I just had two small comments about placing this in a command group, and that the usage message has not been updated from bed to gaf

cmd/gafFilter/gafFilter.go
cfauci: if you run the gonomics command it will print out all the groups

cmd/gafFilter/gafFilter.go
```diff
@@ -0,0 +1,76 @@
...
+
+	gafChan, header := gaf.GoReadToChan(s.InFile)
+	out := fileio.EasyCreate(s.OutFile)
+
+	gaf.WriteHeaderToFileHandle(out, header)
+
+	for curr := range gafChan {
+		pass = true
+		if s.RemoveNot {
+			if strings.Contains(curr.Qualifier, "NOT") {
```
cfauci, line 33: could we make this work only on the column where we expect NOTs to appear instead of the whole gaf line?

cmd/gafFilter/gafFilter.go
```diff
@@ -0,0 +1,76 @@
+// Command Group:
+
+// Output a subset of a gaf file without "NOT" entries in qualifier field
```
cfauci, line 3: the usage statement is a little too specific you can just describe that it filters gaf entries, even if right now all if does is remove NOTs

cfauci (approved):

sweavs111 (approved):
Looks great!

PR #928, opened 2023-10-04
intervalSubsetMatrix fraction option

RavenYL:
# Description
- added fraction option to intervalSubsetMatrix
- added to the usage that input intervals must be unique. In each file, each genomic region can only appear in no more than 1 interval
- changed program algorithm to account for cases where 1 file's 1 interval overlaps multiple intervals in the union.interval (no longer assumes that len(currOverlap) can only be 0 or 1. This introduces ranging over currOverlap with index k)
- for the fraction option, took into account cases where 1 interval in the union.interval could overlap multiple intervals in 1 file, and that affects the overlap size and fraction
- the purpose of this option is to make a heatmap of each haqer (union.interval) overlapping each tissue's chromatin state peaks (files.list), taking into account that 1 haqer may overlap multiple peaks and so fractions are needed to specify this 


interval/tools.go
```diff
@@ -2,9 +2,28 @@ package interval
...
 // CoordsToString takes in a interval and returns a string in the format of chr:start-end
 func CoordsToString(i Interval) string {
 	return fmt.Sprintf("%s:%v-%v", i.GetChrom(), i.GetChromStart(), i.GetChromEnd())
 }
+
+func IntervalSize(i Interval) int {
+	return i.GetChromEnd() - i.GetChromStart()
+}
+
+func OverlapSize(a, b Interval) int {
```
edotau, line 17: Are we making the assumption these chromosomes are from the same genome?

interval/tools.go
```diff
@@ -2,9 +2,28 @@ package interval
...
+	if a.GetChrom() != b.GetChrom() {
+		return 0
+	}
+	end := numbers.Min(a.GetChromEnd(), b.GetChromEnd())
+	start := numbers.Max(a.GetChromStart(), b.GetChromStart())
+	overlapSize := end - start
```
edotau, line 23: You can probably get rid of this and check directly and return the result.
See the suggested changes for _**lines 23, 24 and 27**_
```suggestion
	// overlapSize := end - start
```

interval/tools.go
```diff
@@ -2,9 +2,28 @@ package interval
...
+	if overlapSize <= 0 {
```
edotau, line 24:
```suggestion
	if end <= start {
```

interval/tools.go
```diff
@@ -2,9 +2,28 @@ package interval
...
+		return 0
+	} else {
+		return overlapSize
```
edotau, line 27:
```suggestion
		return end - start
```

interval/tools.go
RavenYL: Yes! Good point. Will add a note to usage

edotau (dismissed):

hcnapier (commented):
This all looks good and makes sense to me. More of a usage question: would we ever want to know whether one HAQER is overlapping multiple peaks? Here we have a great way to account for multiple peak overlaps occurring, but I don't think we have a way to tell when it's happening.

RavenYL:
> This all looks good and makes sense to me. More of a usage question: would we ever want to know whether one HAQER is overlapping multiple peaks? Here we have a great way to account for multiple peak overlaps occurring, but I don't think we have a way to tell when it's happening.

I can imagine that we might! I added an option for this

edotau (approved):

interval/tools.go
```diff
@@ -2,9 +2,29 @@ package interval
...
 }
+
+// IntervalSize calculates the size of the interval
+func IntervalSize(i Interval) int {
+	return i.GetChromEnd() - i.GetChromStart()
+}
+
+// OverlapSize calculates the size of the overlap betwen 2 intervals, assuming they are from the same genome
+func OverlapSize(a, b Interval) int {
+	if a.GetChrom() != b.GetChrom() {
```
sweavs111, line 20: It might be nice anyways to have a Overlap() function in the interval package that just returns a bool if to intervals overlap or not. Then this function could look a little cleaner. Something like
`if !Overlap(a, b) {
return 0
} else {
return numbers.Min(a.GetChromEnd(), b.GetChromEnd()) - numbers.Max(a.GetChromStart(), b.GetChromStart())
}`

Purely stylistic though because this code works as is. Theres a similar set of functions in Bed if you want to look at them

cmd/intervalSubsetMatrix/intervalSubsetMatrix.go
```diff
@@ -32,17 +33,45 @@ func intervalSubsetMatrix(unionFile string, fileListFile string, outFile string)
...
 		currIntervalChan = interval.GoReadToChan(files[i])
 		for j = range currIntervalChan {
-			currOverlaps = interval.Query(unionTree, j, "any")
+			currOverlaps = interval.Query(unionTree, j, "any") // each j (1 bed region from the file of the column) can overlap 0 or 1 or more bed region of the row
 			if len(currOverlaps) > 0 {
-				mat[interval.CoordsToString(currOverlaps[0])][i] = 1
+				for _, k := range currOverlaps { // k >= 1 would mean j overlaps more than 1 bed region of the row
+					if fraction {
+						overlapSize = interval.OverlapSize(k, j)
+						mat[interval.CoordsToString(k)][i] += float64(overlapSize) / float64(interval.IntervalSize(k)) // add number rather than assign number, in case 1 bed region of the row overlaps multiple bed regions in the files of the columns
```
sweavs111, line 59: So if I understand correctly, we are reporting the proportion of bases in a haqer element that overlap an interval set. Is that correct? When haqer window sizes are very large even if the haqer overlaps an small interval completely, it will return a small proportion. Is that the functionality you want?

sweavs111 (commented):
Looks good to me! One clarifying question and one stylistic option. Like Eric said looks like there's a merge conflict as well!

cmd/intervalSubsetMatrix/intervalSubsetMatrix.go
RavenYL: Yes!

sweavs111: great! Everything looks good on my end. approving now

sweavs111:
Great!

sweavs111 (dismissed):
looks great!

edotau (approved):

interval/tools.go
RavenYL: I want to add this so will have more commits!

edotau (approved):

hcnapier (approved):

PR #935, opened 2023-10-12
Probabilistic sequence representations with the pFasta file format

rimangan:
# Description
Here is a proof of concept towards an analog to fasta for probabilistic representations of DNA sequences. Here, each base in a sequence is encoded by a 4-state probability vector, representing the probabilities that a base at that position is A, C, G, or T, respectively. 
In the pFasta (.pFa) file format, these probability vectors are encoded in 64 bits of binary, or 4 sets of 16 bits. Thus, each base probability is encoded in 16 binary bits, and is parsed initially into a float16 half-precision float, and then into a float32 for our internal data structure.
The binary .pFa format begins with a metadata header before the binary. See the example below:
`pFasta_format_1.0\nchr1\t2\nchr2\t3\nEndHeader\n<<begin binary>>`
The metadata header encodes the chromosome names, but also the lengths, much like a chromSizes file. This allows us to know how many binary bits to expect for each chromosome in the pFa, and also a means for the detection of truncated input files.
I've started this draft with a basic reader and writer for pFa. In the future I will expand this with functions to read only subsequences (perhaps specified by beds). Since this is still an early draft, now would be the perfect time for large structural feedback on how we would like to represent probabilistic DNA and pDna sequences going forward in Gonomics.

### Testing

None


dna/pDna/pDna.go
```diff
@@ -0,0 +1,78 @@
...
+	G uint8
+	T uint8
+}
+
+type Uint32Base uint32
+
+// Pack takes an input base in Uint8Base format (where each base probability A, C, G, and T is stored in a uint8)
+// and converts it to a single uint32Base, where each consecutive 8 bits corresponds to the probability of one of the
+// four bases.
+func Pack(base Uint8Base) Uint32Base {
```
dasnellings, line 32: for the uint8base idea I don't think the un/pack functions are necessary. The struct is 32 bit anyways so why not just leave it in struct form? 

Though if we go with Craig's idea of a trimmed down float then conversion functions will be necessary.

rimangan: A few considerations:
1. Internally, we might always convert to floats for most use cases, so the uint representations may just be for reading and writing to disk.
2. If that's the case, we'll probably never use the Uint8Base struct and want to convert directly between the Uint32Base and Float32Base. 
3. So with that said, I'm supportive of removing either the Uint32Base or the Uint8Base struct/type, to eliminate redundancy. I think either solution makes sense to me, though I assume the Float32Base is slightly more lightweight

dna/pDna/pDna.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return uint8(math.Round(-10 * math.Log2(float64(prob))))
+}
+
+// Uint8ToFloat32Prob converts an encoded probability in uint8 format into a probability in float32 format.
+func Uint8ToFloat32Prob(u uint8) float32 {
+	if u == 0 {
+		return 0
+	}
+	return float32(1 / math.Pow(2, float64(u)/10.0))
```
dasnellings, line 77: As I mentioned before, I like Craig's idea of using a modified float stored as a uint8 or uint16. I expect shifting bits around is going to be way faster than casting to float64, calling math.Pow, then casting back. If we give the exponent 5 bits we can have values down to e-32 and if we give the significand 10 bits we can store up to 1024 (enough for 3 significant digits) and we've got 1 bit left to play with.

dna/pDna/pDna.go
rimangan: I agree, sounds like a great plan! I/O from a binary file format into that struct should be straightforward as well

dna/pDna/pDna.go
```diff
@@ -0,0 +1,43 @@
+package pDna
+
+import "math"
+
+// Float32Base encodes a DNA base as a probability vector using float32 precision.
+// Note that gap probabilities can be stored implicitly as 1 - (A + C + G + T).
+type Float32Base struct {
```
dasnellings, line 7: I missed the discussion at code review, but I thought the idea was to make base probabilities float16's in memory, not on disk. Did you guys decide to make the bases float32's in memory instead?

rimangan: Sorry, it's float16 on disk, float32 in memory. The idea is we'll never try to store a whole genome in memory at once, usually reading will be paired with something like fasta.Extract, where we pull our subsequence of interest out of a large file. That's the vision, obviously missing the context in that the Extract command for pFa is still pending

dasnellings (approved):

dna/pDna/pDna.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// equalFloatPrecision is a helper function of EqualBase, which allows us to determine if
+// two float32 are 'equal' within a user-specified degree of precision. Note that this is a
+// relative precision (normalized to one of the values), and catches divide by zero errors.
+func equalFloatPrecision(a float32, b float32, precision float32) bool {
+	if a == 0 && b == 0 {
+		return true
+	}
+	if a == 0 {
```
RavenYL, line 39: I think technically this should be else if a == 0, but since there is a return above it doesn't matter

fasta/pFasta/pFasta.go
```diff
@@ -0,0 +1,134 @@
...
+func ReadPfaHeader(reader *bufio.Reader) []string {
+	var header []string
+	var line string
+	var err error
+
+	for { // this loop will end either at the end of the file (with panic), or when 'EndHeader' is reached.
+		line, err = reader.ReadString('\n')
+		exception.PanicOnErr(err)
+		line = strings.TrimSuffix(line, "\n")
+		line = strings.TrimSuffix(line, "\r")
```
RavenYL, line 94: Hopefully this works with both mac and windows machines (parsing \n and \r)! Sometimes my unix code doesn't work on both but hopefully golang doesn't have this problem

RavenYL (approved):

PR #936, opened 2023-10-15
faFormat -rename

rimangan:
# Description
This option to faFormat allows the user to 'rename' an entry. Consider the following usage:
faFormat -rename old:new in.fa out.fa

With this infile:
>old
ACTGATAC
We would see this outfile:
>new
ACTGATAC

### Testing

None


RavenYL (approved):

cfauci (approved):

craiglowe (approved):
Great!

PR #937, opened 2023-10-18
intervalOverlap nonOverlap Bug Fix

rimangan:
# Description
This PR fixes an issue in intervalOverlap where 'nonOverlap' always returns empty files. The issue is that on main, nonOverlapping entries are evaluated as 'nil' in the answerChan, but when we write to file, we check if the len(answer)==0, which will always be false for nil entries. Here, nil returns are replaced with a slice of len(0).
I also expanded testing to cover 'nonOverlap' and 'mergedOutput', which were not covered on main. Hopefully this will provide stability testing going forward.

### Testing

None


cmd/intervalOverlap/intervalOverlap.go
```diff
@@ -101,7 +102,6 @@ func main() {
 	if len(flag.Args()) != expectedNumArgs {
 		flag.Usage()
 		log.Fatalf("Error: expecting %d arguments, but got %d\n\n", expectedNumArgs, len(flag.Args()))
-		return
```
rimangan, line 104: Just to confirm, this return doesn't need to be here? We're in a main function

dasnellings: Correct, the fatal will kill it before the return.

dasnellings (approved):

cmd/intervalOverlap/intervalQuery.go
```diff
@@ -56,13 +56,14 @@ func queryWorker(tree map[string]*interval.IntervalNode, queryChan <-chan interv
 			answer = answerTrue
 			answerChan <- &queryAnswer{query, answer}
 		} else {
-			answer = nil
+			answer = make([]interval.Interval, 0)
```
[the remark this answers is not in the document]
rimangan: I actually didn't know this

craiglowe (approved):
Wonderful!  Thank you!  Great catch!

PR #938, opened 2023-10-20
add fastq sorting to mergesort

sweavs111:
Add option for mergesort.go for sorting fastq files by name. Can accept paired end fastq files with the -fastqPE option. All that option does is loop over both files and sort them both individually, but it makes it so you don't have to run the program twice. This is a fairly short PR and pretty boilerplate compared to the rest of the mergesort functions


cmd/mergesort/mergesort.go
```diff
@@ -226,14 +253,19 @@ func mergeSort(filename string, outFile string, numRecordsPerChunk int, sortCrit
 }
 
 func main() {
-	expectedNumArgs := 2
 	var numLinesPerChunk *int = flag.Int("tmpsize", 1000000, "The number of records to read into memory before writing to a tmp file.``")
 	var singleCellBx *bool = flag.Bool("singleCellBx", false, "Sort single-cell sam records by barcode.")
 	var sortCriteria string = "byGenomicCoordinates" // default the genomicCoordinates criteria.
+	var fastqPE *bool = flag.Bool("fastqPE", false, "Sort paired end fastq files. Each file will be sorted separately")
```
rimangan, line 259: Can we specify somewhere that these are being sorted by name? I notice that fastqFilter does not take in the sortCriteria string

sweavs111: updated usage

RavenYL (approved):

rimangan (approved):

PR #939, opened 2023-10-20
Get promoter

cfauci:
# Description
getPromoter is designed to return the upstream and downstream region around a TSS. the distance returned is given by the user and defaults to 1kb upstream and 200bp downstream. The user has to give a list of unique gene names and a gtf file as input.

### Testing

None


cfauci (commented):
This has now been checked against real world data and works as expected, including if you just want to return the TSS for a gene, you can set both downstream and upstream to 0 and it will make a single base pair TSS bed with the name of the gene

gtf/filter.go
```diff
@@ -59,3 +61,30 @@ func FilterVariantFiveUtr(v *vcf.Vcf, g map[string]*Gene, c map[string]*chromInf
...
+	var trans *Transcript
+	var newBed = bed.Bed{Chrom: "", ChromStart: 0, ChromEnd: 0, FieldsInitialized: 4}
+
+	for currGene = range genes {
+		name = genes[currGene]
+		for currRecord := range gtf {
+			if gtf[currRecord].GeneName == name {
+				for transcript = range gtf[currRecord].Transcripts {
+					trans = gtf[currRecord].Transcripts[transcript]
+					if trans.Strand {
```
RavenYL, line 78: Can trans.Strand can be checked like a bool? It seems like the lines in gtfFileTest.gtf all have + so I wonder

cmd/getPromoter/getPromoter.go
```diff
@@ -0,0 +1,57 @@
...
+
+	bed.Write(outBed, answer)
+}
+
+func usage() {
+	fmt.Print(
+		"getPromoter will take a list of unique genes of interest and return a bed file of the promoter region proceeding\n" +
+			"the position of the TSS for each isoform by the amount specified with upstream and the amount following the TSS as specified by downstream. This program is strand aware.\n" +
+			"Usage:\n" +
+			"getPromoter [options] uniqueGenes.txt geneInfo.gtf out.bed chrom.sizes\n" +
```
RavenYL, line 32: Would it be helpful to specify in the lines above that geneInfo.gtf and chrom.sizes need to be provided as input?

gtf/filter.go
cfauci: Yeah, it is stored as a bool if you look at the gtf struct because . Isn't an option

rimangan (approved):
Had some comments, nothing essential.

RavenYL (approved):

PR #940, opened 2023-10-26
Polishing haplotypeGenerator

rimangan:
# Description
A few improvements to this program:

- User options to control line length in output and verbose status.
- No longer makes a tree of variants (this was memory prohibitive for large vcf sets). Now uses channels and checks overlaps against bed regions.
- Modernized the usage, testing, and settings framework
- Catches SVs in the VCF file, and excludes them (haplotypeGenerator currently only supports substitutions).
- Changes some data structures to avoid storing all of the haplotypes in memory.

### Testing

None

### DIFF
--- a/cmd/haplotypeGenerator/haplotypeGenerator.go
+++ b/cmd/haplotypeGenerator/haplotypeGenerator.go
@@ -6,8 +6,6 @@ package main
 import (
 	"flag"
 	"fmt"
-	"log"
-
 	"github.com/vertgenlab/gonomics/bed"
 	"github.com/vertgenlab/gonomics/dna"
 	"github.com/vertgenlab/gonomics/exception"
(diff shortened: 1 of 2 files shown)


cfauci (approved):

cmd/haplotypeGenerator/haplotypeGenerator.go
```diff
@@ -6,83 +6,83 @@ package main
...
-
-	for i := range variants {
-		variantIntervals[i] = variants[i]
-	}
-
-	tree := interval.BuildTree(variantIntervals)
+type Settings struct {
+	ReferenceGenomeFile string
+	VcfFile             string
+	RegionBedFile       string
```
sweavs111, line 22: What is the purpose of the bed file? To subset the genome into regions that you care about? Concievably, this would be a no-gap.bed in most cases?

sweavs111: or is the bed file telling you were haplotype blocks exist?

cmd/haplotypeGenerator/haplotypeGenerator.go
```diff
@@ -6,83 +6,83 @@ package main
...
-					currState = int(overlappingVariants[k].(vcf.Vcf).Samples[currSampleIndex].Alleles[0])
-					if currState > 0 {
-						sampleHaplotypes[j].Seq[overlappingVariants[k].(vcf.Vcf).Pos-regions[i].ChromStart-1] = dna.StringToBase(overlappingVariants[k].(vcf.Vcf).Alt[currState-1])
-					}
-				} else {
-					currState = int(overlappingVariants[k].(vcf.Vcf).Samples[currSampleIndex].Alleles[1])
-					if currState > 0 {
-						sampleHaplotypes[j].Seq[overlappingVariants[k].(vcf.Vcf).Pos-regions[i].ChromStart-1] = dna.StringToBase(overlappingVariants[k].(vcf.Vcf).Alt[currState-1])
-					}
+				currState = int(overlappingVariants[k].Samples[j].Alleles[0])
```
sweavs111, line 66: I am a little confused how you know what SNPs are on haplotype A vs B. Is that something specified by the VCF file? With alleles[0] ?

sweavs111 (approved):
I had a few conceptual questions, but the code looked good to me, so i am approving!

cmd/haplotypeGenerator/haplotypeGenerator.go
```diff
@@ -6,83 +6,83 @@ package main
...
-	// fmt.Println("started ok")
-
-	for i := range variants {
-		variantIntervals[i] = variants[i]
-	}
-
-	tree := interval.BuildTree(variantIntervals)
+type Settings struct {
+	ReferenceGenomeFile string
+	VcfFile             string
```
sweavs111, line 21: is there a check anywhere in this program that confirms that the VCF file has phase data

cmd/haplotypeGenerator/haplotypeGenerator.go
rimangan: Yes, this is specified on the VCF

cmd/haplotypeGenerator/haplotypeGenerator.go
rimangan: No. In an unphased VCF, it will still make haplotypes based on diploid variant calls, though they are not guaranteed to be in phase

PR #942, opened 2023-10-30
samAssembler Part 9: Empirical Bayes

rimangan:
# Description
Back to samAssembler after a long hiatus. This PR enables samAssembler to perform variant calling for pseudoassembly with user-generated empirical priors, following an [Empirical Bayes](https://en.wikipedia.org/wiki/Empirical_Bayes_method) approach. (note that this is a Multinomial-Dirichlet model, where we have an empirical, conditional Dirichlet prior distribution).
With the previous versions of samAssembler, the user could specify an informative prior by manually inputting prior parameters, including delta (the divergence rate) and gamma (the transition/transversion bias). These parameters were fed into a mutation model that assumed Hardy-Weinberg equilibrium with no population structure or inbreeding. 
Empirical priors, however, allow us to avoid the user-input hyperparameter values of delta and gamma, as these values are instead estimated from the data. Furthermore, this method relaxes the assumptions about HWE, population structure, and inbreeding. In principle, this should be a large improvement, as samAssembler will achieve greater accuracy in situations where the mutation model is unknown and where there is significant population structure.
The empirical prior functionality also supports pseudocounts for regularization, which can prevent against overfitting.
I'm working on a large-scale test to demonstrate the performance difference with this option, and also I plan to write up the symbolic, formal mathematical treatment of this option.

IMPORTANT NOTE:
A lot of lines changed are from moving files around. I moved a lot out of samAssembler.go and into other files to break up the large file.

### Testing

None


RavenYL:
Note to self: the new functional code in this PR is to calculate priors empirically. There is a lot of rearrangement/renaming code

cmd/samAssembler/prior.go
```diff
@@ -0,0 +1,197 @@
+package main
```
RavenYL, line 1: Note to self: this file (cmd/samAssembler/prior.go) seems like the main new functional code

cmd/samAssembler/build.go
```diff
@@ -0,0 +1,440 @@
+package main
```
RavenYL, line 1: Note to self: it seems like this file is making settings structs

RavenYL (approved):

hcnapier (approved):

PR #943, opened 2023-10-31
digestGenome.go

sweavs111:
This PR is for a new command called digestGenome. It takes in a fasta file and will create a bed file corresponding to restriction digest fragments that would be created after digestion with an enzyme. This restriction digest bed file is useful in Hi-C data processing pipelines. Any sequence motif can be given to this program, and several enzymes that are commonly used in Hi-C library preps are preset options that can be used. This program runs fairly quick--2-3 minutes for the human genome.

The other change I made was to add a function to the DNA package. I added a function called ReturnReverseCompliment, which returns the rev comp sequence instead of editing it in place, which is what the other ReverseCompliment function does. Please let me know if there is a better way to do this, or another workaround I should be using instead. Thanks!

All testing is complete and function, the PR is ready for review!


hcnapier (commented):
I think this looks good. One extra thing I think we covered in code review last week is possibly erroring out if pal == false in order to avoid accidental misuse.

hcnapier (dismissed):

cmd/digestGenome/digestGenome.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+
+	fa := fasta.GoReadToChan(genome)
+	for i := range fa {
+		prevCut = 0 //set the first fragment start base to 0
+		numCut = 0
+		currChrom = i.Name
+		for base = 0; base <= len(i.Seq)-len(re.cutBases); base++ { //loop over every base in the fasta file
+			testBases := i.Seq[base : base+len(re.cutBases)] //get the next base in the fasta file plus the number of bases in the recognition motif. I don't get an index out of range error at the end of the sequence--not sure why
```
rimangan, line 94: I would declare the testBases variable outside the loop to cut down on allocations

sweavs111: good catch

rimangan (approved):

PR #945, opened 2023-11-02
added allocate:traceback into lastz writer output

cfauci:
# Description
needed to add an additional argument to lastZ writer that I discovered previously but hadn't put into the code before


sweavs111 (approved):
looks great

craiglowe (approved):
Great!

hcnapier (approved):

PR #947, opened 2023-11-06
small fix

dasnellings:
Small fix for linear index in bam seeking. Missed the equals case in my last PR.


rimangan (approved):

sweavs111 (approved):
👍

craiglowe (approved):
Can we add the test that was "missing" as a second PR?  I am guessing there is a test that should have been in our codebase that would have failed before this fix?

PR #950, opened 2023-11-18
PwmTools Shuffle and Hierarchical Command Fix

rimangan:
# Description
This PR addresses an issue preventing the user from using options in hierarchical cmds. Additionally, it introduces the 'shuffle' subcommand of pwmTools, which allows the user to create column-wise shuffled versions of input motifs.

### Testing

None


craiglowe:
What is the current status of this PR?  From the description I was expecting a quick bug fix, but from the diff I am seeing what I think are changes not related to the bug fix.

rimangan:
> What is the current status of this PR? From the description I was expecting a quick bug fix, but from the diff I am seeing what I think are changes not related to the bug fix.
I have updated the title and description of this PR. With the merge of the wig PR, this is now ready for review.

cmd/pwmTools/shuffle.go
```diff
@@ -0,0 +1,86 @@
...
+	fmt.Printf("pwmTools shuffle - a tool for producing shuffled control motifs from input motifs.\n" +
+		"pwmTools shuffle in.pwm out.pwm\n" +
+		"options:\n")
+	shuffleFlags.PrintDefaults()
+}
+
+func parseShuffleArgs() {
+	var expectedNumArgs int = 2
+	var err error
+	shuffleFlags := flag.NewFlagSet("shuffle", flag.ExitOnError)
```
cfauci, line 31: I think the below usages need a little more info, I'm not sure what they do still.

cmd/pwmTools/shuffle.go
```diff
@@ -0,0 +1,86 @@
+package main
```
cfauci, line 1: I'm not sure what the purpose of this command is. Why do you need it? Functions missing documentation and command missing group and documentation

cfauci (commented):
All items in command (maybe not info) should have command group and documentation. 

Functions are generally missing documentation

cmd/pwmTools/info.go
```diff
@@ -61,10 +65,16 @@ func pwmInfo(s InfoSettings) {
 	var consensusScore float64
 	var couldScoreConsensusSequence bool
 	var records []motif.PositionMatrix
+	var currCache map[uint64]float64
+
+	if s.Threshold > 1 || s.Threshold < 0 {
```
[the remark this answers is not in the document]
rimangan: It can, s.Threshold may equal 1, but can't be greater than 1

cmd/pwmTools/shuffle.go
rimangan: I agree, I've expanded the usages in the next commit

cmd/pwmTools/format.go
```diff
@@ -0,0 +1,101 @@
...
+		case "Weight":
+			records = motif.PfmSliceToPpmSlice(records, s.PseudoCount)
+			records = motif.PpmSliceToPwmSlice(records)
+		default:
+			log.Fatalf("Error: unrecognized output motif file format. Options are 'Frequency', 'Probability', and 'Weight'. Found: %v.\n", s.OutType)
+		}
+	case "Probability":
+		switch s.OutType {
+		case "Frequency":
+			log.Fatalf("Error: Cannot convert a position probability matrix to a position frequency matrix.")
```
hcnapier, line 82: It might be useful to add a little documentation about which formats each format can be converted to. It’s not difficult to figure out, so maybe not necessary, but may be nice to summarize somewhere

cmd/pwmTools/shuffle.go
```diff
@@ -0,0 +1,86 @@
...
+		"pwmTools shuffle in.pwm out.pwm\n" +
+		"options:\n")
+	shuffleFlags.PrintDefaults()
+}
+
+func parseShuffleArgs() {
+	var expectedNumArgs int = 2
+	var err error
+	shuffleFlags := flag.NewFlagSet("shuffle", flag.ExitOnError)
+	var numShuffle *int = shuffleFlags.Int("numShuffle", 10, "Specify the number of desired shuffle motifs per input motif.")
```
hcnapier, line 37: Did you mean to write "Specify the number of desired **shuffled** motifs..." instead?

cfauci:
i think the only thing left is that the commands need their groups and the documentation so it will load on the godocs page

rimangan:
> i think the only thing left is that the commands need their groups and the documentation so it will load on the godocs page

Done in latest commit, thanks

cfauci (approved):

PR #953, opened 2023-11-22
adding bed name filter to bedFilter

cfauci:
# Description
added a matching and non matching name filter to bedFilter, tests passing


hcnapier (approved):
Looks good to me!

PR #954, opened 2023-11-28
add reorder option to bedFormat

sweavs111:
Adds a new option to bedFormat.go that will reorder the columns if chromStart > chromEnd.

Example:

chrA 20 10  --> chrA 10 20

Added testing.


sweavs111:
yes it is a good question i had downloaded a bed file from the internet. I was looking at it last night and it appeared to have mis-ordered columns, so i had added this quick fix. I am now re-looking at that bed file and it looks like maybe i made that problem up because the original file looks correctly formatted. I will check again and then delete the PR if it is not needed!

PR #956, opened 2023-12-01
update samToWig to handle deletions

sweavs111:
This is an update to the samToWig command. The default behavior of samToWig doesn't change, deletions in the sam alignment are "filled" in for the output wig.  I have added an option that will will not fill in over deletions. To do this i had to make a new function in convert that does sam -> bed but returns a slice of bed that corresponds to each individual alignment stretch (excluding deletion regions). All testing has been added


rimangan:
This is a really nice feature, excited to review

RavenYL (approved):

PR #959, opened 2023-12-12
add nameContains option to faFilter

RavenYL:
# Description
add nameContains option to faFilter

### Testing
added test

### DIFF
--- a/cmd/faFilter/faFilter.go
+++ b/cmd/faFilter/faFilter.go
@@ -70,7 +70,7 @@ func main() {
 	var end *int = flag.Int("end", -1, "Retains the sequence before this position.")
 	var name *string = flag.String("name", "", "Specifies the fasta record name.")
 	var notName *string = flag.String("notName", "", "Returns all fasta records except for this input.")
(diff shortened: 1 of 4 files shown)


cmd/faFilter/faFilter.go
```diff
@@ -66,6 +70,7 @@ func main() {
 	var end *int = flag.Int("end", -1, "Retains the sequence before this position.")
 	var name *string = flag.String("name", "", "Specifies the fasta record name.")
 	var notName *string = flag.String("notName", "", "Returns all fasta records except for this input.")
+	var nameContains *string = flag.String("nameContains", "", "Returns all fasta records whose name contains this input.")
```
sweavs111, line 73: maybe include in the usage if the string matching has to be case sensitive or not?

sweavs111 (commented):
looks good, just one comment about the usage statement

cmd/faFilter/faFilter.go
RavenYL: Just updated! Thanks!

sweavs111 (approved):
looks great!

rimangan (approved):

PR #960, opened 2023-12-13
bedSplit

rimangan:
# Description
This is a command I've implemented in python or awk a few times over the years, decided I need it often enough that it might be a nice program to write. Analogous to the kentUtils faSplit, this program splits an input bed file into a directory of bed files, split by either the name field or the chromosome.

### Testing

None


cmd/bedSplit/bedSplit.go
```diff
@@ -0,0 +1,103 @@
...
+		err = os.Mkdir(s.OutDir, 0755)
+		exception.PanicOnErr(err)
+	}
+
+	records := bed.GoReadToChan(s.InFile)
+	switch s.Mode {
+	case "byName":
+		var seenNames = make(map[string]*fileio.EasyWriter, 0)
+		for v := range records {
+			v.Name = strings.Replace(v.Name, "/", "_", -1)
```
RavenYL, line 34: Is the replacement of "/" with _ in bed name fields because output file names with "/" would mess up the path?
I think it is possible to retain "/" in the bed files, but just change "/" to _ in file names, but it may not be a necessary revision!

RavenYL (approved):

cmd/bedSplit/bedSplit.go
rimangan: Yes this will disrupt paths. I went back and forth stylistically on whether to update "/" in the bed files, but I think this may be programmatically cleaner and more interpretable for the user. Hopefully a rare corner case as well!

cfauci (approved):

PR #961, opened 2023-12-15
Fractional Matrix Exponentiation

rimangan:
# Description
Hi all, this PR moves our matrix tools to a dedicated matrix subpackage of numbers, and introduces a function to raise a symmetric matrix to a fractional power. I also introduce a new dependency for the gonum matrix package, which has really good high-performance linear algebra code.

Consider a square, symmetric matrix A. We can easily define A^2 as simply AA, with standard matrix multiplication. However, I've run into situations where I need something analogous to A^(1.5), or sqrt(A). This can be solved with the identity A^t = exp(t*log(A)) for any fractional power.

Now that expression has a few weird components. First, we need to evaluate the logarithm of a matrix, then we raise the number e to the value of a matrix. What does it mean for e to be raised to a matrix power? Check out the 3blue1brown video below for a fun explanation of matrix exponentials. 

For calculating a matrix logarithm, we can use another identity, right from wikipedia: log(A) = V*log(D')*V^{-1}.
Here V is the right eigenvector matrix, and log(D') is a diagonal matrix containing all of the element-wise logarithms of the eigenvalues of A. We can evaluate this through eigen decomposition of A, which can all be done using the gonum matrix library.

## Relevant Links
[Logarithm of a Matrix](https://en.wikipedia.org/wiki/Logarithm_of_a_matrix) 
[3blue1brown on matrix exponentials](https://www.youtube.com/watch?v=O85OWBJ2ayo) 
[GoDocs](https://pkg.go.dev/github.com/vertgenlab/gonomics)

### Testing

None


cfauci (approved):

cfauci (approved):

RavenYL (approved):
Noted that this will be part of the reconstructSeq rewrite

PR #962, opened 2023-12-15
Update bedToWig.go usage

RavenYL:
I noticed a small difference between cmd name and usage description

# Description
🐛 Bug Report


rimangan (approved):
good find!

craiglowe (approved):
Good catch!

PR #963, opened 2023-12-15
Overlap Threshold

cfauci:
# Description
interval overlap is now capable of examining the amount of overlap between elements and passing them only when the threshold is met

### Testing
testing passes


rimangan (dismissed):
Great feature!

cmd/intervalOverlap/intervalOverlap.go
```diff
@@ -58,7 +58,7 @@ func main() {
 	//var extend *int = flag.Int("extend", 0, "WIP") //TODO
 	var nonOverlap *bool = flag.Bool("nonOverlap", false, "Return records that do NOT have any overlap with records in the select file")
 	var threads *int = flag.Int("threads", 1, "Number of threads to use.")
-	//var percentOverlap *float64 = flag.Float64("percentOverlap", 0, "WIP") //TODO
+	var thresholdOverlap *float64 = flag.Float64("thresholdOverlap", 0, "Returns regions that have a minimum percentage of overlap between select and infile regions.")
```
RavenYL, line 61: Would it help users to have a check that thresholdOverlap is between 0 and 1, and specify in the usage that we need a fraction (as opposed to percentage)?

dasnellings (commented):
This if-else logic is starting to get a bit hard to follow. Not sure just looking at it how best to improve. I am going to play around with it a bit and make a commit. If you don't like it, feel free to revert.

dasnellings:
I pushed a commit that doesn't change any logic, but tries to simplify it a bit. If you don't approve or would prefer it another way feel free to revert/edit.

dasnellings:
Also it seems like thresholdOverlap is incompatible with merged output. Would it be possible to go through the answer slice and return the elements that pass the threshold instead of just returning an dummy slice?

cfauci (commented):
I like the new syntax it was definitely getting a bit overwhelming in the previous version!

cfauci:
> Also it seems like thresholdOverlap is incompatible with merged output. Would it be possible to go through the answer slice and return the elements that pass the threshold instead of just returning an dummy slice? 

I tried to do exactly that and it gave me a bunch of problems that i just didn't have the time to fix. Not really sure why it was happening but the new func may help, I'll try to reimplement.

cmd/intervalOverlap/intervalQuery.go
```diff
@@ -38,32 +37,62 @@ func buildTree(intervals []interval.Interval, aggregate bool) map[string]*interv
...
+
 	if mergedOutput {
+		return answer
+	}
+
+	// else no merged output
+	return make([]interval.Interval, 1)
+}
+
+func passingThreshold(query interval.Interval, answer []interval.Interval, thresholdOverlap float64) []interval.Interval {
```
craiglowe, line 79: I think documentation for this function would help reviewers know what it should be doing

cmd/intervalOverlap/intervalQuery.go
```diff
@@ -38,32 +37,62 @@ func buildTree(intervals []interval.Interval, aggregate bool) map[string]*interv
...
+	if len(answer) == 0 || thresholdOverlap == 0 {
+		return answer
+	}
+	var ovSize, intSize float64
+	for a := range answer {
+		ovSize = float64(interval.OverlapSize(answer[a], query))
+		intSize = float64(interval.IntervalSize(query))
+		if ovSize/intSize >= thresholdOverlap {
+			return make([]interval.Interval, 1)
```
craiglowe, line 88: I am not sure why we are sending back an slice of intervals with length 1, which I think contains default values for the interval?

craiglowe (commented):
Great to see work going into improving this program!  I had a few comments

cmd/intervalOverlap/intervalOverlap.go
cfauci: this is a one to one comparison not a total overlap, That was purposeful. Is there something in the language that you think is unclear or could be specified?

cmd/intervalOverlap/intervalOverlap_test.go
```diff
@@ -11,46 +11,72 @@ import (
...
+		SelectFile:       "testdata/test2.bed",
+		NonOverlap:       false,
+		Threads:          1,
+		Aggregate:        false,
+		Relationship:     "any",
+		MergedOutput:     false,
+		ThresholdOverlap: 0.50,
+		ExpectedFile:     "testdata/expected.threshold.bed",
 	},
+	//{InFile: "testdata/test.bed",
```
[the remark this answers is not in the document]
cfauci: This was a test I wrote when I thought i could make this compatible with mergedOutput which had been giving me trouble. I'm going to look through dan's update and hten remove it if that's whats best

cfauci:
I want to make sure to prioritize this PR in the next code review because something about it isn't sitting right and I'm not totally sure where it's coming from. Everything works fine so long as thresholdOverlap isn't used with MergedOutput so for now we will keep that intact.

edotau (approved):

PR #965, opened 2023-12-31
Reconstruct Updates Part 2: Ancestral Sequence Reconstruction with Substitution Matrices

rimangan:
# Description
This PR updates reconstructSeq to be compatible with more complex mutation models, as specified by a 4x4 substitution matrix. This builds upon code like our forward evolutionary model and matrix tools in the preceding PRs. While this was a lot of effort, unfortunately, I have conducted performance simulations and find no conclusive evidence that the adoption of more complex substitution models improves reconstruction accuracy. Thus, this may be an improvement in theory, though not in practice. 

To benchmark, I used simulate.ETree to make random phylogenies, and simulate.RandomIntergenicSeq to produce a sequence for the root node. Then I simulate forward evolution along the tree using the NonCoding program in the last PR. The forward model was either "JukesSim", for Jukes-Cantor simulation, "Gtr" for general time reversible (the more complex mutation model that has a 3 fold transition bias), or "TransitionSim", where transversions are nearly impossible (an extreme model to test some corner cases). Then I reconstructed using a variety of mutation models for reconstruction. Inaccuracy for each sequence across all nodes is reported.
image

### Testing

None


reconstruct/reconstruct_test.go
```diff
@@ -91,3 +94,223 @@ func TestReconstruct(t *testing.T) {
 		fileio.EasyRemove(v.SimTree)
 	}
 }
+
+// this benchmark takes about 60s CPU to run, so I've commented it out. Go ahead
+// and uncomment if you're interested in reproducing it or altering it. -RJM
+/*
+var EmpiricalReconstructionComparison = []struct {
```
saqzhao, line 101: Should this testcase be commented out?

saqzhao (commented):
Looks good! I just had a question about the EmpiricalReconstructionComparison testcase.

saqzhao (approved):

cfauci (approved):

PR #966, opened 2024-01-07
gtfTools Part I: Moving gtfToBed

rimangan:
# Description
We've talked a few times about the need for tools to work with gtf files, such as a filter function. Here I start this process by making a hierarchical command called gtfTools, and moving the existing 'gtfToBed' cmd to a subcommand of gtfTools called 'toBed'. The tests are the exact same as in 'gtfToBed', and still pass with this move. I will add a part 2 to this PR to introduce 'gtfTools filter' sometime soon.

### Testing

None


cmd/gtfTools/toBed.go
```diff
@@ -0,0 +1,101 @@
...
+	if s.Tss {
+		records := gtf.Read(s.InFile)
+		sizes := chromInfo.ReadToMap(s.ChromSizeFile)
+		beds := gtf.GenesToTssBed(records, sizes, s.Merge)
+		bed.Write(s.OutFile, beds)
+	} else {
+		file := fileio.EasyOpen(s.InFile)
+		out := fileio.EasyCreate(s.OutFile)
+		for line, doneReading = fileio.EasyNextRealLine(file); !doneReading; line, doneReading = fileio.EasyNextRealLine(file) {
+			words := strings.Split(line, "\t")
```
RavenYL, line 85: When converting between file types, is it usually better to do strings.Split as opposed to first parsing into one struct (a gtf variable) and then indexing that struct to convert to another struct (a bed variable)? Or maybe it's better specifically in the case of gtfs?

RavenYL (approved):

cfauci (approved):

PR #967, opened 2024-01-07
gtfTools Part II: Filter

rimangan:
# Description
This PR introduces the gtfTools filter subcommand, which can be used for filtering gtf records. Currently, this supports the option of a new-line delimited list of gene names to be retained. 

## Relevant Links
- [GtfFileFormatDescription](https://useast.ensembl.org/info/website/upload/gff.html)
- [GoDocs](https://pkg.go.dev/github.com/vertgenlab/gonomics)

### Testing

None


cmd/gtfTools/filter.go
```diff
@@ -0,0 +1,84 @@
...
+	records := gtf.Read(s.InFile)
+	for currGene := range records {
+		pass = true
+		if s.GeneNameList != "" {
+			if _, foundInMap = geneNameMap[records[currGene].GeneName]; !foundInMap {
+				pass = false
+			}
+		}
+		if pass {
+			gtf.WriteToFileHandle(out, records[currGene])
```
RavenYL, line 78: Note to self: records[currGene] or just currGene? records[currGene] is correct.
records is the output of gtf.Read so the type is map[string]*Gene.
currGene is the variable that ranges through records, and represents the key of the map.
This is why records[currGene] is correct.

RavenYL (approved):

hcnapier (approved):
Looks good!

PR #968, opened 2024-01-15
SamPE struct

sweavs111:
Paired End Sam struct and channel reading functions.

currently the channel reading function reads a sam channel as normal and then from that channel parses out paired end reads and sends them to a new channel. I did it this way because there is already a lot of sam-reading infrastructure built in, but i could imagine doing it a different way. Opinions?


sam/pairedSam.go
```diff
@@ -1 +1,50 @@
...
+		done = false
+		a, full = <-data
+		b, full = <-data
+		for !done {
+			switch {
+			case full, a.QName == b.QName:
+				peChan <- SamPE{a, b}
+				done = true
+			case a.QName != b.QName:
+				a = b
```
RavenYL, line 44: Note to self: input is a read-name sorted sam file, so if a and b names are not equal, can proceed and set a name to b name (aka move on to the next name in the file)

RavenYL:
The supplementary alignments are interesting! How do you tell them apart from primary alignments? I like the additional field, and can imagine that for some output formats you would omit it in the final write-to-output-file stage

sweavs111:
Yes good question. Supplementary alignments have the flag 2048. I just wrote a version 2 of the implements this logic if you want to check it out.

sam/pairedSam.go
```diff
@@ -1 +1,100 @@
...
+func OrderSamPair(p SamPE) (SamPE, bool) {
+	if p.A.Flag&64 == 64 && p.B.Flag&128 == 128 {
+		return p, true
+	} else if p.A.Flag&128 == 128 && p.B.Flag&64 == 64 {
+		return SamPE{A: p.B, B: p.A, Sup: p.Sup}, true
+	}
+	return p, false
+}
+
+/*		done = false
```
sweavs111, line 88: I kept the v1 code here commented out just in case we decide v1 is better

sam/pairedSam.go
```diff
@@ -1 +1,122 @@
...
+		} else {
+			peChan <- parseReads(tmpSlice)
+			tmpSlice = []Sam{tmp}
+		}
+	}
+	close(peChan)
+}
+
+// parseReads is a helper function for GoReadSamPeToChan that parses a slice of Sam that all have the same read name into a SamPE struct
+func parseReads(tmpSlice []Sam) SamPE {
```
dasnellings, line 62: What if the len of tmpSlice is >2? Doesn't this end up tossing reads?

sam/pairedSam.go
```diff
@@ -1 +1,122 @@
...
+	B   Sam
+	Sup []Sam
+}
+
+// GoReadSamPeToChan takes a read-name sorted sam file and returns a chanel of SamPE and a Header
+func GoReadSamPeToChan(filename string) (<-chan SamPE, Header) {
+
+	//start by reading the file into a chanel of sam
+	data := make(chan Sam, 1000)
+	header := make(chan Header)
```
edotau, line 23: When running this function that takes a single file as an input, would there ever be an instance where you would have more header?

Probably doesn't matter too much, but I was just wondering if something like
`var header Header` would be enough. Does `header := make(chan Header)` only make 1 goroutine? I forgot how that stuff works.

sweavs111: Yeah its a good question. I did it that way because thats how it was done in the other sam readToChan functions. However, I also thought it was weird that make a channel for just the header. I look into this more tomorrow

edotau (dismissed):

dasnellings (approved):

edotau (approved):

PR #970, opened 2024-01-18
Wig Package: piecewise wig outputs

rimangan:
# Description
Currently, when we write wigs, we write the whole chromosome as one wig entry, with one line per value, even if there is a “missing” value at that position. By contrast, in most packages, a new Wig entry would be spawned at each position where there is data. Consider these two wigs, both representing the same chromosome. Here’s the old version:
```
fixedStep chrom=chrA start=1 step=1
2
3
3
3
4
2
4
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
10
11
12
11
10
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
0
```
And here’s what I’d like this to look like in the update:
```
fixedStep chrom=chrA start=1 step=1
2
3
3
3
4
2
4
fixedStep chrom=chrA start=30 step=1
10
11
12
11
10
```
This new PR implements this change and makes all changes to commands to retain stability testing.

Two other notable changes: 
1. Changed '%f' to '%g' in wig.WriteToFileHandle, which will make cleaner, small outputs, and help with output file sizes.
2. Added the DefaultValue as a new Wig struct field. We're often passing this value around together with wigs, and we need it for writing in this new format, so I've decided it belongs in the struct.

Note to reviewers, the function wig.WriteToFileHandle is the only function with serious updates, everything else is minor updates to testing files affected by this package change. 

## Relevant Links
- [WIG File Format Official Documentation](https://genome.ucsc.edu/goldenPath/help/wiggle.html)
- [GoDocs](https://pkg.go.dev/github.com/vertgenlab/gonomics) 


nikithat23 (approved):
This update to the formatting of the wig seems like a smart edit and looks good to me.

wig/wig.go
```diff
@@ -125,32 +126,40 @@ func Write(filename string, records map[string]Wig) {
...
-	if rec.StepType == "fixedStep" {
-		if rec.Span != -1 { //if there was a span in the header line
-			_, err = fmt.Fprintf(file, "%s chrom=%s start=%d step=%d span=%d\n", rec.StepType, rec.Chrom,
-				rec.Start, rec.Step, rec.Span)
+	var inDefaultRegion bool = true // bool to detect when we are in a region of defaultValues, where we don't need to write to file
+	for currIndex := range rec.Values {
+		if rec.Values[currIndex] != rec.DefaultValue {
+			if inDefaultRegion {
+				inDefaultRegion = false
+				_, err = fmt.Fprintf(file, "%s\n", makeWigHeaderString(rec, rec.Start+currIndex-1))
```
RavenYL, line 134: makeWigHeaderString(rec, rec.Start+currIndex-1) calls makeWigHeaderString(record Wig, currIndex int) and prints a header line where start=currIndex+1. Would it be simpler to call makeWigHeaderString(rec, rec.Start+currIndex) and print start=currIndex?

cmd/bedToWig/bedToWig.go
```diff
@@ -53,7 +54,7 @@ func usage() {
 
 func main() {
 	var expectedNumArgs int = 4
-	var missing *float64 = flag.Float64("missingData", 0, "For BedNameToWig, sets the value of the output wig in regions where there is no bed data.")
+	var defaultValue *float64 = flag.Float64("defaultValue", math.MaxFloat64, "(Developer) Set the default value of the wig struct. Set to a value that is unlikely to be found in the real wig data.")
```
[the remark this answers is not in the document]
rimangan: the max value (about 1e300) seems equally unrealistic. I don't believe there is a math.MinFloat64, but I could use -1 * math.MaxFloat64 if you'd prefer

wig/wig.go
rimangan: I thought this was more clear in that it makes it explicit we're converting between a 0-based and 1-based system. If I went with the solution you're describing, I would want to change the variable name away from 'currIndex', as to me 'index' implies 0 based, so start = currIndex looks a bit confusing.

dasnellings (approved):

PR #971, opened 2024-01-19
Out of Range Fix in simulateSam

nikithat23:
Fixed out of range issue with deaminationDistributionSlice where the distancetoEnd is longer than the slice. Compared length of slice to distancetoEnd and only added data point to deamination distribution if distancetoEnd < len(deaminationDistributionSlice).

### DIFF
--- a/simulate/sam.go
+++ b/simulate/sam.go
@@ -240,10 +240,14 @@ func ancientDamage(currFrag []dna.Base, ancientAlias numbers.BinomialAlias, geom
 				// nothing to do
 			case dna.C:
 				currFrag[currRandPos] = dna.T
-				deaminationDistributionSlice[distanceToEnd]++
+				if distanceToEnd < len(deaminationDistributionSlice) {
+					deaminationDistributionSlice[distanceToEnd]++
+				}
 			case dna.G:
 				currFrag[currRandPos] = dna.A
-				deaminationDistributionSlice[distanceToEnd]++
+				if distanceToEnd < len(deaminationDistributionSlice) {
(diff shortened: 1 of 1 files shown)


rimangan (approved):
Great catch!

simulate/sam.go
```diff
@@ -240,10 +240,14 @@ func ancientDamage(currFrag []dna.Base, ancientAlias numbers.BinomialAlias, geom
 				// nothing to do
 			case dna.C:
 				currFrag[currRandPos] = dna.T
-				deaminationDistributionSlice[distanceToEnd]++
+				if distanceToEnd < len(deaminationDistributionSlice) {
```
craiglowe, line 243: Right now it seems like we don't record the deamination event if it is further from the end than the length of the slice.  Could this make debugging hard if we are not recording all the events that happen to the sequence?  Should we instead increase the length of the slice?

craiglowe (commented):
Great to catch this out of range issue!  I made a quick comment about another possible solution.

simulate/sam.go
rimangan: This debug slice was just used for plotting the locations of the deamination events, which is geometrically distributed. Since the geometric distribution is supported over positive integers, technically we could get arbitrarily high numbers (>100 even). However, the probability becomes increasingly rare, even at 40 under our conditions. So we'll need to truncate this distribution at some point. Since this value represents the expected read size for the neanderthal reads, I thought this was as good a place as any.

oliviahan02 (approved):

dasnellings (approved):

PR #975, opened 2024-01-28
MultiFa Bed Sequence Swap

jayashabarishankar:
# Description
I worked on this sequence swap program and its corresponding test file with Riley.  SubsequenceSwap swaps a series of specified regions from a bed file between two sequences in a MultiFa object. The sequence called "background" is taken, and parts of "foreground" are swapped with the corresponding parts in background. The input of this function includes the multiFa file name, the index of the background and foreground sequences from the multiFa, a bed object with the relevant start and stop indices of the swap regions, and the name of the output sequence (the sequence with all swaps). It will return a file containing a Fasta slice (which includes several sequences), with one more sequence that the sequences in the original multiFa; this extra sequence is the background sequence with relevant foreground spots swapped as dictated by the bed file.

### Testing
A test file is also provided and all test data is in the testdata folder.

- I performed a self-review of my code.
-  If it this a core feature, I added thorough tests.

For the bed file and multiFa files, make sure they end with one extra line (hit Enter on the last line)- it is a strange formatting issue for file formats (maybe this is only the case on Windows?)


cmd/multiFaSequenceSwap/multiFaBedSequenceSwap.go
```diff
@@ -0,0 +1,89 @@
+package main
```
rimangan, line 1: You'll notice the gitactions failed for this PR. This is because gonomics expects the directory name for commands to be equal to the command file name. In this case, there is a 'Bed' in the filename, but not in the directory name. The right fix here is to rename these files to toss the 'Bed' part. Then we can recommit and see if it passes tests

cmd/multiFaSequenceSwap/multiFaBedSequenceSwap.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/fileio"
+	"github.com/vertgenlab/gonomics/numbers/parse"
+	"log"
+)
+
+// SubsequenceSwap swaps a series of specified regions from a bed file between two sequences in a MultiFa object
+// The sequence called "background" is taken, and parts of "foreground" are swapped with the corresponding parts in background
```
rimangan, line 14: Let's clarify the language a bit. The key detail to emphasis is that the bed regions are the foreground, and everything outside the beds is the background

cmd/multiFaSequenceSwap/multiFaBedSequenceSwap.go
```diff
@@ -0,0 +1,89 @@
...
+// input of this function includes the multiFa file name, the index of the background and foreground sequences from the multiFa, and a bed object with the relevant
+// start and stop indices of the swap regions
+// it will return a file containing a Fasta slice (which includes several sequences), with one more sequence that the sequences in the original
+// multiFa; this extra sequence is the background sequence with relevant foreground spots swapped as dictated by the bed file.
+func SubsequenceSwap(inFile string, swapRegionsFile string, background, foreground int, outFile string, outSeqName string) {
+	var currStart, currEnd, currPos int //these variables will be used to swap regions later in the code
+
+	// Load the original sequences from the multiFa file.
+	originalSeqs := fasta.Read(inFile) //Converting a multiFa to a Fasta file
+	lines := fileio.Read(swapRegionsFile)
```
rimangan, line 24: Let's cut lines 24 and 25, I believe we had these for debugging

cmd/multiFaSequenceSwap/multiFaBedSequenceSwap.go
```diff
@@ -0,0 +1,89 @@
...
+			answerSeq.Seq[currPos] = originalSeqs[foreground].Seq[currPos]
+		}
+	}
+	var finalSeqs []fasta.Fasta
+	answerSeq.Name = outSeqName
+	finalSeqs = append(originalSeqs, answerSeq)
+
+	fasta.Write(outFile, finalSeqs)
+
+	// Create and return the new Fasta with one more sequence than the input.
```
rimangan, line 55: you can remove this comment, it describes outdated functionality

cmd/multiFaSequenceSwap/multiFaBedSequenceSwap.go
```diff
@@ -0,0 +1,89 @@
...
+}
+func usage() {
+	fmt.Print(
+		"multiFaSequenceSwap - Swaps regions of a sequence with regions from a different corresponding sequence. \n" +
+			"Usage: \n" +
```
rimangan, line 60: Let's add a usage statement. in this case, it would look something like this:
multiFaSequenceSwap inFile.fa regions.bed backgroundIdx foregroundIdx outFile outSeqName

rimangan: You can use another command, like multFaVisualizer, as an example of how to write usage statements in main functions

rimangan (commented):
Great start! My comments are minor, and should be quick to address

cmd/multiFaSequenceSwap/multiFaBedSequenceSwap.go
```diff
@@ -0,0 +1,89 @@
...
+func main() {
+
+	var expectedNumArgs int = 5
+	flag.Usage = usage
+	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	flag.Parse()
+
+	if len(flag.Args()) != expectedNumArgs {
+		flag.Usage()
+		log.Fatalf("Error: expecting %d arguments, but got %d/n",
```
rimangan, line 76: Looks like a /n instead of a newline character '\n'

PR #976, opened 2024-01-29
MultiFa Sequence Swap 

jayashabarishankar:
# Description
New pull request replacing #975 because this is a branch off of main instead of a branch off of fork.

Description
I worked on this sequence swap program and its corresponding test file with Riley. SubsequenceSwap swaps a series of specified regions from a bed file between two sequences in a MultiFa object. The sequence called "background" is taken, and parts of "foreground" are swapped with the corresponding parts in background. The input of this function includes the multiFa file name, the names of the background and foreground sequences from the multiFa, a bed object with the relevant start and stop indices of the swap regions, and the name of the output sequence (the sequence with all swaps). It will return a file containing a Fasta slice (which includes several sequences), with one more sequence that the sequences in the original multiFa; this extra sequence is the background sequence with relevant foreground spots swapped as dictated by the bed file.

Testing
A test file is also provided and all test data is in the testdata folder.


rimangan (approved):

cmd/multiFaSequenceSwap/multiFaSequenceSwap.go
```diff
@@ -0,0 +1,103 @@
...
+	"log"
+)
+
+// MultiFaSubsequenceSwap swaps a series of specified regions from a bed file between two sequences in a MultiFa object
+// The foregroundName is the name of the sequence containing the regions in the bed elements;  backgroundName is the name of the sequence including the regions outside the bed elements
+// input of this function includes the multiFa file name, the index of the background and foreground sequences from the multiFa, and a bed object with the relevant
+// start and stop indices of the swap regions
+// it will return a file containing a Fasta slice (which includes several sequences), with one more sequence that the sequences in the original
+// multiFa; this extra sequence is the background sequence with relevant foreground spots swapped as dictated by the bed file.
+func MultiFaSubsequenceSwap(inFile string, swapRegionsFile string, backgroundName, foregroundName string, outFile string, outSeqName string) {
```
sweavs111, line 17: Nice PR! In general function names in command shouldn't be capitalized. Capitilized functions are public and accessible in other files, which is not necessary for cmd functions.

cmd/multiFaSequenceSwap/multiFaSequenceSwap.go
```diff
@@ -0,0 +1,103 @@
...
+		currEnd = swapRegion.ChromEnd
+
+		// Validate the swap region.
+		if currStart < 0 || currEnd > len(answerSeq.Seq) || currStart >= currEnd {
+			log.Fatalf("Error: Invalid swap region. \n")
+
+		}
+
+		// Swap the corresponding portion of answerSeq with the foreground sequence.
+		for currPos = currStart; currPos < currEnd; currPos++ {
```
craiglowe, line 41: I think this math does not account for gaps in the reference genome, which would be needed for the bed coordinates to always correspond to the same base in the reference genome.  It looks like right now, the bed region is being used as alignment position instead of coordinates in the reference genome.

cmd/multiFaSequenceSwap/multiFaSequenceSwap.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func usage() {
+	fmt.Print(
+		"multiFaSequenceSwap - Swaps regions of a sequence with regions from a different corresponding sequence. \n" +
+			"foregroundName is the name of the sequence that includes the regions in the bed elements.\n" +
+			"backgroundName is the name of the sequence that includes the regions outside the bed elements.\n" +
+			"Input of this function includes the multiFa file name,\n" +
+			"the index of the background and foreground sequences from the multiFa,\n" +
+			"a bed object with the relevant start and stop indices of the swap regions,\n" +
```
craiglowe, line 70: I think it is a little unclear what the bed file represents.  I think most people will think it is for coordinates in the reference/first genome in the alignment, but I think this uses the bed coordinates as alignment positions (and ignores the name).  Maybe the code should be adjusted to have the bed represent reference coordinates, or maybe bed is not the best fit for the input format?

craiglowe (commented):
Great work!  I had a comment about alignment coordinates vs reference coordinates

rimangan (changes requested):
We'll need to fix the refCoords/alnCoords problem Craig mentioned

cmd/multiFaSequenceSwap/multiFaSequenceSwap.go
rimangan: Hi Craig, thanks for looking at this! I just pushed a commit that changes the program to handle the bed as reference positions, not as alignment positions

rimangan (approved):

oliviahan02 (approved):

PR #977, opened 2024-01-30
GC calculator

hcnapier:
# Description

I added a little GC content calculator to examine.go. Pretty simple, the only thing to note is that I ignore all Ns in the sequence. I've seen some GC calculators that randomly choose a base, but that doesn't fit my use-case, so I figure we can add it later if that functionality becomes something people want.

🐛 Bug Report

### Testing

None

### DIFF
--- a/dna/examine.go
+++ b/dna/examine.go
@@ -55,7 +55,7 @@ func CountGaps(seq []Base) int {
 	return gapCount
 }
 
-// GCContent returns the GC content for the input sequence. Note that n/Ns are ignored.
+// GCContent returns the GC content for the input sequence. Note that n/Ns are ignored and output is truncated to the nearest int.
 func GCContent(seq []Base) (gcContent int) {
 	ACount, CCount, GCount, TCount, _, aCount, cCount, gCount, tCount, _, _ := Count(seq)
 	gcContent = ((CCount + GCount + cCount + gCount) * 100) / (ACount + CCount + GCount + TCount + aCount + cCount + gCount + tCount)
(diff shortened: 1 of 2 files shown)


dna/examine.go
```diff
@@ -55,6 +55,13 @@ func CountGaps(seq []Base) int {
 	return gapCount
 }
 
+// GCContent returns the GC content for the input sequence. Note that n/Ns are ignored.
```
sweavs111, line 58: maybe write something in the comment that it truncates to nearest int

sweavs111 (approved):

dna/examine.go
hcnapier: Good idea

dna/examine.go
```diff
@@ -55,6 +55,23 @@ func CountGaps(seq []Base) int {
...
+func GCContent(seq []Base) (gcContent float32) {
+	ACount, CCount, GCount, TCount, _, aCount, cCount, gCount, tCount, _, _ := Count(seq)
+	// Convert count output from int to float
```
hcnapier, line 61: I've added the following lines to convert the int output of Count to float32

sweavs111 (approved):
looks great!

edotau (approved):

dasnellings (approved):

dna/examine.go
dasnellings: You should make a note that this returns a percent rather than a fractional value.

cfauci (approved):

PR #978, opened 2024-01-30
GC filter

hcnapier:
# Description
Another very small PR. This one adds an option to faFilter to filter records based on a max GC content input.

### Testing

None


dna/examine_test.go
```diff
@@ -38,6 +38,12 @@ func TestCountGaps(t *testing.T) {
 	}
 }
 
+func TestGCContent(t *testing.T) {
+	if GCContent(alpha) != 50 {
+		t.Errorf("problem with GCContent")
```
edotau, line 43: Small style suggestion/preference to stay consistent with repository conventions
```suggestion
		t.Errorf("Error: Problem with GCContent")
```

cfauci (approved):

sweavs111 (approved):
everything looks good to me, so i am approving! Like Craig mentioned yesterday, you calculate GC% twice which could be slight optimization, but that can be a separate PR if you want to make that optimization

PR #981, opened 2024-02-01
SimulateEvol: Re-adding missing options

rimangan:
# Description
Small bug fix. SimulateEvol previously had options for transversion/transition bias, writing out an optional VCF file, and specifying the query name. However, these options were removed from the main function by mistake in a recent update. All this PR does is add those back. Note that all the underlying functionality for using these options was already intact. 

### Testing

None


edotau (approved):

nikithat23 (approved):
LGTM

PR #982, opened 2024-02-01
samAssembler Part 10: Updated Prior with Empirical Bayes approach

nikithat23:
Instead of a flat error rate prior being used for the samAssembler pipeline, these edits produce an empirical Bayesian prior from observed amounts of sequencing errors and ancient DNA damage rates from pileups. A weighted average is taken of the sequencing error rate (epsilon), which is then subtracted from the weighted average of the total error rate (epsilon + lambda) to get lambda, the rate of cytosine deamination. Finally, the expected output files for the prior tests were updated to include two header lines reporting the epsilon and lambda values. 

This explains the formal mathematical treatment of the problem.
[Joint_Error_Rate_Estimation_with_SamAssembler_Prior.pdf](https://github.com/vertgenlab/gonomics/files/14144854/Joint_Error_Rate_Estimation_with_SamAssembler_Prior.pdf)


edotau (approved):

rimangan (approved):

rimangan:
I've added a few commits on this. First, I've updated build to be able to accept the empirical priors that Nikitha is generating. Second, I've found a small bug in build, where the order in which we increment refPos can matter in the case of complex INDELs. For example, samAssembler on main breaks when it encounters deletions within deletions, like so:
```
>Ref
AACCGGTT
>Alt1
AA----TT
>Alt2
AAC--GTT
```

rimangan (approved):

dasnellings (approved):

edotau (approved):

PR #983, opened 2024-02-02
Three way fa to vcf

oliviahan02:
# Description
WIP: takes in three-way multiFa alignment and generates a corresponding VCF format. currently only handles substitutions; work in progress for indels.

### Testing

None


cmd/multiFaToVcf/multiFaToVcf.go
```diff
@@ -1,17 +1,18 @@
...
+
 	"github.com/vertgenlab/gonomics/convert"
 	"github.com/vertgenlab/gonomics/exception"
 	"github.com/vertgenlab/gonomics/fasta"
 	"github.com/vertgenlab/gonomics/fileio"
 	"github.com/vertgenlab/gonomics/vcf"
-	"log"
 )
 
 func multiFaToVcf(inFile string, chr string, outFile string, substitutionsOnly bool, retainN bool) {
```
rimangan, line 18: On line 29 in the usage, we'll want to edit the usage message to explain to the user how they can use a pairwise or 3-way FASTA for this program. we'll need to explicitly inform the user that the 3-way is substitutions only. (We could maybe suggest a TODO for some brave person who wants to tackle that problem in the future)

rimangan, line 18: In the main function of this program, we'll need to check whether the input FASTA has 2 or 3 entries and then send it to the corresponding function in convert.go

rimangan: Line 22/23 of this file multiFaToVcf.go is where we'll need to check the length of the fasta (variable is called f) and send it to the right convert function (Pairwise or ThreeWay)

convert/convert.go
```diff
@@ -194,6 +196,78 @@ func bedMidpoint(b bed.Bed) int {
 	return (b.ChromEnd + b.ChromStart) / 2
 }
```
rimangan, line 198: Add a large comment above this function explaining how it works (see some of the other functions in this file as a template). Note that the comment conventionally starts with a space and then the function name.

convert/convert.go
```diff
@@ -194,6 +196,78 @@ func bedMidpoint(b bed.Bed) int {
...
 
+func ThreeWayFaToVcf(f []fasta.Fasta, chr string, out *fileio.EasyWriter, substitutionOnly bool, retainN bool) {
+	var currRefPos, currAlnPos int = 0, 0
+	// var insertion1, insertion2, deletion1, deletion2 bool = false, false, false, false
```
rimangan, line 201: If we're committing to only handling substitutions, you can delete the commented out lines related to INDEL logic

convert/convert_test.go
```diff
@@ -69,6 +70,14 @@ func TestPairwiseFaToVcf(t *testing.T) { //this test is for the default settings
 	}
 }
 
+func TestThreeWayFaToVcf(t *testing.T) {
+	// var err error
+	out := fileio.EasyCreate("tmp.threeWayFaToVcf.txt")
+	threewayInputFa := fasta.Read("testdata/threeWayFaToVcf/input.fa")
+	ThreeWayFaToVcf(threewayInputFa, "chr1", out, false, false)
+
```
rimangan, line 78: We'll need to finish this test so that it passes and checks functionality

cmd/multiFaToVcf/multiFaToVcf.go
```diff
@@ -1,17 +1,18 @@
...
 
-// Generates a VCF file from an input pairwise multiFa alignment with the first entry as the reference
+// Generates a VCF file from an input pairwise or three-way multiFa alignment with the first entry as the reference
 package main
 
 import (
 	"flag"
 	"fmt"
+	"log"
+
```
rimangan, line 10: Check out the file in this directory multiFaToVcf_test.go. Here we'll need to add a test case for when we have 3 sequences. You can also check out the testdata subdirectory to see how tests were set up for the pairwise case and use this as a template

vcf/gVcf.go
```diff
@@ -110,7 +110,7 @@ func sampleToString(s Sample) string {
 			answer += fmt.Sprintf("%s%d", PhasedToString(s.Phase[i]), s.Alleles[i])
 		}
 	}
-	if len(s.FormatData) > 0 {
+	if len(s.FormatData) > 1 {
```
rimangan, line 113: Glad we caught this critical bug fix!

cmd/multiFaToVcf/multiFaToVcf.go
```diff
@@ -20,14 +20,20 @@ func multiFaToVcf(inFile string, chr string, outFile string, substitutionsOnly b
 	out := fileio.EasyCreate(outFile)
 	header := vcf.NewHeader()
 	vcf.NewWriteHeader(out, header)
-	convert.PairwiseFaToVcf(f, chr, out, substitutionsOnly, retainN)
+	if len(f) == 2 {
+		convert.PairwiseFaToVcf(f, chr, out, substitutionsOnly, retainN)
+	} else if len(f) == 3 {
+		convert.ThreeWayFaToVcf(f, chr, out, substitutionsOnly, retainN)
+	}
```
rimangan, line 27: we might need an else case here if it's not 2 or 3, and the else case should lead to a log.Fatalf

cmd/multiFaToVcf/multiFaToVcf.go
```diff
@@ -20,14 +20,20 @@ func multiFaToVcf(inFile string, chr string, outFile string, substitutionsOnly b
...
 	err := out.Close()
 	exception.PanicOnErr(err)
 }
 
 func usage() {
 	fmt.Print(
-		"multiFaToVcf - Generates a VCF file from an input pairwise multiFa alignment with the first entry as the reference." +
```
rimangan, line 29: Let's at \n to the end of each of these lines. This will clean up the print statement

convert/convert.go
```diff
@@ -196,16 +196,18 @@ func bedMidpoint(b bed.Bed) int {
 	return (b.ChromEnd + b.ChromStart) / 2
 }
 
+// ThreeWayFaToVcf takes in a three-way multiFa alignment and writes Vcf entries for segregating sites with the first
+// entry as the reference and the last two fasta entries as the alt alleles.
+// This is done by chromosome since a multiFa contains only one chromosome per file.
```
rimangan, line 201: Specify that this function is substitutions only

convert/convert_test.go
```diff
@@ -70,12 +70,28 @@ func TestPairwiseFaToVcf(t *testing.T) { //this test is for the default settings
...
+	{InFile: "testdata/threeWayFaToVcf/input.fa", OutFile: "tmp.threeWayFaToVcf.txt", ExpectedFile: "testdata/threeWayFaToVcf/expected.vcf", Chrom: "chr1"},
+}
 
+func TestThreeWayFaToVcf(t *testing.T) {
+	var err error
+	for _, v := range ThreeWayFaToVcfTests {
+		out := fileio.EasyCreate(v.OutFile)
+		threewayInputFa := fasta.Read(v.InFile)
+		ThreeWayFaToVcf(threewayInputFa, v.Chrom, out, false, false)
+		if !fileio.AreEqual(v.OutFile, v.ExpectedFile) {
```
rimangan, line 88: Before the line with AreEqual, we want to add these two lines:
`err = out.Close()
exception.PanicOnErr(err)`

vcf/gVcf.go
RavenYL, line 113: Could you help me understand why this line of code is changed?

convert/convert.go
```diff
@@ -194,6 +196,81 @@ func bedMidpoint(b bed.Bed) int {
...
+			} else if f[0].Seq[i] != f[1].Seq[i] && f[0].Seq[i] != f[2].Seq[i] && f[1].Seq[i] == f[2].Seq[i] { // 1/1; substitution at both allele1 and allele2, same alt
+				altAllele = []string{dna.BaseToString(f[1].Seq[i])}
+				samples = []vcf.Sample{
+					{
+						Alleles:    []int16{1, 1},
+						Phase:      []bool{false, false},
+						FormatData: []string{"1/1"},
+					},
+				}
+			} else if f[0].Seq[i] != f[1].Seq[i] && f[0].Seq[i] != f[2].Seq[i] && f[1].Seq[i] != f[2].Seq[i] { // 1/2; substitution at both allele1 and allele2, same alt
```
RavenYL, line 246: I think this is diff alt?

convert/testdata/threeWayFaToVcf/expected.vcf
```diff
@@ -0,0 +1,3 @@
+chr1	3	.	A	C	100	PASS	.	GT	1/1
```
RavenYL, line 1: Would you want to add a test for diff alt, producing something like "A     C,G"?

vcf/gVcf.go
rimangan: Format data is a series of ':' delimited strings. if the format data is of length 1, it doesn't need to be joined by a ':', as only the genotype is in the sample field. Since the genotype is written separately earlier in this function, nothing from FormatData needs to be written out

convert/convert.go
rimangan: good catch

convert/testdata/threeWayFaToVcf/expected.vcf
rimangan: Done in next commit

RavenYL (approved):

rimangan (approved):

PR #987, opened 2024-02-06
gtfTools chromFilter

cfauci:
# Description
This allows a user to filter a gtf by chromosome

### Testing
functioning as expected


cmd/gtfTools/filter.go
```diff
@@ -73,6 +77,13 @@ func gtfFilter(s FilterSettings) {
 			if _, foundInMap = geneNameMap[records[currGene].GeneName]; !foundInMap {
 				pass = false
 			}
+		} else if s.ChromFilter != "" {
```
dasnellings, line 80: I haven't thought deeply about this, but do you think it makes the most sense to have the GeneNameList override the ChromFilter as it currently does? I suppose if you are asking for a particular gene you probably want to have it in the output even if you only asked for records specifically from a different chromosome, but I think it is less intuitive than applying both filters simultaneously. I could go either way.

cfauci: I hadn't thought about applying them together to be honest, but I can work that out for sure.

cmd/gtfTools/filter.go
```diff
@@ -73,6 +77,13 @@ func gtfFilter(s FilterSettings) {
...
+			currTranscripts = records[currGene].Transcripts
+			for t := range currTranscripts {
+				if currTranscripts[t].Chr != s.ChromFilter {
+					pass = false
```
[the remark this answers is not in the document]
cfauci: My preference is all since my current use case is to basically just break the gtf by chroms right now but i could make both an option.

rimangan (approved):

edotau (approved):

hcnapier (approved):

PR #991, opened 2024-02-14
pFaVisualiser

saqzhao:
# Description
* PFaVisualiser prints pFa as human-readable text to file. Result can be read to terminal using ```cat <filename> | column -t```
* RoundSigFigs rounds a given float64 to a significant number of digits.

### Testing
Stability tests for PFaVisualiser (and helper functions) and RoundSigFigs.

### Screenshots & Media
Example of output in terminal:
image

Due to floating point errors, the last one (or two) places might be off by 1. e.g. 0.8325 round to 0.832 instead of 0.833

- [*] I performed a self-review of my code
- [*] If it this a core feature, I added thorough tests
- [*] The command `go fmt` or `make clean` was used on all files included


rimangan (approved):

dasnellings (approved):
Cool PR! I am approving, but for future improvements of this you may consider handling cases where the reference name is longer than the line. I noticed that if I made the chromosome name longer it misaligned the first row. You might consider adding blank space to each line equal to the length of the chromosome name to maintain alignment, though you may think of some more elegant solutions to this problem. Nice job!

Screenshot 2024-03-08 at 5 26 55 PM

PR #992, opened 2024-02-20
gtfTools toBed firstTwoCodonBases

rimangan:
# Description
In sequence conservation analysis, the first two bases of a codon often have significantly higher conservation values than the last base due to the degeneracy of the genetic code. Consider the example below of phyloP signals across exons.
image
As such, I wanted a tool to separate out the first two bases of each codon into a bed file, and added this -firstTwoCodonBases option to the gtfTools toBed subcommand. This program is aware of both frame and strand (see snapshots below of output on the browser).
image

### Testing

None


saqzhao (approved):

sweavs111:
Hey Riley I just reviewed this PR. It looks good and I am ready to approve, I just had a question about the testing files. Why are there two expected.firstTwoCodonBases.bed testing files? I see one is for the cmd and one is for the package, but are these redundant? Also I see one is quite long, is that for the entirety of chr1?

rimangan:
> Hey Riley I just reviewed this PR. It looks good and I am ready to approve, I just had a question about the testing files. Why are there two expected.firstTwoCodonBases.bed testing files? I see one is for the cmd and one is for the package, but are these redundant? Also I see one is quite long, is that for the entirety of chr1?

Hi Seth, these are indeed redundant files, as one lives in the gtf package, and the other lives in the corresponding command folder. They are not genome wide, but rather refer to all exons for a few genes, as specified by the test.gtf file in the gtf package. I wanted a large enough test to see it span over multiple exons, but for a stability test we could consider scaling this down.

cfauci (approved):

PR #994, opened 2024-02-21
clean up: Genome Graph eau

edotau:
# Description

- Mostly spring cleaning, removing functions redundant functions implemented with improvements over time that should be removed.
- A little bit of reorganizing to keep the Seed structs in one file for easier access.
- Graph traversal logic to align sequences were previously implemented in two haves, traversing the left and/or right traversals. Traversals in either direction is very similar so most right/left traversals have been merged into one function. Some of the other traversals are complex, but here is is one of the similar ones below:
- https://github.com/vertgenlab/gonomics/blob/5814d737c4387ab21702e601ae0818d0902f4066/genomeGraph/search.go#L146-L171

### Testing
1. test begins by setting up a simple genome graph for testing. This graph is composed of three nodes (`n0`, `n1`, `n2`), each representing a sequence of DNA bases. Two edges (`e0`, `e1`) are created to connect these nodes, forming paths in the graph
2. test then defines expected results for the right and left traversals. These expected results include the expected alignment score, the expected alignment in terms of matches, mismatches, insertions, and deletions 
https://github.com/vertgenlab/gonomics/blob/5814d737c4387ab21702e601ae0818d0902f4066/genomeGraph/search_test.go#L12

## Genome Graph Traversals
https://github.com/vertgenlab/gonomics/blob/5814d737c4387ab21702e601ae0818d0902f4066/genomeGraph/search.go#L84-L144
https://github.com/vertgenlab/gonomics/blob/5814d737c4387ab21702e601ae0818d0902f4066/genomeGraph/search.go#L22-L82


genomeGraph/align.go
```diff
@@ -114,14 +95,6 @@ func scoreSeed(seed *SeedDev, read fastq.Fastq, scoreMatrix [][]int64) int64 {
 	return score
 }
 
-var HumanChimpTwoScoreMatrix = [][]int64{
```
edotau, line 117: Moved this to align package to be more centralized usage from other packages

genomeGraph/toGiraf.go
```diff
@@ -78,20 +82,6 @@ func readFastqGsw(fileOne string, fileTwo string, answer chan<- fastq.PairedEndB
 	close(answer)
 }
 
-type ScoreMatrixHelper struct {
```
edotau, line 81: Unused

genomeGraph/tournamentTree.go
```diff
@@ -1,351 +0,0 @@
-package genomeGraph
-
-import (
-	"log"
-
```
edotau, line 5: This was an improvement to the first development and is moved to seed.go:
- https://github.com/vertgenlab/gonomics/blob/1d145e8a8bc1529182def67895aeadffff5ca970/genomeGraph/seed.go#L21-L29

genomeGraph/toGiraf_test.go
```diff
@@ -1,26 +0,0 @@
-package genomeGraph
-
-/*
-import (
```
edotau, line 4: Commented out tests, lets just delete this file and we can always revert or roll back

PR #995, opened 2024-02-26
haplotypeGenerator - include reference genome in output

rimangan:
# Description
haplotypeGenerator is a gonomics command that takes a VCF file (specifying variants from an individual genome) and a fasta file (specifying the reference genome sequence) and returns a fasta file of each of the individual genotypes. This PR adds the option to 'includeRef'. When set to true, the program returns a fasta file with one additional element, the reference haplotype, in the first position. Note that as this program currently only supports substitutions from VCFs, this renders the output a valid multiFa.

### Testing

None


cfauci (approved):

craiglowe (approved):
Great!

PR #997, opened 2024-02-26
bedCountBam

sweavs111:
New command bedCountBam.go

This cmd takes an pos sorted and indexed bam file and a bed file and returns read counts for each bed regions (which must have a name field). An option for reporting counts per bp instead of raw counts is present.

Per Dan's suggestion, instead of streaming through the whole alignment file i use the bamSeeker command which is why a sorted/indexed bam is required.

This command should have pretty broad use cases, but I most recently used this to quantify gRNA abundance in a CRISPRi library.

Testing function and files present


dasnellings:
Just saw this come up and wanted to recommend an alternative that may save you some time. Unless you need this to work on sams an alternative method would be to used an indexed bam and call `SeekBamReagion(bamReader, bed.Start, bed.End)` which will return the reads overlapping the bed region. You could then just divide the number of reads by the length of the region. This will likely be a lot faster than reading through a whole same file, especially if you bed regions cover a small portion of the genome.

sweavs111:
Thanks Dan, i didn't know that function existed. definitely will use it

dasnellings (approved):

rimangan (approved):

PR #1002, opened 2024-03-04
PFaVisualiser cmd

saqzhao:
# Description
Command-line for browser PFaVisualiser
```pFaVisualise in.pfa start end out.txt```
Prints a subsequence of a pFasta to file in human-readable format.
These values can be pasted into Excel or other programs to graph as a heatmap / matrix.

### Testing

Standard stability tests

### Screenshots & Media
image

image

### CODE REVIEW

File: browser/pFa.go [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+		_, err = fmt.Fprintf(out, "Start: %d. End: %d. %s: %d.", start, end, formatting, formatNum)
+		exception.PanicOnErr(err)
+	} else {
+		_, err = fmt.Fprintf(out, "Start: %d. End: to end. %s: %d.", start, formatting, formatNum)
+		exception.PanicOnErr(err)
+	}
+
+	if seqName == "" {
+		// only accepts single pFa or a .pFa with multiple pFastas and a specified record name
+		if len(records) > 0 {
```
RavenYL, line 46: if the pfasta file is empty, would len(records) be ==0?

browser/pFa_test.go
```diff
@@ -0,0 +1,233 @@
...
+	NumIters:      6,
+	LineA:         make([]float32, 30),
+	LineC:         make([]float32, 30),
+	LineG:         make([]float32, 30),
+	LineT:         make([]float32, 30),
+	Start:         2,
+	Records: []pFasta.PFasta{pFasta.PFasta{Name: "chr1",
+		Seq: []pDna.Float32Base{
+			pDna.Float32Base{
+				A: 0.23857,
```
RavenYL, line 37: It seems like the pDna bases do not add up to 1, so they are likelihoods not probabilities. Maybe we should decide if we want to enforce probabilities, so each pDna base is "either {0 0 0 0} or all bases add up to 1"?

rimangan: We should have this discussion together, but we were thinking of not restricting the sum to 1, as we could theoretically store things other than valid probability vectors in this format. We could consider, for example, representing DNA sequences with uncertainty where the total magnitude of a vector is representative of another value, like conservation or chromatin accessibility. While we anticipate mostly using these for probabilities, I wanted to keep that option open

browser/pFa_test.go
```diff
@@ -0,0 +1,233 @@
...
+	Expected      string
+}{{LineLength: 10,
+	SetOfLinesIdx: 0,
+	NumIters:      6,
+	LineA:         make([]float32, 30),
+	LineC:         make([]float32, 30),
+	LineG:         make([]float32, 30),
+	LineT:         make([]float32, 30),
+	Start:         2,
+	Records: []pFasta.PFasta{pFasta.PFasta{Name: "chr1",
```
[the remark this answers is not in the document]
rimangan: pFaTools sample is essentially a pfasta-to-fasta. Craig had mentioned making a 'consensus' tool, which takes the consensus sequence instead of a posterior sample, might be another valuable future addition.

As for fasta-to-pfasta, I agree it'd be cool to make a tool that makes one-hot encoded pFasta sequences from a fasta sequence. No uncertainty, but this at least gets us to the right data type

browser/pFa.go
saqzhao: Yes, I think the best thing to do if len(records)==0 would be to give a log.FatalF. I'll try to add that in a commit this week.

rimangan (approved):

browser/testdata/pfa_PFaVisualiser_normalised_input_toy_2.pfa
```diff
@@ -0,0 +1,6 @@
+pFasta_format_1.0
+chr2	1
+chr1butlikerealllllllllllllylong	17
+EndHeader
+f.f2�4f6�2�446�/+5�6+0�/�1u4�076J400�4�4�7�/�-�424F,i0D81 6�54�7�3b1-.�2s/�72F51/}7�-f24�5F2�.�6�6m+w+7t+7m5M1�2�4�/93�1�7A04�54
+7�4�.�0
```
cfauci, line 6: this document needs a new line at the end

cmd/pFaTools/pFaVisualize.go
```diff
@@ -0,0 +1,97 @@
...
+	Chrom	   string
+	StartOfAlignment bool
+	EndOfAlignment bool
+}
+
+// VisualizeUsage defines the usage statement for the pFaTools Visualize subcommand.
+func VisualizeUsage(VisualizeFlags *flag.FlagSet) {
+	fmt.Printf("pFaTools Visualize - Provides human-readable sequence from a given pFa.\n" +
+		"Keyword 'START' for the end argument makes a visualization until the end of the pfasta.\n" +
+		"Keyword 'END' for the end argument makes a visualization until the end of the pfasta.\n" +
```
cfauci, line 32: why do these do the same thing?

cmd/pFaTools/testdata/test_visualize_input_1.pfa
```diff
@@ -0,0 +1,7 @@
+pFasta_format_1.0
+chr2	1
+chr1butrllllllylong	30
+EndHeader
+f.f2�4f6�3Q517v0X57Q0�/&6�8C5�:�7�3c8a8c;�3�1�8�7�/4�;E ;8�8�7�8s432/�6�3�;6G5�/�7f.f2�5�6�3�3M;M;0�/�;�/z;%6�1�335�3P7�5�;�3K7D9T7�:�8�2�4�(�(k;[-s+
+5�8T:�;�/�5�4|5�;�4�7�7�72d8�5�8X.�:�)%6�7�8�;)2�;
+S!�9�3�(.+'�;:�4�6�/;T6�:76m)�8B5�9
```
cfauci, line 7: new line needed

cfauci (commented):
The pfa outputs all need new line characters at the end of the files, or there is something otherwise amiss with the formatting that git doesn't like. Is this a product oft he file type?

cfauci (approved):

rimangan (approved):

PR #1003, opened 2024-03-08
bigWig Package Part 1: bbi header

rimangan:
# Description
This PR starts a package to build a parser for the bigWig file format in gonomics. BigWig and bigBed both start with a shared header called the bbiHeader. Byte-by-byte details are specified below, from the official documentation. This table was found in the supplementary materials of the bigWig paper: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2922891/. Note these specs are for bigWig version 3, but when I ran wigToBigWig, it gives version 4. @craiglowe  would you happen to know if there are newer specs or if they have changed significantly between versions? If not, I'm largely following the version 3 spec sheet to put together this parser. 
To reduce PR size, this first PR just parses the shared header. Currently it is written with bigWig in mind, but can easily be refitted should we choose to develop a bigBed package.
bigWig is organized as a subpackage of wig.

image

### Testing

None


wig/bigWig/bigWig.go
```diff
@@ -0,0 +1,73 @@
...
+	UncompressBufferSize uint32 // Maximum size of decompression buffer needed (nonzero on compressed files).
+	Reserved             uint64 //Unused values, reserved for future expansion. Should always be zero in a valid bigWig.
+}
+
+func readHeader(file *fileio.EasyReader) Header {
+	var header = Header{}
+	err := binary.Read(file, binary.LittleEndian, &header.Magic)
+	exception.PanicOnErr(err)
+	if header.Magic != bigWigMagic {
+		log.Fatalf("Error: bigWig magic was not as expected. Found: %v. Expected: %v.\n", header.Magic, bigWigMagic)
```
dasnellings, line 39: Might be worth incorporating a check here to see if the file was encoded as BigEndian and returning a more specific error if so (since the specs say the magic bytes can be byte swapped). I don't think it's worth adding support for BigEndian architecture, but it's easy enough to do the check if the magic bytes don't match and saying we don't support BigEndian.

dasnellings (commented):
Small suggestion, otherwise looks good.

wig/bigWig/bigWig.go
rimangan: Totally agree, just made that change. Currently we check if it's the BigEndian magic and throw a specific error, and I added a note in the function header that we only support little endian. Like I said, I just used wigToBigWig and it makes little endian headers, the headache will be if this is not standard in kentUtils or other big toolsets

craiglowe:
It might be good to see if we can figure out the specs for version 4, because I am not sure if they will be compatible with each other.  I don't think this will be a running-to-stand still thing, because it looks like version 4 has been the version for about a decade with no further changes.
The comments in this file, might be the best place to get (figure out) the version 4 format:
https://github.com/ucscGenomeBrowser/kent/blob/master/src/inc/bbiFile.h
Line 89 makes it seem like version 3->4 was a minor bug fix.  Maybe lines 10 to 69 can be used as a format spec for version 4 since it should mostly match the version 3 doc you have, but with a minor change?

RavenYL (approved):

dasnellings (approved):

PR #1009, opened 2024-04-15
compareSimilarSeqs

sweavs111:
small PR introducing a new sequence comparison function. This function will take in 2 []dna.Base and a mismatch threshold. If the number of nucleotide mismatches is equal or less to the threshold it will pass, if it is greater, it will fail. It is case insensitive. Testing functions present

### DIFF
--- a/dna/compare.go
+++ b/dna/compare.go
@@ -164,9 +164,9 @@ func min(a int, b int) int {
 	return b
 }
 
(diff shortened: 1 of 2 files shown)


hcnapier (dismissed):

craiglowe (approved):
Great!

hcnapier (approved):

PR #1014, opened 2024-04-25
Fixed copy/paste error in sam.ParseTags

dasnellings:
Fixes an error in sam.ParseTags where the tag type was improperly formatted due to copy pasting code and forgetting to change the formatting type. Of course this slipped through because there were no tests for ParseTags (shame on me). If I had taken 10min to write a test when I wrote ParseTags years ago, then I could have saved myself hours of troubleshooting in 2024 😭 

Also fixed some improper handling of the 'B' case when processing tags in bam writing.


sam/bamWrite.go
```diff
@@ -275,6 +276,10 @@ func writeExtra(bw *BamWriter, s Sam) {
 // retrieveTriplet splits a tag string into type, number, and value.
 func retrieveTriplet(tag string) []string {
 	comp := strings.Split(tag, ":")
+	if comp[1] == "B" {
```
edotau, line 279: 1) would this function effectually run strictly if `len(comp) = 3 or greater than 3`?
2) Do we only return exactly 3 items due to `comp[:3]`
I tried something along the lines of this and it cleans up the need to check the len multiple times. It passed your tests, but I didn't do a deeper dive into the functions as the bam writer is pretty complex:

```golang
func retrieveTriplet(tag string) []string {
	comp := strings.Split(tag, ":")
	if len(comp) >= 3 {
		if comp[1] == "B" {
			comp[1] = comp[1] + comp[2][0:1] // the B case has another value as part of the array that needs to be stored in the second part of the triplet
			comp[2] = comp[2][2:]  
		}
	} else {
		log.Panicf("malformed auxiliary data '%s'", tag)
	}
	comp[2] = strings.Join(comp[2:], ":") // rejoin value component of tag
	comp = comp[:3]
	return comp
}
```

dasnellings: Yes, the return should always be len of 3, though comp may be >= 3. 

I incorporated your suggestion, though with minor edits to keep the check that the members of the triplet are the correct number of characters.

sam/bamWrite.go
```diff
@@ -307,16 +305,22 @@ func writeTriplet(bw *BamWriter, triplet []string) {
 
 	typ := triplet[1]
 	values := strings.Split(triplet[2], ",")
+	if len(values) == 1 && values[0] == "" {
+		values = nil
+	}
+	realTyp := typ
```
RavenYL, line 311: Since realTyp is declared to be := typ, as opposed to a copy of typ, would changing realType to typ[1:2] on line 315 have weird effects since realTyp and typ might be linked? I'm just remembering that I've written code in the past where I had to make copies of similar values and avoid using "a := b".

Also, at code review we agreed that as soon as this bug fix gets merged into main, as long as we don't notice another bug, we will update to a new version of gonomics

edotau (approved):

sam/bamWrite.go
dasnellings: Sorry, it is less clear with the := operator, but `typ` is actually a `string` not a `[]string`. So `realTyp := typ` is copying `typ` not referencing it. So line 315 changes `realTyp` but not `typ`

RavenYL (approved):

PR #1016, opened 2024-04-30
bigWig Part IV: Fixing Chrom Tree

rimangan:
# Description
This PR fixes the bugs present in Part III, and can be merged in its place. I've rearranged some files, so I've left them as two open PRs for now to make it easier to see which lines have changed.

### Testing

None


wig/bigWig/io.go
```diff
@@ -0,0 +1,173 @@
...
+	var header = BbiHeader{}
+	err := binary.Read(file, binary.LittleEndian, &header.Magic)
+	exception.PanicOnErr(err)
+	if header.Magic == bigWigMagicBigEndian {
+		log.Fatalf("Error: bigWig file appears to be in big endian. Current functionality only supports little endian bigWig files.\n")
+	}
+	if header.Magic != bigWigMagic {
+		log.Fatalf("Error: bigWig magic was not as expected. Found: %v. Expected: %v.\n", header.Magic, bigWigMagic)
+	}
+	err = binary.Read(file, binary.LittleEndian, &header.Version)
```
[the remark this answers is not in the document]
rimangan: DecodeBinaryField is clearly a very good suggestion, I'll implement now. Could you clarify what you mean by grouping the binary.Read by int data type?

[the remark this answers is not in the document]
edotau: I was thinking about the different int types used to encode data in (i.e. uint8, uint16, uint32, uint64, etc...)

For example, if all the uint32(s) need to be processed together. `ChromId, ChromStart, ChromEnd` are all uint32(s) so I might be tempted to write something like decodeBinaryUint32() to process a slice of uint32 and run everything in a loop to reduce the number of times I have to code up DecodeBinaryField().

I am not saying that's something we should do here, however I'm always thinking of ways to strategically reduce the number of times I have to make the same function call. (A less is more approach).

Having said that, I didn't look too deep into the bigwig structure/design so I'll leave it up to you to design what's the best fit

wig/bigWig/io.go
```diff
@@ -0,0 +1,141 @@
...
+			numItems++
+		}
+		answer = append(answer, currNode)
+	}
+
+	return answer
+}
+
+// decodeBinaryField reads binary data from an input *fileio.EasyReader to a variable.
+func decodeBinaryField(file *fileio.EasyReader, data any) {
```
edotau, line 138: Looks good, I think it makes sense to put this in the fileio. I'd say it's potentially reusable by other future projects. We live in world of large data sets and binary operations on data is common practice.

Make sure this is covered by a small test. I'm happy to approve once that's complete. Something along the lines of:
```golang
func TestDecodeBinaryField(t *testing.T) {
	// Define test data
	var testData uint32 = 123

	binaryReader := fileio.EasyOpen(filename)

	// Call the function being tested
	var decodedData uint32
	decodeBinaryField(binaryReader, &decodedData)

	// Check if the decoded data matches the test data
	if decodedData != testData {
		t.Errorf("Error: Decoded data does not match test data. Expected: %d, Got: %d", testData, decodedData)
	}
}
```

cfauci (commented):
I agree with Eric that the binary file reader should go in fileio it will almost certainly be reused somewhere else in the future, but it looks great!

wig/bigWig/io.go
dasnellings: If you do opt to put this in fileio I would suggesting putting `LittleEndian` or `Le` or something into the name to denote the endianness.

edotau (changes requested):
A couple comments from the previous review went unaddressed.
1. @dasnellings: brought up a good point that the name should probably denote little (or LE, little, etc...) [endianness](https://developer.mozilla.org/en-US/docs/Glossary/Endianness).
2. Please include a simple test case for the function added to fileio.go. where are you are strictly **not** adding a file into "testdata". but define an example of little-endian `(ex. "0x78 0x56 0x34 0x12")` as an input representation wrap in an. io.Reader{}, and convert to a human readable value. See example/outline below:

```golang
func TestDecodeBinaryField(t *testing.T) {

	var input string = "0x00: f0 de bc 9a 78 56 34 12"
	var expected uint32 = 1234567890

	reader := strings.NewReader(input)

	var decodedData uint32
	DecodeBinaryField(reader, uint32)

	// Check if the decoded data matches the test data
	if decodedData != expected {
		t.Errorf("Error: Decoded data does not match test data.%d != %d\n", decodedData, expected)
	}
}
```

fileio/byteio.go
```diff
@@ -165,3 +166,9 @@ func IntSliceToString(nums []int) string {
 func IntToString(i int) string {
 	return fmt.Sprintf("%d", i)
 }
+
+// DecodeBinaryField reads binary data from an input *EasyReader to a variable.
+func DecodeBinaryField(file *EasyReader, data any) {
```
edotau, line 171:
```suggestion
func DecodeBinaryField(io.Reader, data any) {
```
- I think the input should actually be [io.Reader](https://pkg.go.dev/io#Reader) and not EasyReader. If you use EasyReader here, you'll be essentially handcuffing yourself making it slightly harder to write simple and concise tests . This function will be less flexible to others because any struct that implements the Read() method can call this function and not the other way around.

edotau (approved):
lgtm!

edotau (approved):

cfauci (approved):

PR #1023, opened 2024-05-08
Update to samCoverage & Poisson Distribution Functions

nikithat23:
Wrote the function samCoverage to calculate the genome coverage as the quotient of aligned bases in a sequencing dataset to the total length of ungapped genomic regions in the reference genome. The function checks the fit of the coverage data to a Poisson distribution using the PoissonHistogram function (newly written in the fit package). 

Poisson Distribution-related Functions created/modified to aid with this: 
- PoissonHistogram
- PoissonDist
- PoissonDistLog


cmd/samCoverage/samCoverage.go
```diff
@@ -0,0 +1,103 @@
...
+func samCoverage(s Settings) {
+	out := fileio.EasyCreate(s.OutFile)
+	fmt.Fprintf(out, "Coverage\tPileups\tGroup\n")
+	data, header := sam.GoReadToChan(s.SamFileName)
+	piles := sam.GoPileup(data, header, false, nil, nil)
+	histogram := make([]int, 30)
+	totalCount := 0
+
+	for p := range piles {
+		//if else structure to calculate pile depth
```
rimangan, line 35: Is this comment from an old version? It might be ok to delete

cmd/samCoverage/samCoverage.go
```diff
@@ -0,0 +1,103 @@
...
+	SamFileName   string
+	OutFile       string
+	CountNinDepth bool
+	Verbose       int
+}
+
+// samCoverage calculates genome coverage as the quotient of aligned bases in a sequencing dataset to the total length of ungapped genomic regions in the reference genome.
+func samCoverage(s Settings) {
+	out := fileio.EasyCreate(s.OutFile)
+	fmt.Fprintf(out, "Coverage\tPileups\tGroup\n")
```
rimangan, line 28: There's an unhandled error here. Consider if the outFile is deleted while it's open, or if for some reason it can't be written to. In this case, fmt.Fprintf would return an error, and we'd want to catch if it were non-nil. We can change to something like this:
`_, err = fmt.Fprint(out, "Coverage\tPileups\tGroup\n")`
`exception.PanicOnErr(err)`

cmd/samCoverage/samCoverage.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		histogram[depth]++
+		totalCount++
+	}
+	lambda := fit.PoissonHistogram(histogram)
+	for i, pileups := range histogram {
+		fmt.Fprintf(out, "%v\t%v\tEmpirical\n", i, pileups)
+		y, outlier := numbers.PoissonDist(i, lambda, false)
+		if !outlier {
+			fmt.Fprintf(out, "%v\t%.6g\tExpected\n", i, y*float64(totalCount))
```
rimangan, line 51: See above

fasta/pFasta/pFasta_test.go
```diff
@@ -1,9 +1,12 @@
 package pFasta
 
 import (
-	"github.com/vertgenlab/gonomics/dna/pDna"
 	"math/rand"
+	"os"
```
rimangan, line 5: Thank you for catching this! For everyone else reviewing, this was a small bug in pFasta that Nikitha and I noticed in our meeting. Now fixed

rimangan (commented):
Looks great! Just a few small edits

edotau (commented):
Looks good, summarizing a couple suggestions:

1) Reuse wrapper function `fileio.MustRemove(filename)` to remove file + error handling
https://github.com/vertgenlab/gonomics/blob/caae8389f1d1eb4c798a2a3dbef5286e4d332a92/fileio/fileio.go#L46-L53
2) Compare approximation of float64 values to precision threshold with `numbers.ApproxEqual(a, b, epsilon float64)` instead of string casting and checking for string matching
https://github.com/vertgenlab/gonomics/blob/caae8389f1d1eb4c798a2a3dbef5286e4d332a92/numbers/numbers.go#L213-L217

fasta/pFasta/pFasta_test.go
```diff
@@ -74,7 +76,7 @@ var WriteTests = []struct {
 		Precision: 1e-3, // float16 for writing is quite inaccurate, this fails at 1e-4
 	},
 	{
-		OutFile:   "testdata/out.test.pFa",
```
edotau, line 77: I'd keep this the same as before:
`OutFile:   "testdata/out.test.pFa"`

cmd/samCoverage/samCoverage.go
```diff
@@ -0,0 +1,129 @@
...
+			histogramBuffer := make([]int, depth+10) // this is a slice of 0s used to extend histogram when required
+			copy(histogramBuffer, histogram)
+			histogram = histogramBuffer
+		}
+		histogram[depth]++
+		totalCount++
+	}
+	lambda := fit.PoissonHistogram(histogram)
+	coverageThreshold := ThresholdCalc(s.HighEndFilter, histogram, float64(totalCount))
+	fmt.Fprintf(statSummary, "Lambda\t%v\nCoverageThreshold\t%v\n", lambda, coverageThreshold)
```
rimangan, line 52: We may have another unhandled error here. You can adapt this to look like lines 58/59

cmd/samCoverage/samCoverage.go
```diff
@@ -0,0 +1,129 @@
...
+	for i, pileups := range histogram {
+		_, err = fmt.Fprintf(outHist, "%v\t%v\tEmpirical\n", i, pileups)
```
rimangan, line 54: Here it seems we make a header with three fields but only write two columns. Perhaps we should add group back in? It can be either user defined, or instead, we can just make the third column 'Filename'

rimangan (commented):
Great work! I found a few small remaining issues

rimangan (approved):

cfauci (approved):

PR #1024, opened 2024-05-09
fix: fileio test benchmarks failing eau

edotau:
# Description

🐛 Bug Report
```error
2024/05/08 13:14:34 open testdata/big.fa: no such file or directory
exit status 1
```

Running benchmarks in filio package currently fails if users do not unzip big.fa.gz
```bash
# Run benchmarks in filio package
go test -bench
```

### Testing
- moved stdin from easyio to fileio
created a temp. file that is set to os.Stdin and openning a new file with MustOpen("/dev/stdin")
```golang
os.Stdin = mockStdin
file := MustOpen("/dev/stdin")
```
- stdin with gzip test

### Edge Cases
```text
"",                 // Empty string
"\nHello\nWorld\n", // Multiple lines
"Secret input with special characters !@#$%^&*()_+",
```


rimangan (dismissed):

fileio/easyio.go
```diff
@@ -39,21 +39,9 @@ func EasyOpen(filename string) *EasyReader {
...
 
-	if strings.HasPrefix(filename, "stdin") {
-		// when reading stdin we will assume the input is gzipped
-		// if the file begins with the two magic gzip bytes 1f8d.
-		// If it does, append .gz to the filename so it is parsed
-		// as gzip in the following switch case.
-		answer.File = os.Stdin
-		readerInput, hasMagicGzip = newStdinMagicReader(magicGzip)
-		if hasMagicGzip {
-			filename += ".gz"
```
dasnellings, line 50: With the new setup moving the stdin check to MustOpen, I think this will error if we do the filename manipulation for the case where the input is from stdin and the stream is gzipped.

cfauci: How would this case happen? If the output is set to be gzipped Then it's put into a stdin?

edotau: I think @dasnellings  is correct and that if a user were to pipe and stdout into stdin that is gzip would technically throw an error. I actually believe we can remove the check for the filename all together and just peek at the header for something that is gzipped and simplify the code to something like this:

```golang
// EasyOpen opens the input file. Panics if errors are encountered.
func EasyOpen(filename string) *EasyReader {
	if strings.Contains(filename, "http") {
		return EasyHttp(filename)
	}
	answer := EasyReader{
		File: MustOpen(filename),
	}
	switch {
	case IsGzip(answer.File):
		var err error
		answer.internalGzip, err = pgzip.NewReader(answer.File)
		exception.PanicOnErr(err)
		answer.BuffReader = bufio.NewReader(answer.internalGzip)
	default:
		answer.BuffReader = bufio.NewReader(answer.File)
	}
	return &answer
}
```

edotau: If we would like more verbose checks I am happy to add the checks in with the filename manipulation. I just think if we dont check the file suffix at all and purely check run the check for a gzip with this function, we should achieve the same result 

https://github.com/vertgenlab/gonomics/blob/8e3fc50fac90f40b1520ba99f05980d707fa5998/fileio/magic.go#L18-L43

edotau: @cfauci What Dan brought is if someone were to use the MustOpen in a gonomics CLI in a way similar fashion to this, an error would likely come up:
```bash
cat file.gz | MustOpen("/dev/stdin") | stdout
```

dasnellings: I agree that removing the filename check is appropriate and solves the issue. I think I initially added it to match the behavior of `gunzip`, but I don't think that is necessary. 

The case that I am worried about is chaining together multiple commands that use a gzipped output. something like
`bcftools merge *.vcf.gz | vcfFilter stdin output.vcf.gz`

dasnellings (approved):

fileio/easyio.go
edotau: I think I wrote a small test to confirm that would work as exacted by mocking a gzip stdin, but i can try to find time to confirm with a tool like bcftools

cfauci:
Thank you for that discussion, I didn't get any notifications about this but it was really helpful!

PR #1030, opened 2024-05-25
feat: 35% speed up for writing vcf files

edotau:
# Description

- **~35%** improvements and optimizations for writing vcfs.
- when constructing strings, strings.Builder{} appears to be better performing when combining strings from multiple data fields using something like fmt.Sprintf("%s%n...").
- in the 'string.Builder{}' implementation, require a slightly more involved process to build out the final string, however the downstream performance gains is beneficial, especially when working with larger vcf files.

```
BenchmarkPreviousWrite-12                 4707           1229806 ns/op         20186 allocs/op
BenchmarkNewBetterWrite-12                6258            960211 ns/op          9274 allocs/op
```
- Key performance improvements to note: ~35% faster, and roughly 50% less memory allocation

### Testing
- At first glance, this pr may feel beefy but mostly contains tests for any modified code as well as test coverage in `filer.go`, most of first just check for conditions in vcf samples.

- More benchmarking:
```
edotau@eau-macBook-m3-pro:~/src/github.com/vertgenlab/gonomics/vcf$ go test -bench=. -benchtime=5s

BenchmarkWrite-12                   4707           1229806 ns/op          365237 B/op      20186 allocs/op
BenchmarkNewWrite-12               6274            955157 ns/op          963356 B/op       9274 allocs/op
BenchmarkWriteGzip-12               3270           1769856 ns/op         5903577 B/op      20230 allocs/op
BenchmarkNewWriteGzip-12            3499           1677263 ns/op         6501878 B/op       9318 allocs/op

PASS
ok      github.com/vertgenlab/gonomics/vcf      54.065s
```


vcf/io.go
```diff
@@ -200,47 +199,29 @@ func FormatToString(format []string) string {
...
 // Write writes a []Vcf to an output filename.
 func Write(filename string, data []Vcf) {
 	file := fileio.EasyCreate(filename)
 	WriteVcfToFileHandle(file, data)
-	err := file.Close()
-	exception.PanicOnErr(err)
-}
-
-// PrintVcf prints every line of a []Vcf.
-func PrintVcf(data []Vcf) {
```
edotau, line 235: Removing unused functions.

vcf/gVcf.go
```diff
@@ -84,35 +83,13 @@ func ReorderSampleColumns(input Vcf, samples []int16) Vcf {
...
-			answer += sampleToString(sample[i])
+			builder.WriteString("\t")
 		}
+		builder.WriteString(sample.String())
 	}
-	return answer
-}
-
-// sampleToString uses just an array of Sample structs to write to a string for simple gVCFs with just the allele info in notes.
-func sampleToString(s Sample) string {
```
edotau, line 100: Merged with String() method

vcf/io.go
```diff
@@ -186,61 +190,32 @@ func NextVcf(reader *fileio.EasyReader) (Vcf, bool) {
...
 }
 
 // WriteVcf writes an individual Vcf struct to an io.Writer.
-func WriteVcf(file io.Writer, input Vcf) {
-	var err error
-	if len(input.Format) == 0 {
-		_, err = fmt.Fprintf(file, "%s\t%v\t%s\t%s\t%s\t%v\t%s\t%s\n", input.Chr, input.Pos, input.Id, input.Ref, strings.Join(input.Alt, ","), input.Qual, input.Filter, input.Info)
-	} else {
-		_, err = fmt.Fprintf(file, "%s\t%v\t%s\t%s\t%s\t%v\t%s\t%s\t%s\t%s\n", input.Chr, input.Pos, input.Id, input.Ref, strings.Join(input.Alt, ","), input.Qual, input.Filter, input.Info, FormatToString(input.Format), SamplesToString(input.Samples))
+func WriteVcf(writer io.Writer, record Vcf) {
```
edotau, line 204: Consolidated Write functions for better reusability

vcf/filter_test.go
```diff
@@ -0,0 +1,203 @@
+package vcf
+
+import (
+	"testing"
+)
+
+func TestIsHeterozygous(t *testing.T) {
```
edotau, line 7: Very basic tests for helper functions to filter samples in vcf records

vcf/gVcf.go
```diff
@@ -40,18 +39,18 @@ func ChromPosToUInt64(chrom int, start int) uint64 {
 
 // PrintSampleNames takes a vcf header and prints the sample names from the "#CHROM" line
 func PrintSampleNames(header Header) string {
```
edotau, line 41: Note sure if this is used anyhere, may consider deleting

vcf/methods.go
```diff
@@ -1,21 +1,66 @@
...
 	"strings"
 
 	"github.com/vertgenlab/gonomics/dna"
 )
 
 // String implements the fmt.Stringer interface for easy printing of Vcf with the fmt package.
 func (v Vcf) String() string {
-	return fmt.Sprintf("%s\t%v\t%s\t%s\t%s\t%v\t%s\t%s\t%s\t%s\n", v.Chr, v.Pos, v.Id, v.Ref, strings.Join(v.Alt, ","), v.Qual, v.Filter, v.Info, v.Format, SamplesToString(v.Samples))
+	var buf strings.Builder
+	buf.Grow(256) // Pre-allocate some space for efficiency.
```
dasnellings, line 14: I have often thought about implementing a writer that functions similarly to the `GoReadToChan` functions except it outputs a sender channel that you can send records to for writing. The benefit is that you don't need to reallocate infrastructure for each write call and it would generally promote improved performance across the codebase as responsibility for writing is naturally delegated away from the main thread. I think with switching to `strings.Builder` the benefit of a `GoWriteFromChan` system would be even greater (though might require a specialized function that inputs the buffer). 

Anyway, nothing that needs to be changed. Just sharing some thoughts. I am a big fan of this improvement.

vcf/methods.go
```diff
@@ -1,21 +1,66 @@
...
+	var buf strings.Builder
+	if s.FormatData == nil {
+		buf.WriteByte('.')
+		return buf.String()
+	}
+	if s.Alleles == nil {
+		buf.WriteByte('.')
+	} else {
+		for i := 0; i < len(s.Alleles); i++ {
+			if i > 0 && i < len(s.Phase) {
```
dasnellings, line 53: I am interpreting this code as pulling the middle element in the Alleles slice and always interpreting that as the a phase element. If that is correct, there may be a bug here. VCF specs support polyploid input so the genotype field can be something like `1/1|2/3`. In these cases I believe this code would panic.

dasnellings: Might be good to add a test case for that as well. Maybe also a test case for when the allele number is double digits. I am pretty sure double digit alleles are handled correctly when you are building the Alleles slice, but I'd feel better if we had a test for it.

dasnellings (changes requested):
I think there may be a small bug, but otherwise I'm a big fan of the improvement.

vcf/methods.go
edotau: Could you give me a couple of string representation of the scenarios you are talking about? Not totally sure what I would see
- when you say genotype field can be: `./.` or just `.`
- allele number is double digits. (do you mean it's possible to see 10/11?)

vcf/methods.go
edotau: I definetely thought of something along those lines. One implementation of this method was to pass in the `strings.Builder` variable, so the function could look something like:
`func (v Vcf, buf strings.Builder) String() string {}` but I wasn't sure how it would perform within goroutines. I decided that this would be a good first pass and ensure satefy. I can add a TODO: to revisit this later

convert/convert.go
```diff
@@ -222,7 +222,7 @@ func ThreeWayFaToVcf(f []fasta.Fasta, chr string, out *fileio.EasyWriter) {
 					{
 						Alleles:    []int16{1, 0},
 						Phase:      []bool{false, false},
-						FormatData: []string{"1/0"},
```
edotau, line 225: This string is created from looping through the slice of alleles instead of FormatData, see below:
https://github.com/vertgenlab/gonomics/blob/362af0f127450e9a173350c395594bd7056dd66d/vcf/methods.go#L55-L61

vcf/io.go
```diff
@@ -186,61 +191,24 @@ func NextVcf(reader *fileio.EasyReader) (Vcf, bool) {
...
 // Write writes a []Vcf to an output filename.
 func Write(filename string, data []Vcf) {
 	file := fileio.EasyCreate(filename)
 	WriteVcfToFileHandle(file, data)
-	err := file.Close()
-	exception.PanicOnErr(err)
-}
-
-// PrintVcf prints every line of a []Vcf.
-func PrintVcf(data []Vcf) {
```
edotau, line 235: Removed, unused

vcf/methods.go
edotau: Added TODO(s) for the cases `.` and `./.`

dasnellings (approved):
Thanks for the additional tests. Nice to see VCF getting a big speedup!

rimangan (approved):

PR #1032, opened 2024-06-04
Ontology3

cfauci:
added an output from ontologyEnrichment for the enrichments of elements near GO terms, obs/expected calculation

### DIFF
--- a/ontology/ontology.go
+++ b/ontology/ontology.go
@@ -220,10 +220,12 @@ func ThreeDGreat(queries []bed.Bed, chromSizes map[string]chromInfo.ChromInfo, g
 
 	inputEnrichOut := fileio.EasyCreate(name + ".inputEnrichments.txt")
 	_, err = fmt.Fprintf(inputEnrichOut, "Term\tName\tEnrichment\n")
+	var queryProportion float64
 	var answer float64
 	log.Print(len(kCache))
 	for k := range kCache {
-		answer = float64(kCache[k] / n)
+		queryProportion = float64(kCache[k] / n)
+		answer = queryProportion / proportionsForTerms[k]
 		_, err = fmt.Fprintf(inputEnrichOut, "%s\t%s\t%e\n", k, ontologies[k].Name, answer)
 		exception.PanicOnErr(err)
 	}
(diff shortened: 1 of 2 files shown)


ontology/ontology.go
```diff
@@ -10,10 +10,12 @@ import (
...
 	"github.com/vertgenlab/gonomics/interval"
-	"github.com/vertgenlab/gonomics/numbers"
+	//"github.com/vertgenlab/gonomics/interval/lift"
+	//"github.com/vertgenlab/gonomics/numbers"
 	"github.com/vertgenlab/gonomics/ontology/gaf"
 	"github.com/vertgenlab/gonomics/ontology/obo"
 	"log"
+	"math"
 	"strings"
 )
```
cfauci, line 21: genesAssignmentsFromGaf needs to be able to take genes listed in the original obo as well, or we need to exclusively use the common gaf method that I came up with as a temporary fix

bed/modify.go
```diff
@@ -111,3 +112,26 @@ func MergeHighMem(records []Bed, mergeAdjacent bool, keepAllNames bool) []Bed {
 	outList = append(outList, currentMax)
 	return outList
 }
+
+func BedsWithNameSet(regions []Bed, names []string) []Bed {
```
rimangan, line 116: Docs

numbers/distribution.go
```diff
@@ -276,17 +315,18 @@ func BinomialLeftSummation(n int, k int, p float64, logOutput bool) float64 {
...
 			return 0
 		} else {
 			return 1
 		}
 	}
-	if float64(n)*p > 10 && float64(n)*(1-p) > 10 {
+	if float64(n)*p > 5 && float64(n)*(1-p) > 5 && !exact {
 		return evaluateRightBinomialSumApproximate(n, k, p, logOutput)
 	} else {
+		log.Print("ended up here")
```
rimangan, line 329: Remove?

numbers/distribution.go
```diff
@@ -276,17 +315,18 @@ func BinomialLeftSummation(n int, k int, p float64, logOutput bool) float64 {
...
+func BinomialRightSummation(n int, k int, p float64, logOutput bool, exact bool) float64 {
 	if k == 0 {
 		if logOutput {
 			return 0
 		} else {
 			return 1
 		}
 	}
-	if float64(n)*p > 10 && float64(n)*(1-p) > 10 {
+	if float64(n)*p > 5 && float64(n)*(1-p) > 5 && !exact {
```
rimangan, line 326: Could you tell me about this change?

PR #1034, opened 2024-06-06
interval overlap helpful error message

dasnellings:
intervalOverlap just silently finishes if the file you input is not supported. This adds an error message when you try to use an unsupported file.

### DIFF
--- a/interval/query.go
+++ b/interval/query.go
@@ -1,6 +1,7 @@
 package interval
 
 import (
+	"log"
 	"path"
 
 	"github.com/vertgenlab/gonomics/axt"
@@ -57,6 +58,8 @@ func ReadToChan(inputFile string, send chan<- Interval) {
 		for val := range receive {
 			send <- val
 		}
+	default:
+		log.Fatalf("ERROR: file type of %s not supported by interval ReadToChan")
 	}
 	close(send)
 }


interval/query.go
```diff
@@ -57,6 +58,8 @@ func ReadToChan(inputFile string, send chan<- Interval) {
 		for val := range receive {
 			send <- val
 		}
+	default:
+		log.Fatalf("ERROR: file type of %s not supported by interval ReadToChan")
```
sweavs111, line 62: looks good to me, but looks like theres a go fmt issue that you may want to look into

sweavs111 (approved):

interval/query.go
```diff
@@ -57,6 +58,8 @@ func ReadToChan(inputFile string, send chan<- Interval) {
 		for val := range receive {
 			send <- val
 		}
+	default:
+		log.Fatalf("ERROR: file type of %s not supported by interval ReadToChan", inputFile)
```
edotau, line 62: Not a huge deal but would it be possible to do: 
```suggestion
		log.Fatalf("Error: file type of %s not supported by interval ReadToChan", inputFile)
```
Capital letter + lowercase. Either way works for me, as long as we stay consistent with repo styling.

edotau (approved):
One small style preference, but otherwise good

edotau (approved):

rimangan (approved):

PR #1048, opened 2024-06-30
matchedBed setting for simulateBed

sweavs111:
This PR adds a "matchedBed" option to simulate bed. 

A common application for simulateBed is to make a random bed file as a comparison point for a real data bed file in enrichment calculations. With matchedBed you can provide the real data bed file to simulateBed and it will calculate the average bed region size and number of bed regions. it will use these values for simulate bed to create the companion 'rand' bed file.

All tests created and passing


cfauci (dismissed):
Great idea!

cfauci (approved):

hcnapier (approved):

PR #1049, opened 2024-07-02
vcfToString no new line

sweavs111:
This PR removes the newline character from the `.String()` method for VCF structs. This is useful when you want to further modify a string with VCF data before printing. 

To still have `WriteToFileHandle` be smooth, I changed the `WriteVCF` function to use `Fprintf` instead of `Fprint` and added the newline char that way. This mirrors how the bed package handles this same issue.

The only testing file I had to alter was the actual VCF --> string test. All other VCF-related tests passed as is. So i don't believe this will mess up any existing code. BUT PLEASE LET ME KNOW IF YOU THINK IT WILL!

### DIFF
--- a/vcf/io.go
+++ b/vcf/io.go
@@ -200,7 +200,7 @@ func WriteVcfToFileHandle(file io.Writer, input []Vcf) {
 
 // WriteVcf writes an individual Vcf struct to an io.Writer.
 func WriteVcf(writer io.Writer, record Vcf) {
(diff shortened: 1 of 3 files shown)


rimangan:
I support this solution. Could you add a comment in the methods.go documentation for the String function explaining that it does not add a \n character?

dasnellings (approved):

cfauci (approved):

PR #1050, opened 2024-07-12
tests: fixed tests comparing strings

edotau:
# Description
- No functions change, only update to tests using `ApproxEqual()` to compare values to 6 decimal places as the default (instead of casting to strings and comparing)
- Small formatting and removing old comments


dasnellings (approved):

numbers/distribution_test.go
```diff
@@ -19,13 +18,13 @@ func TestBinomialDist(t *testing.T) {
 	expected2, _ := BinomialDist(20, 20, 0.6, false)
 	input3 := ExplicitBinomialDist(20, 0, 0.6)
 	expected3, _ := BinomialDist(20, 0, 0.6, false)
-	if fmt.Sprintf("%e", input1) != fmt.Sprintf("%e", expected1) {
+	if ApproxEqual(input1, expected1, defaultEpsilon) {
```
craiglowe, line 21: I agree that we this is a place where we should make the tests a little more rigorous.  However,  I think this is changing from relative precision to absolute precision.  For example, some of these answers will be pretty small probabilities and currently 1e-7 and 1e-15 will not be reported as "almost equal," but with the change I think any numbers that are both less than 1e-6 will be considered equal.  I think the comparison we want is something closer to "abs(a-b)/b < threshold"  rather than "abs(a-b) < threshold".  Maybe we should update ApproxEqual, or maybe we should write a new function called ApproxEqualRelative?

craiglowe (commented):
It is a great idea to improve these tests, but I don't think we want to change from relative to absolute precision.  I think a tiny update could keep these improvements, but have us stay with relative precision.

numbers/distribution_test.go
sweavs111: we were talking about this relative comparison function and think that we should perhaps divide the (a-b) expression by the min(a, b). This would be the most conservative test.

edotau: Yes, you are both absolutely correct, I think the function should be changed to reflect a relative comparison. I've updated the function and added a few more corner/edge cases based on your feedback

@sweavs111: I think there could be some cases where using the Min(a, b) would be a better test, but I think it makes more sense to update the `ApproxEqual()' to divided by the `Max()'

Let's say you have:
```
a = 1000
b = 1001
```
Using Max:

```
Relative Error = |a - b| / max(|a|, |b|) = 1 / 1001 ≈ 0.000999
```
- This accurately reflects that the difference is very small compared to the numbers themselves.

Using Min:

```
Relative Error = |a - b| / min(|a|, |b|) = 1 / 1000 = 0.001
```
- This overestimates the error, making it seem more significant than it actually is.

craiglowe: Hi Eric, thanks for providing this improved code!  I don't totally understand the example where 0.000999 is more correct than 0.001.  Can you explain some more?  What would you think about dividing by the average of a and b, instead of the min or max?

edotau: When I had initially wrote the example, my interpretation of this function is it's an error approximation and I had felt taking the Max was a more aggressive approximation vs. taking the Min() was a more conservative approach.

When I had written this function the use case in mind was to better perform assertions between float64 numbers for testing purposes. With that being said I don't have anything strong opinions against taking an average or any others `[Max, Min, Ave]`. I will change the function to take an average instead.

motif/compare.go
```diff
@@ -38,7 +39,8 @@ func ApproxEquals(alpha, beta string, epsilon float64) bool {
 			answerValue := parse.StringToFloat64(answerFields[index])
 
 			// Compare the parsed values for near equality
-			if !numbers.ApproxEqual(queryValue, answerValue, epsilon) {
+			// TODO: debug why this function fails with ApproxEqual() - relative
+			if !(math.Abs(queryValue-answerValue) <= epsilon)  {
```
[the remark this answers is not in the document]
edotau: forgot to remove this from previous version. removed and all tests are now passing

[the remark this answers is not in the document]
edotau: never mind seems like this is still giving issues.

Issue #821, opened 2023-04-17
Replace rand.Seed


dasnellings:
`rand.Seed` has been depreciated as of Go 1.2. We need to replace instances of `rand.Seed` in gonomics by generating a new random source that is only used by the local function and using the new source instead of the global rand source throughout the function.

PR #1053, opened 2024-07-12
fix: update rand.Seed() to use source seed value eau

edotau:
# Description

- https://github.com/vertgenlab/gonomics/issues/821
- Because rand.Seed has been deprecated as of Go 1.2, we now need to generate a new random source (rand.New(rand.NewSource(seedValue))) for each function, we avoid using the global random number generator, which can lead to unintended side effects and make the code less predictable and harder to test. Using a local random source ensures that the random number generation is isolated to the specific function, providing better control over randomness and reproducibility in simulations and tests.
- this may seem like a lengthy PR, however it is mainly a refactor and identifying places where setting the random source generator fails tests.

### Proposed changes
Pass in `*rand.Rand{}` into basic random generator functions (see below):
- https://github.com/vertgenlab/gonomics/blob/0dea8380c42bae0a29f4458e7307354b6d8681d8/numbers/random.go#L7-L22
- Large size due to updating testing files

## Relevant Links
- https://go.dev/blog/randv2

### Updated random generator from a local source
- https://gabrieleromanato.name/go-rand-seed-has-been-deprecated-a-solution
```go
package main

import (
"fmt"
"math/rand"
"time"
)

func main() {
     r := rand.New(rand.NewSource(time.Now().UnixNano()))
     randN := r.Intn(10)
     fmt.Println(randN) // Ex. 6
}
```

### TODO:
- Split + organize into smaller PR
- May seem big but PR includes gzip test files to reduce repo size

@dasnellings pointed out within the source code:

```text
// Unlike the default [Source] used by top-level functions, this source is not
// safe for concurrent use by multiple goroutines.
```

As a result another implementation to think about is adding either a weight group or mutex for some sync when generation numbers within goroutines:
 
```golang
// RandomSource encapsulates a random source and provides methods to generate random numbers.
type RandomSource struct {
	mu   sync.Mutex
	rand *rand.Rand
}

// NewRandomSource initializes a new RandomSource with a given seed.
func NewRandomSource(seed int64) *RandomSource {
	return &RandomSource{
		rand: rand.New(rand.NewSource(seed)),
	}
}

// RandIntInRange generates a random integer in the range [min, max).
func (rs *RandomSource) RandIntInRange(min, max int) int {
	rs.mu.Lock()
	defer rs.mu.Unlock()
	return rs.rand.Intn(max-min) + min
}
```


dasnellings:
Still in the process of reviewing, but I just saw this comment from the `NewSource()` godoc
```
// Unlike the default [Source] used by top-level functions, this source is not
// safe for concurrent use by multiple goroutines.
```
I still haven't wrapped my head around the implications of this. In order to maintain accurate test outcomes for functions with goroutines, we would need to get a different source for each one (or lock the source with a mutex). Though I suppose if we were pulling random numbers from multiple goroutines it should have already caused problems even using the default source... I guess we just don't have any tests where we pull random numbers from multiple goroutines???

I'm not sure yet if we should/can do anything about this, but I forsee this being a tricky bug in the future.

edotau:
> Still in the process of reviewing, but I just saw this comment from the `NewSource()` godoc
> 
> ```
> // Unlike the default [Source] used by top-level functions, this source is not
> // safe for concurrent use by multiple goroutines.
> ```
> 
> I still haven't wrapped my head around the implications of this. In order to maintain accurate test outcomes for functions with goroutines, we would need to get a different source for each one (or lock the source with a mutex). Though I suppose if we were pulling random numbers from multiple goroutines it should have already caused problems even using the default source... I guess we just don't have any tests where we pull random numbers from multiple goroutines???
> 
> I'm not sure yet if we should/can do anything about this, but I forsee this being a tricky bug in the future.

- Thanks, I know it's heavy so I'll leave it here for a while. I'd say a bulk/majority is updating testing files and some formatting. Seems like a lot of code is still making it into the repo without `go fmt`
- Not sure when I would use a seed source along with goroutines but i think it's most straight forward to have each goroutine have its own independent random number generator, avoiding any need for synchronization.
- Also open to new ideas and implementations. While refactoring these functions, I was thinking if it would be worth it to develop a random generator interface implement custom methods

fileio/fileio.go
```diff
@@ -137,17 +139,35 @@ func ReadHeader(reader *bufio.Reader) ([]string, error) {
...
-	var fileADone, fileBDone = false, false
-	var lineA, lineB string
-
 	fA := MustOpen(a)
 	defer fA.Close()
 	fB := MustOpen(b)
 	defer fB.Close()
-	readerA := bufio.NewReader(fA)
-	readerB := bufio.NewReader(fB)
+
```
edotau, line 148: Compare gzip files to quickly check for assertions

fileio/fileio_test.go
```diff
@@ -55,13 +57,28 @@ func TestEqual(t *testing.T) {
...
+		if r := recover(); r == nil {
+			t.Errorf("Error: Unknown file type should throw error...\n")
+		}
+	}()
+
 	// Will cause log.Fatal in this test context
 	MustCreate("")
-}
 
+	var buf bytes.Buffer
```
edotau, line 69: didn't like my previous test for catching empty string error. this test captures log error if inputing a blank string

edotau:
> Still in the process of reviewing, but I just saw this comment from the `NewSource()` godoc
> 
> ```
> // Unlike the default [Source] used by top-level functions, this source is not
> // safe for concurrent use by multiple goroutines.
> ```
> 
> I still haven't wrapped my head around the implications of this. In order to maintain accurate test outcomes for functions with goroutines, we would need to get a different source for each one (or lock the source with a mutex). Though I suppose if we were pulling random numbers from multiple goroutines it should have already caused problems even using the default source... I guess we just don't have any tests where we pull random numbers from multiple goroutines???
> 
> I'm not sure yet if we should/can do anything about this, but I forsee this being a tricky bug in the future.

Another way we could go is the implement the source as a struct with mutex to allow some sync/lock capability to those who want to implement goroutine driven functions while still retaining a simplistic approach to those writing simplier functions, for example something like this would work as well:

```golang
// RandomSource encapsulates a random source and provides methods to generate random numbers.
type RandomSource struct {
	mu   sync.Mutex
	rand *rand.Rand
}

// NewRandomSource initializes a new RandomSource with a given seed.
func NewRandomSource(seed int64) *RandomSource {
	return &RandomSource{
		rand: rand.New(rand.NewSource(seed)),
	}
}

// RandIntInRange generates a random integer in the range [min, max).
func (rs *RandomSource) RandIntInRange(min, max int) int {
	rs.mu.Lock()
	defer rs.mu.Unlock()
	return rs.rand.Intn(max-min) + min
}
```

PR #1054, opened 2024-07-14
fix: equal checks for gzip + better error handling

edotau:
# Description

- Fatal/Panics when appropriate got changed by mistake from PR #1024
    - fixed previously implemented fileio test for empty filename
- better error handling + exceptions
- `equal()` improved to handle gzip case (strictly for testing purposes)

In the case which 2 files are the same, but one is gziped, `equal()` will return true. Behavior could be up for discussion, but as of now is strictly geared towards testing purposes.


edotau:
> Hi Eric! Some questions about the pr: (1) RecoverPanicErr seems to log "expected panic" even when there isn't a panic, which is perhaps misleading. (2) It seems testPanicOnErr doesn't always trigger a panic, so I'm not sure I see the use case for RecoverPanicErr() (3) Could you tell me a bit about the recover mechanism? Shouldn't we log the actual panic value and handle it when it occurs? (4) log.Panicf can be recovered, right? Doesn't that go against the spirit of MustCreate, which by name implies that it has to die on error? This could make the function name confusing

Thanks for taking a look, but I'm going to go ahead and close this because it was written 4 or 5 months ago and I can't remember the motivation and these more mostly testing improvements which don't seem too meaningful to merge

PR #1057, opened 2024-07-15
Interval overlap mergeOutput

sweavs111:
update to intervalOverlap to have the -mergeOutput output print both overlapping intervals on the same line in the output file (instead of on consecutive lines)


hcnapier (approved):

craiglowe (approved):
Great!

Issue #686, opened 2022-06-21
value of sam cigar for unmapped read


craiglowe:
Right now an unmapped read sometimes has a "*" for the cigar and sometimes the cigar is nil.  We should probably standardize this and nil seem better since it is less memory.

PR #1060, opened 2024-07-18
refactor: merge cigar + byteCigar struct

edotau:
# Description

- Refactor ByteCigar + Cigar this struct with the following constants:
```go
type Cigar struct {
	RunLength int
	Op        byte
}

// Cigar operation runes as defined in the SAM format specification to represent the different alignment sequences
const (
	Match     byte = 'M' // Alignment match (can be a sequence match or mismatch)
	Insertion byte = 'I' // Insertion to the reference
	Deletion  byte = 'D' // Deletion from the reference
	Ns        byte = 'N' // Skipped region from the reference
	SoftClip  byte = 'S' // Soft clipping (clipped sequences present in SEQ)
	HardClip  byte = 'H' // Hard clipping (clipped sequences NOT present in SEQ)
	Padded    byte = 'P' // Padding (silent deletion from padded reference)
	Equal     byte = '=' // Sequence match
	Mismatch  byte = 'X' // Sequence mismatch
	Unmapped  byte = '*' // Represents an unmapped read in a SAM file.
)
```
- Issue: #686
    - All cigar functions currently check for `len([]Cigar) == 0` for unmapped cigars, with the exception of `ToString()`, seems related to 'Recycle' functions used in SeekBam()
    - An attempt to standardize checking cigar slices for mapped/unmapped. Currently only checks if the length of slice is equal to zero, however, we can add additional checks in one place if we deemed necessary.
```go

// IsUnmapped checks if provided cigars are unmapped. Note: Handle additional unmapped scenarios if needed.
func IsUnmapped(cigars []Cigar) bool {
	return len(cigars) == 0 // (empty slice)
        // Additional cases here:
}
```

### Testing
Benchmarks for parsing functions:
```text
BenchmarkCigarToString-12                5057869               237.7 ns/op            24 B/op          5 allocs/op
BenchmarkCigarNewToString-12            31650681                35.53 ns/op            8 B/op          1 allocs/op
PASS

BenchmarkCigarFromString-12             20072713               305.0 ns/op           116 B/op          5 allocs/op
BenchmarkCigarNewFromString-12          37938165               159.7 ns/op           136 B/op          6 allocs/op
```
See the following files:
- cigar_test.go
- compare_test.go
- tools_test.go


cigar/cigar.go
```diff
@@ -44,49 +81,42 @@ func NumDeletions(input []Cigar) int {
...
-func ToString(c []Cigar) string {
-	if len(c) == 0 {
-		return "*"
+func ToString(cigars []Cigar) string {
+	var buf strings.Builder
+	var err error
+
+	if IsUnmapped(cigars) {
+		exception.PanicOnErr(buf.WriteByte(Unmapped))
+		return buf.String()
```
sweavs111, line 90: Im guessing this returns an empty string?

cigar/cigar.go
```diff
@@ -3,22 +3,59 @@
...
+	SoftClip:  4,
+	HardClip:  5,
+	Padded:    6,
+	Equal:     7,
+	Mismatch:  8,
+}
+
+// IsUnmapped checks if provided cigars are unmapped. Note: Handle additional unmapped scenarios if needed.
+func IsUnmapped(cigars []Cigar) bool {
+	return len(cigars) == 0 // (empty slice)
```
sweavs111, line 52: so if the cigar struct has a '*', it will return that the read is mapped? Should we switch the function to handle the length check and the `operation == *` check

sweavs111 (commented):
I like the updated cigar struct it feels very clean and clear. I am not very familiar with the genomeGraph and giraf file types, so i couldn't deeply review those portions. I just had one question/suggestion on the IsUnmapped function. I seem to remember that when working with sam data I was running into bugs because sometimes the unmapped reads were empty cigars and sometimes they were '*'. Especially when reading from bam. I don't remember if we ever fixed that issue in the bam reader function (maybe you know), but it might be safer to check for both of those

edotau:
> I like the updated cigar struct it feels very clean and clear. I am not very familiar with the genomeGraph and giraf file types, so i couldn't deeply review those portions. I just had one question/suggestion on the IsUnmapped function. I seem to remember that when working with sam data I was running into bugs because sometimes the unmapped reads were empty cigars and sometimes they were '*'. Especially when reading from bam. I don't remember if we ever fixed that issue in the bam reader function (maybe you know), but it might be safer to check for both of those in the IsUnmapped function

Currently, unmapped reads are inconsistently handled across the codebase, with checks for both `len(cigars) == 0` and `cigars == nil`. This inconsistency has caused errors in the past, as referenced in (#842).

This PR unifies the logic by always checking `len(cigars) == 0` for unmapped reads. This approach is more reliable than checking for `cigars == nil` because it avoids unnecessary memory reallocation while maintaining consistency. When a read is unmapped, the cigar slice will simply have a length of zero, rather than being `nil`.

The function `Unmapped([]cigars)` (defined in the cigar package) will now serve as the standard way to check if a read is unmapped, ensuring that the logic is uniform across the repo. If any edge cases arise in the future, adjustments can be made within this function.

The check for `Op == '*'` is no longer necessary, as the `FromString()` function returns a zero-length cigar slice for unmapped reads. Similarly, when converting the cigar back to a string, an empty cigar slice will correctly be written as `'*'`.
https://github.com/vertgenlab/gonomics/blob/bfa0047ed97a0049100154a42cced8c8b027de6c/cigar/cigar.go#L101-L114
https://github.com/vertgenlab/gonomics/blob/bfa0047ed97a0049100154a42cced8c8b027de6c/cigar/cigar.go#L84-L98

sweavs111:
Ah I see, thank you for the clarification. I assume that all sam/bam readers use the FromString() function to parse the cigar? If so, all looks good to me!

edotau:
> Ah I see, thank you for the clarification. I assume that all sam/bam readers use the FromString() function to parse the cigar? If so, all looks good to me!

Yes that's correct. You would identify the column in the Sam row that contains the cigar and call `FromString()`. For bam the Op(s) are encoded as Uint32(s) in which there are some global definitions to help cast the data to the corrrect type/value:
https://github.com/vertgenlab/gonomics/blob/bfa0047ed97a0049100154a42cced8c8b027de6c/cigar/cigar.go#L34-L48

sweavs111 (approved):
great!

cigar/cigar.go
```diff
@@ -44,49 +81,42 @@ func NumDeletions(input []Cigar) int {
...
-			currentNumber = currentNumber + fmt.Sprintf("%c", v)
-		} else if validOp(v) {
-			currentCigar := Cigar{RunLength: parse.StringToInt(currentNumber), Op: v}
-			output = append(output, currentCigar)
-			currentNumber = ""
-		} else {
-			log.Panicf("Invalid character: %c", v)
+		return make([]Cigar, 0) // Unmapped returns a cigar slice of zero
+	}
+	var ans []Cigar = make([]Cigar, 0, 1)
```
dasnellings, line 105: Nice, I think cap == 1 is the right call.

dasnellings (approved):
Cigar is much cleaner this way. One small comment, but LGTM!

PR #1063, opened 2024-08-05
generateBedRegion as three functions

saqzhao:
# Description
generateBedRegion is split up into three functions to 
1) count viable windows in the bed
2) generate 1 bed
3) continuously send generated regions to a channel.

🐛 Bug Report

### Testing

Uses currently existing test cases for simulateBed cmd


simulate/bed.go
```diff
@@ -5,39 +5,63 @@ import (
...
+		} else {
+			return bed.Bed{
+				Chrom: searchSpace[j].Chrom, 
+				ChromStart: searchSpace[j].ChromStart + tmp - 1, 
+				ChromEnd: searchSpace[j].ChromStart + tmp - 1 + regionLength, 
+				Name: searchSpace[j].Name,
+				FieldsInitialized: 4}, true
+		}
+	}
+	// TODO: I want it to fatal error if it can't generate answer, but when I had it as a return in the if-else above, it kept on giving "no return" error
```
dasnellings, line 47: Go functions must end in a return if they are declared as having return variables. In a case like this where you actually want to fatal (or panic, which I would recommend here) you just put the fatal/panic before the return statement. The return is then unreachable, but allows you to compile the function correctly. 

Example:
```
     // stuff above
     log.Panic("error message here")
     return bed.Bed{}, false
}
```

dasnellings (commented):
couple style comments, but otherwise looks good!

simulate/bed.go
```diff
@@ -5,39 +5,63 @@ import (
 	"github.com/vertgenlab/gonomics/numbers"
 )
 
+// CountWindows counts the total viable windows of length regionLength in the sequence searchSpace
+func CountWindows(searchSpace []bed.Bed, regionLength int) int {
+	var length, totalWindows int
+	for i := 0; i < len(searchSpace); i++ {
```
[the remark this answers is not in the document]
saqzhao: Thank you for the suggestion! I applied it to the other loops as well.

simulate/bed.go
```diff
@@ -5,39 +5,63 @@ import (
...
+		if length >= regionLength {
+			totalWindows = totalWindows + (length - regionLength + 1)
+		}
+	}
+
+	return totalWindows
+}
+
+// GenerateBedRegion searches the regions of searchSpace (a noGap.bed input file, as a parsed struct) and randomly selects a continuous region of length regionLength 
+func GenerateBedRegion(searchSpace []bed.Bed, tmp int, regionLength int) (bed.Bed, bool) {
```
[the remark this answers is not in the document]
saqzhao: This has been renamed to randPos for clarity.

simulate/bed.go
```diff
@@ -5,39 +5,63 @@ import (
...
+		length = searchSpace[j].ChromEnd - searchSpace[j].ChromStart
+
+		// check that the window can fit a sequence of that length e.g. want 10 bp, but window is 5bp
+		chromWindows = length - regionLength + 1
+
+		if chromWindows < 1 {
+			continue
+		}
+		if tmp-chromWindows > 0 {
+			tmp -= chromWindows
```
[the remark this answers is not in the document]
saqzhao: The random number is the overall position in the entire input bed, and we want to get the position specific to that region.

simulate/bed.go
saqzhao: Thanks for letting me know, I've added a log.Panic.

simulate/bed.go
```diff
@@ -5,39 +5,63 @@ import (
...
 )
 
+// CountWindows counts the total viable windows of length regionLength in the sequence searchSpace
+func CountWindows(searchSpace []bed.Bed, regionLength int) int {
+	var length, totalWindows int
+	for i := 0; i < len(searchSpace); i++ {
+		length = searchSpace[i].ChromEnd - searchSpace[i].ChromStart
+
+		if length >= regionLength {
+			totalWindows = totalWindows + (length - regionLength + 1)
```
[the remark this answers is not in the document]
saqzhao: Got it, thanks!

dasnellings (dismissed):

edotau (dismissed):
lgtm!

In the future it would be good to write a few quick and small tests for when writing out small functions. For example, `CountWindows()` could be easily tested by defining a Bed struct followed by a region size to see if it is indeed performing as expected. (See below for a small example):

```golang
// Only serves as example. Actual values for testing could vary
func TestCountWindows(t *testing.T) {
	expected := NUMBER
	windows := []Bed{
		{Chrom: "chr1", ChromStart: 0, ChromEnd: 50},
	},
	length := 50
	result := CountWindows(windows, length)
	if result != expected {
		t.Errorf("Error: CountWindows() = %d; expected %d", result, expected)
	}
}
```

This is especially important when working with more complex functions that involve memory management via pointers.

simulate/bed.go
```diff
@@ -1,43 +1,71 @@
 package simulate
 
 import (
+	"log"
 	"github.com/vertgenlab/gonomics/bed"
 	"github.com/vertgenlab/gonomics/numbers"
 )
 
+// CountWindows counts the total viable windows of length regionLength in the sequence searchSpace
+func CountWindows(searchSpace []bed.Bed, regionLength int) int {
```
RavenYL, line 10: note to self: regionLength is windowSize

simulate/bed.go
```diff
@@ -1,43 +1,71 @@
...
+
+// GenerateBedRegion searches the regions of searchSpace (a noGap.bed input file, as a parsed struct) and randomly selects a continuous region of length regionLength 
+func GenerateBedRegion(searchSpace []bed.Bed, randPos int, regionLength int) (bed.Bed, bool) {
+	var chromWindows int
+	var length int
+	// iterating through each ungapped window
+	for j := range searchSpace {
+		length = searchSpace[j].ChromEnd - searchSpace[j].ChromStart
+
+		// check that the window can fit a sequence of that length e.g. want 10 bp, but window is 5bp
```
RavenYL, line 31: note to self: window here means searchSpace length

simulate/bed.go
```diff
@@ -1,43 +1,71 @@
...
+		if length >= regionLength {
+			totalWindows += length - regionLength + 1
+		}
+	}
+
+	return totalWindows
+}
+
+// GenerateBedRegion searches the regions of searchSpace (a noGap.bed input file, as a parsed struct) and randomly selects a continuous region of length regionLength 
+func GenerateBedRegion(searchSpace []bed.Bed, randPos int, regionLength int) (bed.Bed, bool) {
```
craiglowe, line 24: What is randPos?  From the description, it seems like only searchSpace and regionLength would be needed as inputs?

craiglowe: I see now that the programmer first needs to needs to get a random number between 0 and CountWindows and pass this as randPos.  I think a description of the allowable range for randPos is all that is needed.

simulate/bed.go
```diff
@@ -1,43 +1,71 @@
...
+
+		// check that the window can fit a sequence of that length e.g. want 10 bp, but window is 5bp
+		chromWindows = length - regionLength + 1
+
+		if chromWindows < 1 {
+			continue
+		}
+
+		// Decrement randomly generated overall position until it fits within the region
+		if randPos > chromWindows {
```
RavenYL, line 39: note to self: decrement here means to subtract previous searchSpace windows.

Example: the searchSpace consists of 3 regions in the genome, chr1 1 11, chr2 5 15, chr3 2 12. Say there are 6 windows in total in this searchSpace, 2 in each of the 3 regions, randPos is generated as an int between 0 and 6, and if randPos is 5, then it will look in region 1 (windows 1 and 2 are here), region 2 (windows 3 and 4 are here), and region 3 (windows 5 and 6 are here), land in region 3, 5-4=1 (line 40 randPos -= chromWindows), report window 5 out of all 6 windows

RavenYL (dismissed):

simulate/bed.go
RavenYL: Actually I agree with this, I think the randPos should be chosen from either [0, totalWindowNumber-1] or [1, totalWindowNumber]

saqzhao: Got it! I added a comment describing randPos to the GeneratebedRegion function. Would you mind quickly reviewing again?

craiglowe: I apologize for the delay.  Should it be [0, totalWindows-1]?

saqzhao: Yes, thanks for catching!

rimangan (approved):

edotau (approved):

craiglowe (approved):
Great!

PR #1066, opened 2024-08-14
Added chromName argument to multiFaSequenceSwap

nthoduguli:
Two small changes to the multiFaSequenceSwap program: 
 - added an argument called "chromName", asking for the desired chromosome of the swap regions; if the swap region in the bed file isn't on that chromosome, it gets skipped 
 - removed the OutSeqName argument, instead standardized the name of the swapped sequence appended to the multiFa to always be "backgroundName.swapped"

In line with these changes, I edited the previous test cases and added a third to test the skipping of non-desired swap regions not on the provided chromosome.


rimangan (approved):
LGTM!

hcnapier (approved):
Looks great! Leaving this unmerged so @jayashabarishankar has a chance to look it over as well.

PR #1067, opened 2024-08-21
PR adding the L50 statistic to assemblyStats.go + misc testing files

sweavs111:
This PR adds the L50 statistic to the assemblyStats.go function. The L50 is the number of contigs that make up the N50 statistic. 
Screenshot 2024-08-21 at 3 07 10 PM

Additionally, I added a missing test file for assemblyStats + test files for fasta/assemblyStats.go

### DIFF
--- a/fasta/assemblyStats.go
+++ b/fasta/assemblyStats.go
@@ -12,37 +12,48 @@ import (
 // AssemblyStats takes the path to a fasta file and a flag for whether lower case letters
 // should count as assembly gaps.  Five ints are returned, which encode:
 // the N50 size, half the size of the genome, size of the genome, size of the largest contig, and the number of contigs.
-func AssemblyStats(infile string, countLowerAsGaps bool) (int, int, int, int, int) {
+func AssemblyStats(infile string, countLowerAsGaps bool) (int, int, int, int, int, int) {
 	records := Read(infile)
-	var genomeLength int
 
 	contigList := MakeContigList(records, countLowerAsGaps)
 
-	for i := 0; i < len(contigList); i++ {
-		genomeLength += contigList[i]
-	}
+	genomeLength := calculateGenomeLength(contigList)
+
 	sort.Ints(contigList)
 	halfGenome := genomeLength / 2
 
-	N50 := CalculateN50(contigList, halfGenome)
+	N50, L50 := CalculateN50L50(contigList, halfGenome)
 	numContigs := len(contigList)
 	largestContig := contigList[len(contigList)-1]
 
-	return N50, halfGenome, genomeLength, largestContig, numContigs
+	return N50, L50, halfGenome, genomeLength, largestContig, numContigs
 }
 
-// CalculateN50 takes a slice of contig lengths and the size of half
-// the genome.  It returns the N50 size.
-func CalculateN50(contigList []int, halfGenome int) int {
+// calculateGenomeLength is a helper function that takes in a slice of int representing contigs in an assembly and returns the genome size (int)
+func calculateGenomeLength(contigList []int) int {
+	if len(contigList) == 0 {
+		log.Fatalf("Error: Cannot calculate genome length -- contig list is empty")
+	}
+	var genomeLength int
+	for i := 0; i < len(contigList); i++ {
+		genomeLength += contigList[i]
+	}
+	return genomeLength
+}
+
+// CalculateN50L50 takes a slice of contig lengths and the size of half the genome. It returns the N50 size and L50 size.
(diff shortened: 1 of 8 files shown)


fasta/assemblyStats.go
```diff
@@ -114,6 +125,8 @@ func WriteAssemblyStats(assemblyName string, outfile string, N50 int, halfGenome
 	exception.FatalOnErr(err)
 	_, err = fmt.Fprintf(file, "N50: %d\n", N50)
 	exception.FatalOnErr(err)
+	_, err = fmt.Fprintf(file, "L50: %d\n", L50)
```
edotau, line 128: What do you think about splitting the N50 and L50 (s) into two functions?

fasta/assemblyStats.go
```diff
@@ -12,37 +12,48 @@ import (
...
+	}
+	var genomeLength int
+	for i := 0; i < len(contigList); i++ {
+		genomeLength += contigList[i]
+	}
+	return genomeLength
+}
+
+// CalculateN50L50 takes a slice of contig lengths and the size of half the genome. It returns the N50 size and L50 size.
+func CalculateN50L50(contigList []int, halfGenome int) (int, int) {
```
edotau, line 45: What do you think about splitting the N50 and L50 (s) into two functions?

sweavs111: Yes its a good question, and I agree that it isn't very elegant. I opted to have them both come from the same function because the L50 is baked into the N50 function (essentially just counting how many times the loop was completed). If we broke them into two different functions we would sort of just be doing the same calculation and running the same code two times. Not sold on that though, happy to hear other opinions on this.

hcnapier: I agree with the elegance point as well. I think for me it comes down to how frequently you're going to be using both of these statistics together. If you're going to want both of them most of the time anyway, then it feels slightly more efficient to keep them together. If you're expecting to want either N50 or L50 most of the time, then I'm in favor of splitting them up.

fasta/assemblyStats.go
```diff
@@ -12,37 +12,48 @@ import (
...
-	var genomeLength int
 
 	contigList := MakeContigList(records, countLowerAsGaps)
 
-	for i := 0; i < len(contigList); i++ {
-		genomeLength += contigList[i]
-	}
+	genomeLength := calculateGenomeLength(contigList)
+
 	sort.Ints(contigList)
```
RavenYL, line 22: contigList is sorted here when AssemblyStats is called, before CalculateN50L50 is called. Do you want to update the usage of CalculateN50L50 to make sure that if CalculateN50L50 is called outside of AssemblyStats, they know to sort their contigList beforehand?

fasta/assemblyStats.go
hcnapier: We talked about this in code review today and decided that keeping them together probably makes sense unless there's a big efficiency loss involved in calculating the L50 along with the N50 (which seems unlikely)

fasta/assemblyStats.go
sweavs111: Yes I agree that it is important for people to know that. I believe it already say it: "`// CalculateN50L50 takes a sorted slice of contig lengths and the size of half the genome. It returns the N50 size and L50 size.`"

fasta/assemblyStats_test.go
```diff
@@ -29,3 +32,32 @@ func ContigAllAreEqual(input []int, expected []int) bool {
...
+	exp := []int{30, 1}
+	input := MakeContigList(inputFasta, true)
+	sort.Ints(input)
+	N50, L50 := CalculateN50L50(input, calculateGenomeLength(input)/2)
+	if N50 != exp[0] || L50 != exp[1] {
+		t.Errorf("Error in CalculateN50L50: Expected N50: %d L50: %d. Observed N50: %d L50: %d", exp[0], exp[1], N50, L50)
+	}
+}
+
+func TestWriteAssemblyStats(t *testing.T) {
```
edotau, line 53: I think the code would be better organized and easier to test in the future if the fasta package contained single functions to calculate the assembly stats (i.e CalcN50, CalcL50, etc...) and the WriteAssemblyStats function should be moved into the cmd and modified there depending on user needs. The fasta package should contain structs and functions that use the fasta struct and I feel like the WriteAssembly function breaks that pattern

edotau: Would be good to add a TODO at the very least since the `WriteAssemblyStats()` function was preexisting

craiglowe (approved):
Great!

hcnapier (approved):

PR #1068, opened 2024-08-23
Motif match comp filter

RavenYL:
# Description
Because motifMatchComp is not currently designed to handle motifs with length > 32, I added a filtering step to motifMatchComp to notify the user when they input motifs with length > 32 and remove those motifs from downstream work. This replaces the fatal error when encountering a motif with length > 32

### Testing
Tests still pass

### DIFF
--- a/motif/matchComp.go
+++ b/motif/matchComp.go
@@ -36,22 +36,31 @@ func MatchComp(s MatchCompSettings) {
 	var couldScoreConsensus bool
 	var revCompMotif PositionMatrix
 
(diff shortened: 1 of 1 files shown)


edotau (dismissed):

hcnapier (dismissed):

RavenYL:
TODO: change this from auto-filter to an option in the cmd

rimangan (approved):
Nice change!

hcnapier (approved):

edotau (approved):

PR #1070, opened 2024-08-29
MultiFaToChain - multiWay

rimangan:
# Description
This PR makes a small adjustment to MultiFaToChain to allow for an arbitrary number of sequences in the input fasta file. Previously, MultiFaToChain required a pairwise multiFa and would define a chain between these two sequences. Now, a user may specify a query sequence from any of the sequences in a multiFa. As the first sequence of a multiFa defines the coordinate system, note that the first sequence in the multiFa must still be the target in the output chain (unless swapTandQ is used). 

### Testing

None


cmd/multiFaToChain/multiFaToChain.go
```diff
@@ -21,21 +21,32 @@ const (
...
 
-func multiFaToChain(inFile string, tName string, qName string, outFile string, swapTandQ bool) {
-	var records []fasta.Fasta = fasta.Read(inFile)
+type Settings struct {
+	InFile       string
+	TName        string
+	QName        string
+	OutFile      string
+	SwapTandQ    bool
+	QuerySeqName string
```
saqzhao, line 30: note to myself: to specify which infile seq

saqzhao (approved):

edotau (approved):

PR #1071, opened 2024-09-10
tfMatchComp fix refStart and handle unscorable sequences

RavenYL:
# Description
- fix refStart (🐛 Bug Report): tfMatchComp operates (scans TFBS, scores matchiness, and compares between the 2 species) using the multiFa input, without knowing the refStart position (without knowing where on the chromosome the multiFa came from, which would be represented by the reference species's coordinate of the starting position of the multiFa alignment). Therefore, I changed the code to make sure that refStart is is only used as an externally-provided constant to add to the ChromStart and ChromEnd calculated from the input multiFa when writing output bed regions.
- handle unscorable sequences: I've run into some cases where the sequence could not be scored. **How should this be handled? I propose 2 options. The code currently uses Option 1**
     - Option 1: when !couldScoreSequence, minResidualRefScore and minResidual remain as initialized values, -Inf and +Inf, indicating "N/A", can be filtered out / ignored by user in output
     - Option 2: when !couldScoreSequence, treat score as 0, calculate residual
- For reference, the original tfMatchComp PR is https://github.com/vertgenlab/gonomics/pull/923, which contains a link to the slides explaining the algorithm: https://docs.google.com/presentation/d/1Y9odAFzi8bL244BqeITUUHFX3d87duuEZIElRUAUA_Q/edit#slide=id.g1cf9aa157ba_0_311


edotau:
- PR report looks great! It makes me happy that you are putting more work into doing a good PR write up.  It definitely helps with the review process.

- I also noticed that you are also running into an issue i've been seeing in the gitaction build. Is this related to what you are trying to fix?

Screenshot 2024-09-11 at 7 32 36 AM

rimangan (approved):

rimangan (changes requested):
Gitactions failing

edotau (approved):

rimangan (approved):

PR #1072, opened 2024-09-11
MultiFaToVcf - Fixing substitutionsOnly in 3-way

rimangan:
# Description
🐛 Bug Report Small bug fix: multiFaToVcf, when fed a 3-way alignment, is supposed to only consider substitutions. A check was missing for gaps, which led to invalid VCF outputs with '-' in the Ref or Alt fields. Fixed.

### Testing

None

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -208,7 +208,9 @@ func ThreeWayFaToVcf(f []fasta.Fasta, chr string, out *fileio.EasyWriter) {
 	}
 
 	for i := range f[0].Seq {
-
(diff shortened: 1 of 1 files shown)


saqzhao (approved):

craiglowe (approved):
This looks good!
I had a little trouble understanding the "... with the first entry as the reference and the last two fasta entries as the alt alleles."  Maybe it could say, "... with the first entry as the reference and the last two fasta entries as the alleles appearing in the vcf file."  The part that tripped me up is that the last two fasta files don't need to be alt alleles, but one of them can also be a reference allele.

PR #1075, opened 2024-09-15
samInfo Reorganization

rimangan:
# Description
This PR reorganizes the cmd 'samCoverage' into the broader 'samInfo' supercommand. Additionally, it adds a subcommand to make histograms of read lengths in an input SAM/BAM file.

### Testing

None


nthoduguli (approved):
Looks good to me!

cmd/samInfo/readLength_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"testing"
+)
+
+var ReadLengthTests = []struct {
+	InFile       string
+	OutFile      string
+	ExpectedFile string
+}{
+	{InFile: "testdata/readLength/small.sam",
+		OutFile:      "testdata/readLength/test.readLength.txt",
```
edotau, line 14: Could you try to tests a few of these function directly instead of creating additional expected test files?
For example, you could write  a test called
```go
func TestThresholdCalc(threshold, covHistogram, totalCount) {
    expected := 10 // or any manual value
    number := ThresholdCalc()
    if number != expected {
        t.log("error with ThresholdCalc()")   
    }
}
`
and assert for equality by creating 1 (or more) values. Writing more "meaningful" tests will reduce the number of in line annotations from CodeCov and make future refactoring efforts more seemless if all we have to do is change test values father than an arrary of expected test files

File: cmd/samInfo/readLength_test.go [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+
+func TestReadLength(t *testing.T) {
+	var s ReadLengthSettings
+	for _, v := range ReadLengthTests {
+		s = ReadLengthSettings{
+			InFile:  v.InFile,
+			OutFile: v.OutFile,
+		}
+		readLength(s)
+		if !fileio.AreEqual(v.OutFile, v.ExpectedFile) {
```
edotau, line 27: Directly test the read length functionality rather than reading in expected test files

edotau (changes requested):
Can you try to write a few tests that directly tests some of the small functions and remove and reduce the number of expected test files added into the repo?

saqzhao (approved):

PR #1077, opened 2024-09-25
net struct with rw functions

sweavs111:
This PR introduces the [net](https://genome.ucsc.edu/goldenPath/help/net.html) struct. I placed it as a subdirectory of the chain directory, since nets are build off of chains. 

Broadly, net files represent alignments (often whole genome alignments between individuals or species). The net file consists of a "net" header line with chrom name and size. Below that are series of "fill" and "gap" lines that describe the alignment on that chromosome. A "fill" is block of alignment, and the "gap" lines that are indented below the fill line are gaps in that block of alignment. If a gap is large enough, and non-syntenic sequence aligns there, a new "fill" line denotes that there is sequence that aligns there. This new "fill" line will be indented from the previous fill and gap lines to represent that this is a second level alignment that supplements the first level alignment. There can be a tertiary fill to the second level and so on...
Once the whole chromosome has been described as fills or gaps, there is a new "net" header line with the next chromosome information. 

There are 7 required fields, and the struct consists of these fields. In addition to these, the struct also has target chromosome name, which is normally specified by a header line, and the level of net, which is denoted by levels of indentation in the net file. Additionally there are non-required fields in the net file format, and those are kept as a single string. 

One choice I made was to make "gap" lines as the same level as the "fill" field they are referring to, even though they are indented one space from their parent fill field. I felt it was more clear that way.

I have created reading and writing functions for nets, methods for compatibility with the interval interface, and all tests are passing.

Screenshot 2024-09-26 at 11 14 27 AM

Screenshot 2024-09-26 at 11 14 44 AM


chain/net/net.go
```diff
@@ -0,0 +1,176 @@
...
+		ExtraFields: strings.Join(newData[7:], " "),
+		spaces:      currLvl[0][1],
+	}
+
+	return curr, currTName, currLvl
+}
+
+// determineLevel is a helper struct for the NewNet function. It parses how many spaces proceed the data on the line
+// of the file to determine the Level of the net. It also returns the data without the proceeding spaces for easy parsing
+func determineLevel(data []string, currLvl [][]int) ([][]int, []string) {
```
edotau, line 86: I would consider creating a new struct to track the levels:
- but that may require a lot of refactoring so I'll leave that decision to you. 

```golang
type LevelInfo struct {
    Level  int
    Spaces int
    LvlKey []int
}
```

chain/net/net.go
```diff
@@ -0,0 +1,176 @@
...
+	file := fileio.EasyCreate(outfile)
+	for i := range nets {
+		currChrom = nets[i].TName
+		if currChrom != prevChrom {
+			fileio.WriteToFileHandle(file, fmt.Sprintf("net %s %d", currChrom, chromSizes[currChrom].Size))
+		}
+		fileio.WriteToFileHandle(file, ToString(nets[i], true))
+		prevChrom = currChrom
+	}
+	err := file.Close()
```
edotau, line 132: I'd just wrap the close statement in the exception error handling helper function so you dont need to declare a new err variable to only be used once.

```suggestion
	exception.PanicOnErr(file.Close())
```

chain/net/net.go
```diff
@@ -0,0 +1,176 @@
...
+	for i := range data {
+		if data[i] == "" {
+			spaces++
+		} else {
+			break
+		}
+	}
+	data = data[spaces:]
+
+	if spaces == currLvl[0][1] {
```
edotau, line 97:
```golang
switch {
    case spaces == currLvl[0][1]:
        (same logic)
    case spaces > currLvl[0][1]:
        (same logic)
    case spaces < currLvl.Spaces:
        (same logic)
        switch data[0] {
        case "fill":
            (same logic)
        case "gap":
            if spaces == 0 {
            (same logic)
    default:
        fmt.Errorf("Error: unhandled case in determineLevel...")
    }
    ```

edotau (commented):
Overall pretty good PR. The only thing I would change is the `ToString()` to improve the performance and everything else are only suggestions that i'll leave up to you to decide.

chain/net/net.go
sweavs111: yes i like this a lot. It was one of those things where i thought it was going to be simple, but i kept having to add complexity. Best to make a struct for readability

craiglowe (approved):
Great!

RavenYL (approved):

PR #1078, opened 2024-09-25
cutFinalNBases option for faFilter

sweavs111:
This PR adds a new option for faFilter. -cutFinalNbases will cut a user-specified number of bases from the end of every fasta record in a file. This is useful when you want to cut out a constant region at the end of some fasta records, but each fasta record may have a different numbers of bases, so the -end option isn't suitable for use. Testing files present and passing


hcnapier (approved):

craiglowe (approved):
Great!

PR #1079, opened 2024-10-04
add pad option to bedMerge

sweavs111:
This PR adds a new option to bedMerge. The 'pad' option will merge bed entries if they are a certain distance away from each other. It essentially extends the length over which `-mergeAdjacent` works over. This option is equivalent to the `-d` option with the `bedtools mergeBed`, however, has the benefit that it keeps the names, scores, and extra bed fields, which the bedtools cmd cannot
Screenshot 2024-10-04 at 2 24 07 PM

To accomplish this I had to refactor the program a bit. Instead of mergeAdjacent working in the code as a bool, it is now an integer. This integer represents the distance is the "pad" value. If the -mergeAdjacent option is used, it is equivalent in the code to a pad value of 0. I use the existing function `bed.MinimumDistance` to calculate the distance between records and if that distance is lower than the the pad value, the bed entries will be merged.  Please let me know if this refactor is too confusing/messy.

One caveat is that in the 'bed.MinimumDistance' code, if two entries are directly adjacent, it will return 1, and not zero. Similary, if the `a.chromEnd = 10`, and `b.chromStart = 11`, `bed.MinimumDistance(a, b)` returns 2. To handle this, under the hood, I add one to the input pad value. Turning the input pad value from 0 base to 1 base. I would appreciate an extra pair of eyes to make sure this makes sense, and I am not overlooking any corner cases. Thanks!

### DIFF
--- a/bed/modify.go
+++ b/bed/modify.go
@@ -82,8 +82,10 @@ func MergeLowMem(b <- chan Bed, mergeAdjacent bool) <- chan Bed {
 
 // MergeHighMem retains input Bed entries that are non-overlapping with other input bed entries and merges together overlapping bed entries.
 // Merged bed entries will retain the maximum score in the output.
-func MergeHighMem(records []Bed, mergeAdjacent bool, keepAllNames bool) []Bed {
+func MergeHighMem(records []Bed, mergeAdjacent int, keepAllNames bool) []Bed {
 	var outList []Bed
+	var minDist int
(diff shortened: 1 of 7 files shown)


edotau (dismissed):

craiglowe (approved):
Great!

hcnapier (approved):

PR #1081, opened 2024-10-09
multiFaCount

RavenYL:
# Description
- Scan multiple Fasta alignment for a user-specified pattern ('present bases (A,C,G,T, not gap- or N)' for now) and report counts.
- Both mode: Count the number of positions where 2 sequences both match the user-specified pattern

### DIFF
--- a/cmd/multiFaCount/multiFaCount.go
+++ b/cmd/multiFaCount/multiFaCount.go
@@ -66,7 +66,6 @@ func multiFaCount(s Settings) {
 func usage() {
 	fmt.Print(
 		"multiFaCount - Scan multiple Fasta alignment for a user-specified pattern ('present bases (A,C,G,T, not gap- or N)' for now) and report counts.\n" +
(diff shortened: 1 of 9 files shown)


cmd/multiFaCount/multiFaCount.go
```diff
@@ -0,0 +1,111 @@
...
+		QueryName:       flag.Arg(0),
+		Both:            *both,
+		Either:          *either,
+		SecondQueryName: *secondQueryName,
+		ThirdQueryName:  *thirdQueryName,
+	}
+
+	multiFaCount(s)
+
+	// Caveat: if QueryName or SecondQueryName is misspelled, will be reference sequence by default
```
RavenYL, line 106: Raven to fix: multiFa.go, find query index, "return nameIndexMap[queryName]", change to get 2 variables (whether or not found, actual index), and if not found, error out, if found, then return index

hcnapier (approved):

craiglowe (approved):
Great!

PR #1083, opened 2024-10-18
ilsReconstructSeq

saqzhao:
# Description

[TRAILS](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1010836) is a HMM that infers incomplete lineage sorting and estimates topologies that incorporate this information, as well as the probability for each topology.

Using these TRAILS-produced topologies, we can use reconstructSeq to produce pFASTAs at the desired node for each topology. ```ilsReconstructSeq``` is a command-line function that takes in these posterior probabilities calculated by TRAILS and weights the corresponding pFASTA, to give a more accurate estimate of the reconstructed node that accounts for incomplete lineage sorting.

Additional tools added to pFasta and pDNA:
```pDna.SumsToOne``` checks that a pDNA base sums to 1.
```pDna.RandBase``` randomly generates a pDNA base that sums to 1.
```pDna.Scale``` takes in a multiplier and pDNA base, and scales each probability in the base by the multiplier.
```pDna.Sum``` takes in two pDNA bases and adds their values across the 
```pFasta.RandSeq``` randomly generates a pFasta sequence, where each base sums to 1.

### Testing

Tests were added for all added tools using toy data.


rimangan (approved):

nthoduguli (approved):
Looks good to me, Sarah!

PR #1084, opened 2024-10-18
overlapProportionRecursive for interval package

sweavs111:
Short PR introducing a function for the interval package. It takes in two intervals and returns true if they overlap each other recursively by a certain input proportions.


saqzhao (dismissed):

hcnapier (approved):

craiglowe (approved):
Great!

PR #1088, opened 2024-11-22
Add bam read groups based on read tags

dasnellings:
Relatively simple new command to annotate bam files with read groups corresponding to read tags. For context, my use case is for annotating 10x scATAC bam files with read groups corresponding to cell barcodes so that `bcftools mpileup` will treat each cell as a different sample and generate per-cell pileup information.


cmd/bamTagToReadGroup/bamTagToReadGroup.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func safeclose(f io.Closer) {
+	err := f.Close()
+	exception.PanicOnErr(err)
+}
+
+// addTagsToHeader adds each element in tagValues as a read group in the header
+// format is `@RG	ID:$TAGVALUE	SM:$TAGVALUE	LB:$TAGVALUE`
+func addTagsToHeader(header *sam.Header, tagValues []string) {
```
edotau, line 100: Seems a bit cumbersome, but I think using string builder to concat strings would in theory is more performant.

```go
import(
    "strings"
)

func addTagsToHeader(header *sam.Header, tagValues []string) {
    var str strings.Builder

    // Iterate over existing header lines
    for _, line := range header.Text() {
        if !strings.HasPrefix(line, "@RG") {
            str.WriteString(line)
            str.WriteString("\n")
        }
    }

    // Add new @RG lines
    for _, value := range tagValues {
        str.WriteString("@RG\tID:")
        str.WriteString(value)
        str.WriteString("\tSM:")
        str.WriteString(value)
        str.WriteString("\tLB:")
        str.WriteString(value)
        str.WriteString("\n")
    }

    // Update the header text
    header.SetText(str.String())
}
```

edotau: I didn't perform an checks or test with this suggested block so be sure to check if it's right

cmd/bamTagToReadGroup/bamTagToReadGroup.go
```diff
@@ -0,0 +1,141 @@
...
+		writeChan <- rec
+	}
+	close(writeChan)
+
+	// wait for goroutine to finish writing before return on main thread
+	wg.Wait()
+}
+
+func safeclose(f io.Closer) {
+	err := f.Close()
```
edotau, line 94: Not sure if it makes a difference, but would it make sense to handle the panic exception directly without creating a new err variable?

```suggestion
	exception.PanicOnErr(f.Close())
```

cmd/bamTagToReadGroup/bamTagToReadGroup.go
```diff
@@ -0,0 +1,141 @@
...
+	flag.Usage = usage
+
+	if *infile == "" {
+		usage()
+		log.Fatalln("ERROR: must provide a bam file with -i")
+	}
+
+	if *tagId == "" || len(*tagId) != 2 {
+		usage()
+		log.Fatalln("ERROR: -tagId must be 2 characters")
```
edotau, line 45: Style consistency:

```suggestion
		log.Fatalln("Error: -tagId must be 2 characters")
```

cmd/bamTagToReadGroup/bamTagToReadGroup.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if *tagValuesFile == "" {
+		usage()
+		log.Fatalln("ERROR: must provide a tag values file with -tagValues")
```
edotau, line 50:
```suggestion
		log.Fatalln("Error: must provide a tag values file with -tagValues")
```

cmd/bamTagToReadGroup/bamTagToReadGroup.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		if strings.HasPrefix(splitExtra[i], tagId+":") {
+			tagValue = splitExtra[i][5:] // skips characters defining tag (e.g. CB:Z:TagValue)
+		}
+	}
+
+	// add read group if tag value is present in input values file
+	if tagValueMap[tagValue] {
+		newExtra = append(newExtra, fmt.Sprintf("RG:Z:%s", tagValue))
```
edotau, line 138: Maybe it make sense to refactor the `addTagsToHeader` to something like `CreateTag` or `AddTag` to return a string representation of the new tag w/ its value so you can reuse the same code for both the header and updating the records.

edotau (commented):
Something to think about is I felt some of these functions could be organized into the sam package sense they could be generalized for tag mutation, but i'll leave that decision up to you

cmd/bamTagToReadGroup/bamTagToReadGroup.go
dasnellings: Yeah it would. Committed.

cmd/bamTagToReadGroup/bamTagToReadGroup.go
dasnellings: It would be more performant, but the computational expense for processing the header is essentially negligible compared to processing of the bam records. In this case, I think it makes sense to keep the simpler and slightly worse implementation IMO.

cmd/bamTagToReadGroup/bamTagToReadGroup.go
dasnellings: Ah good catch. I am an all caps boy myself, but I'll commit all these edits

cmd/bamTagToReadGroup/bamTagToReadGroup.go
dasnellings: Agreed. Playing with tags is a bit error-prone when starting from bams since changes to the text wont propagate to the output unless you run ParseExtra first. Will probably save us some headaches in the future by having builtin functions for this. I am going to implement these now.

edotau (approved):

cmd/bamTagToReadGroup/bamTagToReadGroup.go
```diff
@@ -0,0 +1,136 @@
...
+		newHeaderText = append(newHeaderText, fmt.Sprintf("@RG\tID:%s\tSM:%s\tLB:%s", tagValues[i], tagValues[i], tagValues[i]))
+	}
+	header.Text = newHeaderText
+}
+
+func updateRecord(rec *sam.Sam, tagId string, tagValueMap map[string]bool) {
+	var err error
+
+	// remove existing read group tag if present
+	err = sam.RemoveTag(rec, "RG")
```
sweavs111, line 117: I may be missing something, but how could the sam record already have a tag value in the struct if it was just read from a file?

sweavs111 (approved):
Hey Dan, sorry it took me so long to get to this. I am approving because the code looks good I believe and it seems to have ample testing, and I don't want to cause even more of a delay to you. Am I understanding correctly that this doesn't actually filter the bam file, it just adds read-group tags to to reads (in your test case, cell barcodes) for later filtering? 

I additionally had some comments on the usage statement to make things more clear that you can chose to add or not

PR #1090, opened 2024-12-13
faFindFast add gcContent option

RavenYL:
# Description
- Added gcContent option to faFindFast
- Note that this option is not currently written to be used together with other options that print annotation fields to the output bed file
- In the past, I have generated files using commit "[added test files](https://github.com/vertgenlab/gonomics/pull/1090/commits/1fff0c480d406e1d386cbe98a8a0fed7064851c9)", which allowed options like -longOutput and -outputAlnPos to also print gcContent calculations

### Testing
`./faFindFast -gcContent -chrom chr1 -windowSize 5 -firstQueryName Human -secondQueryName Chimp testdata/test_gc.fa testdata/expected_gc.bed`


rimangan (approved):

craiglowe (approved):
Great!

PR #1094, opened 2025-01-08
Trim OverlapEnrichments

rimangan:
# Description

When calculating overlap enrichments between intervals, we often use background regions to limit our analysis to specific genomic regions. 
With the argument 'trimToRefGenome', we can exclude query intervals from the analysis that are found outside of these background regions. 
On main, this program ignores elements like the one on the right of this figure, which partially overlap the background regions. However, this can be quite restrictive, particularly in the use case where background regions are gene features or open chromatin regions, which may have positional instabilities across biosamples or technical replicates. 
To address this issue, I've implemented a new `relationship` option, which allows the user to specify the interval.Query relationship (currently within by default). If set to 'any', the program will now consider query elements that partially overlap background elements. However, the query region is trimmed so that it is contained within the background elements (Note this was the critical decision here, where there was a lot to consider, curious for others opinions).
image

### Testing

None

### DIFF
--- a/cmd/overlapEnrichments/overlapEnrichments.go
+++ b/cmd/overlapEnrichments/overlapEnrichments.go
@@ -6,6 +6,8 @@ package main
 import (
 	"flag"
 	"fmt"
+	"github.com/vertgenlab/gonomics/bed"
+	"github.com/vertgenlab/gonomics/numbers"
 	"log"
 
 	"github.com/vertgenlab/gonomics/exception"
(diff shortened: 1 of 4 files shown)


RavenYL:
Note to Self:

Current behavior: "within relationship", "??" is "x", that "??" query region is discarded pre overlapEnrichment analysis
This PR behavior: default behavior is same as current behavior, when set to "any relationship", "??" is reduced/trimmed to the part that matches "background", and that part is not discarded, aka included in overlapEnrichment analysis

What are "background regions"? Usually think of no gap whole genome. "query regions" are a strict subset of "background regions", and overlapEnrichment asks if "query regions" is statistically the same as picking random regions from "background regions"

Originally, "background regions" == "reference genome". This is probably why "??" region was discarded (if a region in "query" exceeds chromosome end, then it is not valid, discard)
Now, "background regions" can be "regions of interest like haqers"
"query regions" can be "chromatin state peaks", etc.

RavenYL (approved):
Thank you! I like the trimToRefGenome deprecation message!

hcnapier (approved):
LGTM, I appreciate the thorough usage and descriptive error messages

PR #1096, opened 2025-01-29
Parse Scientific Notation

RavenYL:
# Description
- StringToInt can't handle scientific notation. This caused the reading of some GWAS trait bed files to fail 
- changed StringToInt to add float64 intermediate
- added tests for StringToInt, and now parsing scientific notation works


sweavs111 (dismissed):
looks good to me!

hcnapier (approved):
Looks good, I like the added test!

craiglowe (approved):
Great!

PR #1097, opened 2025-02-02
bedTrim.go

sweavs111:
This command will trim the length of each bed record in a file by N percent (keeping the center of the bed record). 

While I have no current plans to add to this (which is why it is in cmdx) i tried to write the command in a way that people could easily add options to the command, such as trimming fixed numbers of bases from either side, trim from just a single side, shift all regions, etc...).

If it appears that this command would exist better as an option for an existing command, let me know!


hcnapier (approved):
This looks good to me!

ndzikowski (commented):
Looks good to me

ndzikowski (approved):
^

PR #1098, opened 2025-02-17
multiFaToVcf -secondQueryName

RavenYL:
# Description
multiFaToVcf -secondQueryName allows the program to take in a multiFa that has >2 sequences, but still do a pairwise conversion of multiFa to vcf, where the firstQuery sequence is the reference sequence (i.e. the first sequence entry in the multiFa), and the secondQuery sequence is specified by the user by sequence name


sweavs111 (approved):
woohoo!

craiglowe (approved):
Great!

PR #1099, opened 2025-03-03
locateCG.go

ndzikowski:
# Description

locateCG() compares 2 aligned sequences and finds CpG sites that were "gained" or "lost" in the first sequence relative to the second sequence. Currently, it only searches for CpG sites gained/lost due to substitutions, skipping over gaps, Ns, etc. 

The user inputs: 1) a fasta file containing two aligned sequences (usually, this is an entire chromosome) and 2) the chromosome name. (If inputting a sequence smaller than an entire chromosome, the chromosome name can be substituted with anything). 

The command then outputs a file containing the following information for each CpG site that changed: chromosome, reference coordinates, type of change (gain/loss), alignment coordinates, and reference base (in sequence 1)/alternate base (in sequence 2) for the substitution that altered the CpG site. Note that the output file name is formatted automatically based on the provided chromosome name, and that the user does not need to specify an output file name.

This data can be useful for downstream applications such as checking for enrichment within different kinds of genomic elements, overlaps with disease-relevant SNPs, etc. 

### Testing

- [X] Test files available in /testdata. Test returned no errors. 
- [x] Ran command with no errors on hg38/HCA aligned sequences for each chromosome; randomly chose CG site coordinates to check on UCSC genome browser. 

- [X] The command `go fmt` or `make clean` was used on all files included (automatic)


cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,193 @@
...
+		flag.Usage()
+		log.Fatalf("Error: expecting %d arguments, but got %d\n", expectedNumArgs, len(flag.Args()))
+	}
+
+	//args
+	inFa := flag.Arg(0)
+	chrom := flag.Arg(1)
+
+	s := Settings{
+		InFa:      inFa,
```
sweavs111, line 58: This can be simplified and save some lines of code with:
`InFa:    flag.Arg(0)`
and
`ChromName:     flag.Arg(1)`

you could then delete lines 54 and 55

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,193 @@
...
+
+	//run locate() function on provided args; will output a slice of structs
+	locateCG(s)
+}
+
+func locateCG(s Settings) {
+	var err error
+	var output []CGsite
+	//set chromosome name from args
+	chrom := s.ChromName
```
sweavs111, line 70: you technically don't need to do this, and could just input s.ChromName wherever you use the variable chrom below. This would save a line of code uless you're doing this for another reason, readability, etc...

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,193 @@
...
+	//initialize pair of matched reference/alignment positions
+	refStart := 0
+	alnStart := 0
+
+	fmt.Printf("Beginning loop\n")
+	//for length of first sequence
+	for i := 0; i < len(firstSeq)-1; i++ {
+
+		//store dna.Bases
+		f1, f2 := firstSeq[i], firstSeq[i+1]
```
sweavs111, line 99: you will get a minor memory efficiency gain if you initialize the f1 and f2 variable outside of the for loop similar to what you did in lines 67, 68. There are a few other examples of this below

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,193 @@
...
+		s1, s2 := secondSeq[i], secondSeq[i+1]
+
+		if !(dna.DefineBase(f1)) || !(dna.DefineBase(f2)) || !(dna.DefineBase(s1)) || !(dna.DefineBase(s2)) {
+			continue
+		}
+
+		var refBases, altBases string
```
sweavs111, line 106: I would declare this at the top

cmd/locateCG/testdata/expected.txt
```diff
@@ -0,0 +1,4 @@
+Chrom	RefStart	RefEnd	Type	AlnStart	AlnEnd	Ref	Alt
+chr8	4	5	loss	4	5	A	C
+chr8	22	23	gain	22	23	C	A
+chr8	82	83	gain	83	84	C	G
```
sweavs111, line 4: Might be nice add an example where the C and the G are gained/lost. Both bases will show up in the Ref/Alt columns, right?

sweavs111 (commented):
Looks good! I just had a few style comments

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/vertgenlab/gonomics/fasta"
+	"github.com/vertgenlab/gonomics/fileio"
+)
+
+type Settings struct {
+	InFa      string //input fasta file with 2 aligned sequences to compare
+	ChromName string //e.g. chr1
+}
+
+type CGsite struct {
```
RavenYL, line 19: For a future PR, you can consider making a package for cg, or moving some of the structs and functions to a package like dna!

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func locateCG(s Settings) {
+	var err error
+	var output []CGsite
+	var f1, f2, s1, s2 dna.Base
+	var refBases, altBases string
+
+	//read in fasta file
+	f := fasta.Read(s.InFa)
```
RavenYL, line 69: I think you might want a check here to make sure that the user is giving a fasta file with only 2 sequences, otherwise log fatal. Alternatively, you can write in options for specifying -firstQueryName (think about if firstQueryName can only be f[0].Name aka the reference species, so that correct reference species position coordinates can be written) and -secondQueryName, and allow the user to give a fasta file with >2 sequences

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,186 @@
...
+	s := Settings{
+		InFa:      flag.Arg(0),
+		ChromName: flag.Arg(1),
+	}
+
+	//run locate() function on provided args; will output a slice of structs
+	locateCG(s)
+}
+
+func locateCG(s Settings) {
```
RavenYL, line 62: FYI for formatting, we usually put helper functions before func usage() and func main()

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,186 @@
...
+		log.Fatalf("Seq1 and seq2 not equal in length.")
+	}
+	//pre-allocate size of output
+	//output = make([]CGsite, 0, 10000)
+
+	//initialize pair of matched reference/alignment positions
+	refStart := 0
+	alnStart := 0
+
+	fmt.Printf("Beginning loop\n")
```
RavenYL, line 90: Do you want the user to see fmt.Printf statements? If they are just for debugging, you can comment them out before the PR gets merged into main

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,186 @@
...
+			refStart = output[len(output)-1].StartPos
+			//reset the alnStart to the last corresponding alignment position
+			alnStart = output[len(output)-1].AlnStart
+			//this will now be the new ref/aln position pair for conversion
+		}
+	}
+	fmt.Printf("Finished loop\n")
+
+	//create output file name and initialize file, write header line
+	outfileName := s.ChromName + "CGs.txt"
```
RavenYL, line 141: For a future PR, you can consider writing an option for the user to specify -outfileName on the command line!

cmd/locateCG/locateCG.go
```diff
@@ -0,0 +1,186 @@
...
+
+type Settings struct {
+	InFa      string //input fasta file with 2 aligned sequences to compare
+	ChromName string //e.g. chr1
+}
+
+type CGsite struct {
+	Chrom    string //derived from ChromName
+	StartPos int    //start pos of CG site in reference genome of sequence 1
+	EndPos   int    //end pos of CG site in reference genome of sequence 1
```
RavenYL, line 22: Out of curiosity, is each CGsite always 2 bases (CG), so endPos is always startPos+1, and AlnEnd is always AlnSstart+1? If there are many CGs in a row, they get reported as separate CGsites in the slice []CGsites, and separate lines in the output file, right?

RavenYL (commented):
Great work!!! I have a few minor comments but I am excited to re-review this and get it merged into main!

RavenYL:
This PR reminds me again of how there are 2 types of multi-Fastas, one is "chr1.fa" containing >human, >chimp, >hca etc. which is an alignment file, and the other is "human.fa" containing >chr1, >chr2, >chr3 etc. which is a genome file. Are there "official" names for these?

craiglowe:
It would be good for us to come up with some more specific language around this similarity/difference.  The example where chromosome 1s for a bunch of species are all aligned against each other, would be a "multi-fasta", which is an alignment in fasta format (all sequences will have the same number of characters since it is an alignment).  The example where there is a fasta file of all the human chromosomes, would just be a "fasta" since those are not aligned to each other and it is just a collection of sequences.

cmd/locateCG/locateCG.go
ndzikowski: **Priority. Thank you!

cmd/locateCG/locateCG.go
ndzikowski: Yes, always 2 bases, and consecutive CG sites do get reported in separate lines

RavenYL:
Genome vs Alignment multi-fasta
- length can sometimes distinguish alignment multi-fasta (all species sequences same length) from genome multi-fasta (chr1 and chr2 not the same length)
- note that alignment fasta coordinate system is only valid for reference sequence aka sequence[0]. Query sequences like chimpanzee may have shuffled sequences, gaps, so coordinates not reliable
- We can try to name alignment multi-fasta files ".mfa"

RavenYL (approved):

PR #1100, opened 2025-03-24
Invalid bed region (ChromStart >= ChromEnd)

ndzikowski:
# Description

Adds an error check for bed regions with start = end or start > end coordinates to processBedLine(), the helper function used by our bed reading functions.

This has come up as an issue when running some Gonomics commands that read in bed files but require a region's length to be >=1, such as overlapEnrichments. There is currently no way to alert the user that these "invalid" regions exist in the file.

### Testing

- [X] Tested bed.go, PASS
- [X] Tested all Gonomics commands for backwards compatibility, FAIL

This failed in a few places:

- 2025/03/24 15:15:03 Error: Found region with length <= 0 : chr2 20      20      species1_gap_size0
 Filter out regions with length <= 0 before proceeding.
FAIL    github.com/vertgenlab/gonomics/cmd/globalAlignmentAnchor        0.970s

- 2025/03/24 15:15:04 Error: Found region with length <= 0 : chr1 122157831       122157831       del_eC  0
 Filter out regions with length <= 0 before proceeding.
FAIL    github.com/vertgenlab/gonomics/cmd/mafIndels    0.944s

- 2025/03/24 15:15:07 Error: Found region with length <= 0 : chr14        70711998        70711998        BMP1    70711997        .
 Filter out regions with length <= 0 before proceeding.
FAIL    github.com/vertgenlab/gonomics/cmd/ontologyEnrichment   1.960s

It seems like these are important; we should discuss together at CR. 

- [X] The command `go fmt` or `make clean` was used on all files included

### DIFF
--- a/bed/bed.go
+++ b/bed/bed.go
@@ -99,9 +99,6 @@ func Read(filename string) []Bed {
 
 	for line, doneReading = fileio.EasyNextRealLine(file); !doneReading; line, doneReading = fileio.EasyNextRealLine(file) {
 		current := processBedLine(line)
-		/*if current.ChromEnd >= current.ChromStart {
-			log.Fatalf("Error: Found region with length=0: %s", line)
-		}*/
 		answer = append(answer, current)
 	}
 	err = file.Close()
@@ -115,7 +112,7 @@ func processBedLine(line string) Bed {
 	startNum := parse.StringToInt(words[1])
 	endNum := parse.StringToInt(words[2])
 
-	if endNum >= startNum {
+	if startNum >= endNum {
 		log.Fatalf("Error: Found region with length <= 0 : %s\n Filter out regions with length <= 0 before proceeding.\n", line)
 	}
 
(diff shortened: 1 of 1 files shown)


RavenYL:
Action items for the first test failure:
1. globalAlignmentAnchor "step 2" code remove writing and reading bed, just pass bed variable around internally in the program
2. change bed to slice of pointer to bed, use nil instead of start==end to indicate gap_size_0 aka illegal entry
3. if debugging takes >1hr, then comment out test that fails, make github issue

RavenYL:
Update on the second test failure:

- 2025/03/24 15:15:04 Error: Found region with length <= 0 : chr1 122157831 122157831 del_eC 0
Filter out regions with length <= 0 before proceeding.
FAIL github.com/vertgenlab/gonomics/cmd/mafIndels 0.944s

PR to use file compare instead of bed compare in _test.go. This avoids using bed.Read on the del_eC entry in the output file

PR #1101, opened 2025-04-11
Sam bed to bases

sweavs111:
function in Sam in which you provide a sam record and a bed record and a []dna.Base is returned corresponding to the sam sequence covered by the bed region. 

If deletions or introns exist in the sam sequence, they will be annotated with gaps in the output (-). If insertions exist in the sam sequence they will be represented with lower-case bases in the output. The length of the output will be the same length as the bed record at a minimum, but can be longer if insertions exist in the sam sequence within the bed interval.

### DIFF
--- a/sam/sequence.go
+++ b/sam/sequence.go
@@ -67,7 +67,6 @@ func SamBedToBases(s Sam, b bed.Bed) []dna.Base {
 			idxStart = idx - (samPos - b.ChromStart)
 
 			//simplest case, grab the sequence and return
-			idxStart = idx - (samPos - b.ChromStart)
 			if samPos >= b.ChromEnd {
 				if s.Cigar[j].Op == cigar.Match {
 					idxEnd = idx - (samPos - b.ChromEnd)


sam/sequence.go
```diff
@@ -0,0 +1,111 @@
...
+	var withinBed bool = false
+	var ans, tmp []dna.Base
+
+	if !within(b, s) {
+		log.Fatalf("Error in SamBedToBases: The bed interval isn't completely within the sam inerval")
+	}
+	samPos := s.GetChromStart()
+
+	for j := range s.Cigar {
+		if s.Cigar[j].Op == cigar.HardClip {
```
hcnapier, line 25: Note to self: skip hard clipped bases

sam/sequence.go
```diff
@@ -0,0 +1,111 @@
...
+			continue
+		}
+		//update sam sequence index and sam sequence position
+		if cigar.ConsumesQuery(s.Cigar[j].Op) {
+			idx += s.Cigar[j].RunLength
```
hcnapier, line 30: Note to self: iterate sam sequence position if consume reference (Match, Deletion, Ns, Equal, Mismatch); iterate sam index position if consume query (Match, Insertion, SoftClip, Equal, Mismatch).

sam/sequence.go
```diff
@@ -0,0 +1,111 @@
...
+		//if sam position is above the bed start value, we need to set the start index of the sequence we want to grab
+		if samPos >= b.ChromStart && !withinBed {
+			withinBed = true
+			//get the start sequence index of the bed region
+			idxStart = idx - (samPos - b.ChromStart)
+
+			//simplest case, grab the sequence and return
+			idxStart = idx - (samPos - b.ChromStart)
+			if samPos >= b.ChromEnd {
+				if s.Cigar[j].Op == cigar.Match {
```
hcnapier, line 72: Note to self: if match, grab sequence; if deletion/ns, add gaps

hcnapier (dismissed):
Looks good!

sam/sequence.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		//if sam position is above the bed start value, we need to set the start index of the sequence we want to grab
+		if samPos >= b.ChromStart && !withinBed {
+			withinBed = true
+			//get the start sequence index of the bed region
+			idxStart = idx - (samPos - b.ChromStart)
+
+			//simplest case, grab the sequence and return
+			idxStart = idx - (samPos - b.ChromStart)
```
ndzikowski, line 70: Just want to confirm - is this a duplicate line?

ndzikowski (commented):
Looks all good to me other than the possible duplicate line in /sam/sequence.go (line 70)

sam/sequence.go
sweavs111: good catch! deleted duplicate line and resubmitted

hcnapier (approved):

PR #1104, opened 2025-05-14
globalAlignmentAnchor Debug Bed

RavenYL:
# Description
🐛 Bug Report
- new bed.Read checking for start==end failed here because test 3 reads an intermediate (helper function output) bed file with an illegal entry (start==end)
- removed the writing and reading of bed files in globalAlignmentAnchor step 1 and step 2 code
- instead, for step 1 and step 2, pass bed variable around internally in the program
- TODO for the future for ideal output handling: change []bed.Bed to a slice of pointers to bed, and use nil to indicate gap_size_0, instead of start==end or chr=="gap_size_0" or name=="gap_size_0"

### Testing
- updated tests 1 and 2
- commented out test 3, which is the one that fails because it tests helper functions for step 1 and step 2 code
- did not delete existing testdata files, which include the expected intermediate (helper function output) bed files


hcnapier (approved):
LGTM!

PR #1107, opened 2025-07-22
findAndReplace regex Bug Fix

hcnapier:
# Description
Small fix. The most memory and time intensive step for findAndReplace -regex is compilation of the regex patterns. As written, the entire pattern file was compiled for each line. This makes the program take a very long time to run for long pattern files and/or query files with many lines. I've moved the pattern compilation step outside of the line by line for loop, so it only compiles one time for each command call, dramatically increasing the program's speed.

### Testing

None


sweavs111 (approved):
looks great

sweavs111 (approved):
looks great

PR #1108, opened 2025-07-28
fastqReplaceSeq

hcnapier:
# Description
fastqReplaceSeq finds a matching slice of base from the sequence line of a fastq and replaces it with a new slice of base. As written, findReplaceSeq finds matching sequences base-by-base, beginning at the start of the line. I've added an option to output only records that have an altered sequence field, but the default is to output all records in the file, regardless of whether or not they were altered. The program will currently fail if the user inputs an illegal findReplace file, if the "find" sequence is larger than the "query" sequence, or if no replacements are made.

My use case is replacing PIP-seq 16bp cell barcodes with 10X whitelist cell barcodes to allow for cellRanger analysis of PIP-seq sequencing runs. This implementation should be faster than findAndReplace, which is prohibitively slow on large files.


cmd/fastqReplaceSeq/fastqReplaceSeq.go
```diff
@@ -0,0 +1,171 @@
...
+	"strings"
+
+	"github.com/vertgenlab/gonomics/dna"
+	"github.com/vertgenlab/gonomics/exception"
+	"github.com/vertgenlab/gonomics/fastq"
+	"github.com/vertgenlab/gonomics/fileio"
+)
+
+// readFindReplaceDNA converts a delim separated findReplace file to a map of []dna.Base
+func readFindReplaceDNA(filename string, delim string) map[string][]dna.Base {
```
hcnapier, line 20: NOTE: Change this from a map to a find/replace struct of 2 []dna.Base

cmd/fastqReplaceSeq/fastqReplaceSeq.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	err = in.Close()
+	exception.PanicOnErr(err)
+
+	return findReplaceMap
+}
+
+// compareBase returns an integer related to the lexographical order of nucleotides.
+// i.e. A < C < a < c < Dot < Gap.
+func compareBase(alpha dna.Base, beta dna.Base, ignoreCase bool) int {
```
hcnapier, line 53: TODO: make uppercase in dna package and take out of here

cmd/fastqReplaceSeq/fastqReplaceSeq.go
```diff
@@ -0,0 +1,171 @@
...
+// outputs true if find sequence is found in query sequence
+func findSeq(seq []dna.Base, find []dna.Base, ignoreCase bool) bool {
+	if len(find) > len(seq) {
+		log.Fatalf("Error: Length of find sequence must be less then or equal to length of query sequence.")
+	}
+	found := true
+	for currBase := range find {
+		if found {
+			if compareBase(seq[currBase], find[currBase], ignoreCase) != 0 {
+				found = false
```
hcnapier, line 77: TODO: Remove found variable and return false here

cmd/fastqReplaceSeq/fastqReplaceSeq.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+		}
+	}
+	return found
```
hcnapier, line 81: Return true here

cmd/fastqReplaceSeq/fastqReplaceSeq.go
```diff
@@ -0,0 +1,156 @@
...
+type Settings struct {
+	inFile              string
+	outFile             string
+	findReplaceFile     string
+	findReplaceDelim    string
+	ignoreCase          bool
+	replacedRecordsOnly bool
+}
+
+func usage() {
```
sweavs111, line 83: It might be nice to put an example/format of the replace file in the usage statement.

I assume (by default) it is something like:
findBases \t replaceBases
eg:
ACTG \t GGTT

cmd/fastqReplaceSeq/fastqReplaceSeq.go
```diff
@@ -0,0 +1,156 @@
...
+	var replaceCounter int = 0
+	fq := fastq.GoReadToChan(s.inFile)
+	out := fileio.EasyCreate(s.outFile)
+	var err error
+
+	findReplacePairs = readFindReplaceDNA(s.findReplaceFile, s.findReplaceDelim)
+
+	for currRecord := range fq {
+		found := false
+		for _, currPair = range findReplacePairs {
```
sweavs111, line 106: Instead of having to loop through all the find and replace pairs for every fastq entry, what if you store the find and replace in a map as strings. I am thinking with the find as key and replace as the value. You can then query the map with the fastq sequence (as a string), which will tell you if the sequence needs to be replaced without the need to loop. I think that might be much faster

sweavs111 (commented):
looks good to me. Just one style change, one usage update, and one variable that can be declared outside a loop

sweavs111 (commented):
Changes LGTM. I am ready to approve the code. I did notice that it seems like you're trying to merge fastq/replaceName.go which only has a header? I am reading that correctly. If so, I would delete that file from the PR.

fastq/replaceName.go
```diff
@@ -0,0 +1 @@
+package fastq
```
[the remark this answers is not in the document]
hcnapier: That's weird, thanks for the easy fix!

sweavs111 (approved):
GREAT!

ndzikowski:
This all looks good to me, I really can't think of anything to add to or fix in the code itself. There are additional cases you can test in your test data to be sure that everything is working properly (for example: case sensitivity, the "find" sequence occurring somewhere in the middle rather than the beginning, etc.), but these might not be absolutely necessary for your application.

ndzikowski (approved):
Approve

PR #1110, opened 2025-08-30
locateCG updates: single vs. two-genome, conserved CpGs, + more

ndzikowski:
# Description

Expanded functionality for locateCG:
- The default mode now takes a single sequence fasta and outputs a BED file with the coordinates of every CpG site in the sequence. These can be used as "universe" files for enrichment analyses of specific sets of CpG sites (for example: is this set of differentially methylated CpG sites enriched for {TFBS,genomic region, etc.}, relative to all of the CpG sites in the genome?)
- If -compare mode is used, the program takes a multiFa with exactly 2 aligned sequences and finds gained, lost, or conserved CpG sites in sequence 1 relative to sequence 2 (*this is what the program used to do by default*)
- Types of CpG comparisons: gain, loss, and **conserved (new**). Specified with -cgtype={"gain","loss","cons"} in -compare mode.
- Added **ToFile()**: file-writing helper function
- Updated start/end coordinate format of CpG sites to match BED file format: was always 0-based, but now updated to also be half-open (end coordinate-exclusive)
- "Ref" and "Alt" bases now report the 2 bases in both sequences, regardless of whether a substitution occurred (previously, only the substituted bases for gains and losses were reported)

*Note: Would especially appreciate suggestions for making code more concise & clean. Thank you!*

### Testing
- []Tested locateCG.go, PASS


cmd/locateCG/locateCG.go
```diff
@@ -14,91 +15,170 @@ import (
...
 	StartPos int    //start pos of CG site in reference genome of sequence 1
 	EndPos   int    //end pos of CG site in reference genome of sequence 1
 	Type     string //gain or loss in sequence 1 relative to sequence 2
-	AlnStart int    //corresponding alignment position of reference StartPos
-	AlnEnd   int    //corresponding alignment position of reference EndPos
 	Ref      string //sequence 1 base(s)
 	Alt      string //sequence 2 base(s)
+	AlnStart int    //corresponding alignment position of reference StartPos
+	AlnEnd   int    //corresponding alignment position of reference EndPos
+	//FieldsInitialized int ////number of fields that are initialized, used for smart writing
```
hcnapier, line 32: Can this commented out line be removed?

cmd/locateCG/locateCG.go
```diff
@@ -14,91 +15,170 @@ import (
...
 }
 
-func locateCG(s Settings) {
+// helper function for file writing in --compare mode
+func ToFile(fileName string, records []CGdiff) {
```
hcnapier, line 36: Question for others: should this be a lowercase function since it's only being used internally?

cmd/locateCG/locateCG.go
```diff
@@ -14,91 +15,170 @@ import (
...
+	}
+
+	seq := f[0].Seq
+	//if sequence is empty
+	if len(seq) == 0 {
+		log.Fatalf("Error: fasta sequence is empty.\n")
+	}
+
+	//pre-allocate size of output
+	//output = make([]CGsite, 0, 10000)
```
hcnapier, line 77: Necessary to keep this?

cmd/locateCG/locateCG.go
```diff
@@ -14,91 +15,170 @@ import (
...
+
+	//for length of fasta sequence
+	for i := 0; i < len(seq)-1; i++ {
+		//store dna.Bases in pairs
+		f1, f2 = seq[i], seq[i+1]
```
hcnapier, line 82: This is a personal preference, but I like descriptive variable names over short variable names because I find they make things a little more readable.

cmd/locateCG/locateCG.go
```diff
@@ -14,91 +15,170 @@ import (
...
 	}
-	//add check for equal lengths
+
+	//check for equal sequence lengths
 	if len(firstSeq) != len(secondSeq) {
 		log.Fatalf("Seq1 and seq2 not equal in length.")
 	}
+
 	//pre-allocate size of output
 	//output = make([]CGsite, 0, 10000)
```
hcnapier, line 126: Necessary to keep this?

cmd/locateCG/locateCG.go
```diff
@@ -14,91 +15,170 @@ import (
...
-			refBases += dna.BaseToString(f2)
-			altBases += dna.BaseToString(s2)
-		}
 
-		//if there is a CG in sequence 1
+		//store "Ref" bases from sequence 1 and "Alt" bases from sequence 2
+		refBases = dna.BaseToString(f1) + dna.BaseToString(f2)
+		altBases = dna.BaseToString(s1) + dna.BaseToString(s2)
+
+		//if CG in sequence 1
```
hcnapier, line 148: If you wanted to make this a little more readable and a little cleaner, you could make a set of functions that addresses each of these comparisons. For example: you could write `isConserved`, that outputs TRUE if there's a CG in seq1 and seq2 and FALSE otherwise. Then you can just call `isConserved` here and add it to the output if necessary

cmd/locateCG/locateCG.go
```diff
@@ -14,91 +15,170 @@ import (
...
-			if !(s1 == dna.C && s2 == dna.G) {
-				//convert alignment position to reference position
+			//and CG in sequence 2
+			if s1 == dna.C && s2 == dna.G {
+
+				//convert alignment positions to reference positions
+				startPos := fasta.AlnPosToRefPosCounter(f[0], i, refStart, alnStart)
+				endPos := startPos + 1
+				//add to "conserved CpGs"
+				if s.CGtype == "cons" {
```
hcnapier, line 157: From my understanding, we only need to check for conservation if `s.CGtype == "cons"`. If that's the case, this if statement could be moved earlier to prevent checking for conservation and assigning the start and end positions when it's not necessary.

hcnapier (commented):
Looks good overall! I left a few comments on commented-out lines that could potentially be removed as well as a few suggestions for ways to make the code a bit cleaner. Happy to discuss more if you'd like!

PR #1111, opened 2025-08-30
PfaFindFast expansion

RavenYL:
# Description

🐛 Bug Report

### Testing

None


dna/pDna/pDna.go
```diff
@@ -43,6 +43,26 @@ func IsN(p Float32Base) bool {
 	}
 }
 
+// IsConfident returns true for a position if the most likely base >= threshold
```
RavenYL: Does not check for proper probability vector (A+C+T+G == 1, not gap, not unnormalized likelihoods that add up to !=1), but should be ok, especially since gaps should fail any threshold > 0

PR #1112, opened 2025-10-02
countCG: count CpGs (or CpG site changes) in genomic regions

ndzikowski:
# Description
countCG does exactly what the name implies. Details:
- Input: 
    1. Directory of per-chromosome Fastas (or multiFas), for all chromosomes covered by query regions
    2. BED file with all query regions (all chromosomes)
    3. Outfile name
- Modes:
    1. Default: takes single-sequence fastas + a BED file and counts all CpG sites within the BED regions
    2. --compare Mode: takes a multiFa filtered down to 2 aligned sequences + a BED file and counts all gained, lost, and constant CpG sites, in sequence 1 relative to sequence 2, within the BED regions
- Output:
     A file containing the regions and their counts

Addt'l notes:
- Fastas must be in one directory and follow the naming convention "chr1.fa", "chr2.fa", etc. 
- BED regions must have a unique "name" in column 4, as the region and counts info gets merged at the end by name. 
-Happy to move helper functions into different folders in Gonomics 

- [X] The command `go fmt` or `make clean` was used on all files included

Unfortunately, in --compare mode, the program currently counts something like: seq 1 ACCG vs. seq 2 AC-G as a "loss" in followed by a "gain", although this is more like a "constant" CpG, since you can't really tell which of those Cs got inserted. Situations like these are reflected in the test fastas I provided.

I'm not sure how often a situation like this actually arises during alignment. I have some ideas for a fix, but it will have to happen further down the line, as the solutions I've tried so far all have downsides.


cmd/countCG/countCG.go
```diff
@@ -0,0 +1,367 @@
...
+		output = append(output, bed.Bed{Chrom: chrom, ChromStart: startAlnPos, ChromEnd: endAlnPos, Name: regName, FieldsInitialized: 4})
+
+		//reset the ref/alignment position pair to that of the current region, so that it can be used for the next region
+		lastRefPos = region.ChromEnd
+		lastAlnPos = endAlnPos
+	}
+	return output, nil
+}
+
+func countCpG(fastaFile string, alnBed []bed.Bed) (regionCpGs []bed.Bed, err error) {
```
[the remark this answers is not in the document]
ndzikowski: I'll add some additional info to usage + some comments in the code too. To clarify, "C-G" in seq1 CAN be counted as a CpG, it just depends on the surrounding context/the bases in seq2. For example: seq1 C-G seq2 CGG should be counted as: 1 constant CpG site. In seq1, nextBase() pulls over the G to the left, so the pairwise comparison would be CG vs. CG (constant). The only instance where this gets iffy is: "C-G" vs. "CCG", where it gets counted as one gained CG vs one lost CG. The program, as written, should be working this way? I hope I am not wrong

cmd/countCG/countCG.go
```diff
@@ -0,0 +1,367 @@
...
+		//get the following information
+		chrom := region.Chrom
+		start := region.ChromStart
+		end := region.ChromEnd
+		name := region.Name
+
+		//Find the current region in the fasta sequences:
+		//(Here, we assume that both the genome of interest and the ancestral genome are unmodified
+		//from the original multiple alignment, so the same alignment coordinates apply to both)
+		firstRegseq := firstSeq[start:end]
```
RavenYL, line 207: Do bed and golang slice indexing match up? I know start definitely match up. bed region.ChromEnd does not include the end base. Does firstSeq[first:end] include the end base?

cmd/countCG/testdata/two_genome_expected.txt
```diff
@@ -0,0 +1,6 @@
+Chrom	Start	End	Name	Gain	Loss	Cons
```
RavenYL, line 1: Maybe this is for a future PR, and maybe this will help with QC and debugging, I'm wondering if it will be useful to have an output file that prints the chrom, start, end, actual alignment (e.g. species1=C-G species2=CCG), and processed alignment (e.g. species1=CG species2=CC), whenever there is a CpG gain/loss?

ndzikowski: I agree! Right now, locateCG returns something similar to the "actual alignment" in the output file for gains/losses, I would be happy to modify the way it does this or incorporate it into countCG in the future

PR #1124, opened 2026-02-23
add bedregion option to samInfo mapq

sweavs111:
# Description

This is a follow up PR to the one that got merged in yesterday. It adds an option to the saminfo mapq function where the user can specify a bed region and only alignments overlapping that bed region will be analyzed for the mapping quality histogram.

The consequential function to look at is the bedRegionsOnly sub function. All other code should be the same as before, but I had to break it up into multiple sub-functions so that both bedRegionsOnly  and the original wholeGenome functions could use the sub functions.

lmk if you have any questions!

🐛 Bug Report

### Testing

None


cmd/samInfo/coverage.go
```diff
@@ -38,15 +38,17 @@ func parseCoverageArgs() {
 	err = coverageFlags.Parse(os.Args[2:])
 	exception.PanicOnErr(err)
 	coverageFlags.Usage = func() { coverageUsage(coverageFlags) }
+
+	var countNinDepth *bool = coverageFlags.Bool("countNinDepth", true, "If true, count 'N' reads towards total depth of pileups.")
+	var verbose *int = coverageFlags.Int("verbose", 0, "Set to 1 to reveal debug prints. Verbose in this program reports Poisson parameter lambda.")
+	var highEndFilter *float64 = coverageFlags.Float64("highEndFilter", 0.001, "Percent threshold from right end of distribution to be filtered out")
+
```
sweavs111: This is to fix a bug in the "coverage" subcommand where the optional arguments weren't defined until after the logfatal check for number of arguments. So previously they weren't printed out when the user types` ~/go/bin/samInfo coverage`

hcnapier (approved):
LGTM

craiglowe (approved):
Great!

PR #1125, opened 2026-03-05
add 'stats' option to axTools

sweavs111:
# Description
This is a PR that adds an option to axTools to calculate summary statistics for AXT files. The statistics currently are: alignment size (based on reference interval) and percent identity over the whole interval. Within this option is another option to calculate summary statistics only for AXT records that overlap a bed record from an input bed file.

Overall, I think axTools could benefit from a re-org that adds settings structs, better testing, subcommands. However, it was a bit out of the scope of this PR, and the time I have to allocate right now. Just something to keep on folks' radar if anyone is looking for a coding project.

🐛 Bug Report

### Testing

None


interval/lift/lift.go
```diff
@@ -203,3 +203,21 @@ func checkCompatability(a axt.Axt, region interval.Interval) bool {
...
+		log.Fatalf("The interval you are trying to assay is not entirely within the axt reference coordinates.")
+	}
+	idxStart, idxEnd := refCoordToRefIdx(a, refInterval)
+	return percentIdentity(a, idxStart, idxEnd)
+}
+
+func percentIdentity(a axt.Axt, idxStart, idxEnd int) float64 {
+	var c int
+	for i := idxStart; i < idxEnd; i++ {
+		if dna.ToUpper(a.RSeq[i]) == dna.ToUpper(a.QSeq[i]) {
```
craiglowe: Should we check to make sure that at least one of the values is a "normal" base?  I am thinking about things like gaps aligned to each other and how we should treat Ns

craiglowe (commented):
Looks great!  I had a few small comments

hcnapier (commented):
Looks good to me. I agree with many of Craig's small comments, but don't have much else to add

craiglowe (approved):
Wonderful!

hcnapier (approved):

PR #1126, opened 2026-03-27
fixing simulation test

craiglowe:
This fixes a test that is failing for me (and others?) on multiple machines.  I see the program was updated 2 years ago, but the tests have not been updated for 3 years.  The recent updates to the program look like they would change the output, but the test that was failing purely tests that the output is the same for the same input.  Riley and Sarah, can you two double check this?


rimangan:
Hi Craig, I'm passing tests on main and failing on your branch, perhaps a machine specific issue? I'm not immediately sure how to proceed.

craiglowe:
Is it a Go version issue?  Would you be open to upgrading to the latest version and see what happens?

________________________________

rimangan:
@craiglowe updating the go version indeed resolved the issue, your branch now passes while main fails, so I'm adding my approval to this PR

rimangan (approved):

saqzhao (approved):
Looks good to me!

PR #1130, opened 2026-05-08
Failing testcases from failure to close files

saqzhao:
# Description

🐛 Bug Report

There are several instances where files are opened but not closed, so when the tests generate some output file and try to remove it, the machine still has a lock on the file and cannot remove the file.

There is also an issue with how paths are joined (`\` for Windows and `/` for Linux) in the cmd/gonomics which interferes with how the testing module is created (see this issue https://github.com/golang/go/issues/30616)

### Testing

Fixes so that previous tests are now passing

### DIFF
--- a/cmd/lastZWriter/lastZWriter.go
+++ b/cmd/lastZWriter/lastZWriter.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -23,7 +24,7 @@ func MakeArray(lastZ string, pairwise string, speciesListFile string, refListFil
 	}
 	speciesList := fileio.Read(speciesListFile)
 	refList := fileio.Read(refListFile)
-	fileio.EasyCreate(outText)
+	// fileio.EasyCreate(outText)
 	var parameters []string
 	var matrix string
 	var allLines []string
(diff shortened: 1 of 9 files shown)


cmd/lastZWriter/lastZWriter.go
```diff
@@ -23,7 +24,7 @@ func MakeArray(lastZ string, pairwise string, speciesListFile string, refListFil
 	}
 	speciesList := fileio.Read(speciesListFile)
 	refList := fileio.Read(refListFile)
-	fileio.EasyCreate(outText)
+	// fileio.EasyCreate(outText)
```
craiglowe: If this file is not needed, maybe we delete the line instead of commenting it out?
Am I correct that this would return a file handle, but since we are not capturing the return value it would just open a file handle that we would then not be able to access/close?

sort/mergeSort.go
```diff
@@ -147,13 +152,17 @@ func chunkData[E any](data <-chan E, recordsPerChunk int, less func(a, b E) bool
 // gob decoder.
 func writeTmpFile[E any](chunk []E, tmpDir string) *os.File {
 	file, err := os.CreateTemp(tmpDir, "sort_chunk_")
-	defer file.Close()
 	exception.PanicOnErr(err)
+	// defer file.Close()
```
craiglowe: Maybe delete this line instead of commenting out?

craiglowe (dismissed):
Thank you!  I was not getting errors (I only test on linux), but you are definitely fixing bugs that were not being caught by other programs/people.  I am approving, but I think that it might be better to completely delete the lines you are commenting out.

cmd/lastZWriter/lastZWriter.go
saqzhao: Got it, these will be removed! In line 42 when we write the output with ```fileio.Write(outText, allLines)```, fileio.Write actually also creates the output file, so we are essentially double creating this file.

sort/mergeSort.go
saqzhao: Will do!

rimangan (approved):

craiglowe (approved):
Thank you!

PR #1134, opened 2026-07-09
fixing issues with failing tests on main (ready for review)

saqzhao:
# Description

🐛 Bug Report
Fixing issues caused by upgrade to go version 1.25.0 and image 0.38.0 in the go.mod file.

**Issue 1**
```rand.Seed(s.SetSeed)``` has been deprecated, must be replaced with
```rng := rand.New(rand.NewSource(s.SetSeed))```
 and then ```rng``` is called by ```generatedRandNum = rng.Float64()``` or similarly.

**Current temporary fix:**
put ```godebug randseednop=0``` in go.mod

**Issue 2**
Non-constant format string in call to log.Fatalf / fmt.Fprintf, etc. (present in go versions 1.24+)
_Example 1_
```log.Fatalf(fmt.Sprintf("Error: unexpected value when converting colType to rune %d", a))```

should be changed to 
```log.Fatalf("Error: unexpected value when converting colType to rune %d", a)```

_Example 2_
```_, err = fmt.Fprintf(file, sb.String())```
should be changed to
```_, err = fmt.Fprint(file, sb.String())```

**Issue 3**
Missing checksum in go.sum, ran ```go mod tidy```

### DIFF
--- a/align/affineGap_highMem.go
+++ b/align/affineGap_highMem.go
@@ -227,10 +227,10 @@ func affineGap_highMem(alpha []dna.Base, beta []dna.Base, scores [][]int64, gapO
 func AffineGapChunk(alpha []dna.Base, beta []dna.Base, scores [][]int64, gapOpen int64, gapExtend int64, chunkSize int64) (int64, []Cigar) {
 	var alphaSize, betaSize int64 = int64(len(alpha)), int64(len(beta))
 	if alphaSize%chunkSize != 0 {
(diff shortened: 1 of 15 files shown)


dasnellings:
Thanks for looking into this! Your fix seems reasonable to me. I think we will definitely want to shift to the new random number paradigm, but it seems like this will be a nontrivial change to implement. I am onboard to leave that for a later date and implement your changes to get tests working.

dasnellings (approved):
lgtm!

PR #1137, opened 2026-07-30
Pileup hardclipping fix

dasnellings:
Small bugfix to handle a corner case in sam.Pileup where a read with insertion as the terminal Op in the cigar errors out if it follows hard-clipped bases.


hcnapier (approved):
LGTM

